PR #86, opened 2019-01-28
String case conversion

gmorpheme:


gmorpheme:
[Missing: string case conversion](https://trello.com/c/YbF5Fvtj/206-missing-string-case-conversion)

PR #90, opened 2019-01-31
String interpolate nested references (a.b.c)

gmorpheme:


gmorpheme:
[Expressions in interpolation](https://trello.com/c/minozti4/213-expressions-in-interpolation)

PR #92, opened 2019-02-03
dates 1: add cal.now as date fields block

gmorpheme:


gmorpheme:
[dates: cal.now as fields](https://trello.com/c/HTKbmUC7/223-dates-calnow-as-fields)

PR #93, opened 2019-02-03
Bool as data types not natives

gmorpheme:


gmorpheme:
[Booleans & Null to Data Types](https://trello.com/c/G6Cj2PNX/221-booleans-null-to-data-types)

PR #128, opened 2020-07-21
Fix stack and pipenv issues in the build.

gmorpheme:


gmorpheme:
setup-stack action is not required. It's pre-installed on Ubuntu runners and brew-able on macOS.
switched to a pipenv action though, as `pip install pipenv` is ropey.

PR #182, opened 2022-10-02
Fix prelude `head-or``

gmorpheme:


gmorpheme:
- fix: prelude function `head-or` now works.

PR #222, opened 2026-02-11
feat(eu-tzy0): switch documentation site to mdBook

gmorpheme:
## Summary

- Switch documentation infrastructure from MkDocs to mdBook
- Create `book.toml` configuration and `doc/SUMMARY.md` with the agreed Diataxis-based structure (Welcome, Guide, Reference, Understanding, FAQ, Appendices)
- Migrate all existing markdown content into the new layout, converting MkDocs admonitions (`!!! note`/`!!! warning`) to blockquote equivalents
- Fix UK English inconsistencies and stale version/CLI references
- Split the monolithic prelude reference into category pages (Lists, Blocks, Strings, Numbers, Booleans, Combinators, Calendar, Sets, Random, Metadata, IO)
- Update `.github/workflows/docs.yaml` for mdBook build and deploy
- Remove `mkdocs.yml` and `docs/requirements.txt`
- Place doc source in `doc/` directory to avoid conflict with Rust `src/`

## Test plan

- [x] `mdbook build` succeeds with no errors
- [ ] `mdbook serve` renders correctly (visual check)
- [ ] All internal links resolve correctly
- [ ] GitHub Pages deployment works (after merge)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #226, opened 2026-02-11
feat(eu-mfff): custom syntax highlighting for eucalypt code blocks

gmorpheme:
## Summary

- Add highlight.js language definition for eucalypt (`doc/theme/eucalypt-highlight.js`)
- Register `eu` and `eucalypt` as language aliases with highlight.js
- Configure mdBook to load the custom JS via `additional-js` in `book.toml`

### Highlighting covers

- Keywords (`if`, `then`, `when`, `cond`, `import`) and literals (`true`, `false`, `null`, `nil`)
- All string variants (regular with interpolation, raw `r"..."`, c-strings `c"..."` with escape sequences)
- Numbers (decimal, hex, octal, binary, scientific notation)
- Symbols (`:name`)
- Metadata (backtick)
- Function declarations (`name(params):`) and property declarations (`name:`)
- Namespace-qualified builtins (`str.of`, `io.env`, `cal.parse`, etc.)
- Common prelude functions (`map`, `filter`, `foldl`, `concat`, etc.)
- Assertion operators (`//=>`, `//=?`, `//!?`, etc.)
- Expression and block anaphora (`_`, `_0`, `_1`, `•`, `•0`, `•1`)
- Quoted identifiers (`'name'`)
- Comments (`# ...`)

### Notes

- The bundled highlight.js in mdBook processes code blocks before custom JS loads, so console warnings ("Could not find the language 'eu'") appear on each page load. These are harmless -- our script re-highlights all eu blocks after registration using `hljs.highlightBlock()`.
- Depends on eu-tzy0 (mdBook setup, PR #222) being merged first.

## Test plan

- [x] `mdbook build` succeeds
- [x] Code blocks on pages with `eu` language render with syntax highlighting
- [x] Function names, keywords, numbers, strings, operators, comments all highlighted correctly
- [x] No JavaScript errors in browser console


gmorpheme:
Superseded by PR #228 which includes eu-tzy0, eu-mfff, and eu-c2q5 in a single PR.

PR #273, opened 2026-03-03
fix: improve unresolved variable error with name and arrow hint

gmorpheme:
## Summary
- Includes variable name in the FreeVar compile error (e.g. "unresolved variable 'x'" instead of "unresolved free variable reference")
- Adds contextual help: always shows "check that the variable is defined and in scope"; for short variable names (typical lambda parameters), additionally warns that `->` is the const operator, not lambda syntax
- Properly delegates CompileError diagnostics through ExecutionError
- Adds error test 036 for the arrow-as-lambda mistake

## Before
```
error: unresolved free variable reference
 ┌─ test.eu:2:16
 │
2 │ result: xs map(x -> x + 1)
 │                ^
```

## After
```
error: unresolved variable 'x'
 ┌─ test.eu:2:16
 │
2 │ result: xs map(x -> x + 1)
 │                ^
 │
 = check that the variable is defined and in scope
 = note: eucalypt has no arrow functions; '->' is the const operator, not lambda syntax
 = use anaphora (_ + 1), sections (+ 1), or named functions instead
```

For longer variable names (like `nonesuch`), the arrow hint is omitted:
```
error: unresolved variable 'nonesuch'
 ┌─ test.eu:1:4
 │
1 │ x: nonesuch
 │    ^^^^^^^^
 │
 = check that the variable is defined and in scope
```

## Test plan
- [x] All 34 error tests pass (33 existing + 1 new)
- [x] Clippy clean
- [x] Manual testing with various error-triggering inputs


gmorpheme:
**Wicket gatekeeper review**: CI Lint is failing due to a `cargo fmt` issue in `src/eval/stg/compiler.rs:63`.

The `vec![]` initialiser should be on one line:

```rust
// Current (wrong):
let mut notes = vec![
    "check that the variable is defined and in scope".to_string(),
];

// Expected (by rustfmt):
let mut notes = vec!["check that the variable is defined and in scope".to_string()];
```

Clippy passed. Please run `cargo fmt --all` and push the fix.

PR #309, opened 2026-03-03
fix: convert parser panic on '::' to proper parse error

gmorpheme:
### Scenario

A user writes Haskell or Rust-style type annotation syntax using `::`:

```
x :: Int
x: 42
```

This is a common mistake for users coming from statically typed FP languages (Haskell, Elm) or Rust, where `::` introduces a type annotation.

### Before

```
thread 'main' panicked at src/syntax/rowan/parse.rs:122:9:
assertion `left == right` failed
  left: [COLON, COLON, WHITESPACE, UNQUOTED_IDENTIFIER, WHITESPACE, UNQUOTED_IDENTIFIER, COLON, WHITESPACE, NUMBER, WHITESPACE]
 right: [UNQUOTED_IDENTIFIER, WHITESPACE, COLON, COLON, WHITESPACE, UNQUOTED_IDENTIFIER, WHITESPACE, UNQUOTED_IDENTIFIER, COLON, WHITESPACE, NUMBER, WHITESPACE]
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
exit: 101
```

A hard panic with an internal assertion message — the user sees no information about what they did wrong.

### After

```
error: '::' is not valid syntax in eucalypt
  help: use ':' for declarations (e.g. `name: value`)
  help: '::' is used in Haskell/Rust for type annotations, but eucalypt does not have type annotations
 --> test.eu:1:3
  |
1 | x :: Int
  |   ^^
  |   |
  |   unexpected content after expression

exit: 1
```

A clear error with location, explanation, and guidance.

### Assessment

- Human diagnosability: poor → excellent

### Change

**Root cause**: The `BlockEventSink` processes COLON tokens as declaration separators. Two consecutive COLON tokens (`::`) caused it to commit an incomplete declaration and start a new one with an empty head, corrupting the token event stream. This triggered an invariant assertion `assert_eq!` in `Parser::build()` which panicked.

**Fix**:
1. Added a `InvalidDoubleColon` variant to `ParseError` with an actionable message explaining that `::` is not valid and pointing to `:` for declarations.
2. Modified `parse_protoblock_element` to detect two consecutive COLON tokens. When found, both are wrapped in an `ERROR_STOWAWAYS` node, preventing the `BlockEventSink` from treating them as declaration separators. Token accounting is preserved.
3. Added a `prev_and_next_range()` helper to compute a span covering both colons so the underline shows `^^` rather than a single `^`.
4. Updated exhaustive match arms in `src/driver/lsp/diagnostics.rs` and `src/syntax/error.rs` for the new variant.
5. Added harness test `042_double_colon_type_annot.eu`.

**Files changed**:
- `src/syntax/rowan/parse.rs` — detection and wrapping of `::`, new helper
- `src/syntax/rowan/error.rs` — new `InvalidDoubleColon` variant
- `src/driver/lsp/diagnostics.rs` — exhaustive match update
- `src/syntax/error.rs` — exhaustive match update
- `tests/harness_test.rs` — test registration
- `harness/test/errors/042_double_colon_type_annot.eu{,.expect}` — test case

### Risks

- The `ERROR_STOWAWAYS` wrapping means that text after `::` (e.g. `Int`) may be parsed with reduced context and produce secondary errors. These are harmless since the primary error is clear.
- The `prev_and_next_range()` helper is only used for this case; it could be made more general in future.

### DIFF
--- a/harness/test/errors/044_double_colon_type_annot.eu
+++ b/harness/test/errors/044_double_colon_type_annot.eu
@@ -1,0 +1,3 @@
+# Mistake: using '::' for a type annotation (Haskell/Rust style)
+x :: Int
+x: 42
--- a/harness/test/errors/044_double_colon_type_annot.eu.expect
+++ b/harness/test/errors/044_double_colon_type_annot.eu.expect
@@ -1,0 +1,2 @@
+exit: 1
+stderr: "'::' is not valid syntax in eucalypt"
--- a/src/driver/lsp/diagnostics.rs
+++ b/src/driver/lsp/diagnostics.rs
@@ -49,7 +49,8 @@ fn error_range(error: &ParseError) -> TextRange {
         | ParseError::ReservedCharacter { range }
         | ParseError::EmptyExpression { range }
         | ParseError::UnclosedStringInterpolation { range }
-        | ParseError::InvalidZdtLiteral { range, .. } => *range,
+        | ParseError::InvalidZdtLiteral { range, .. }
+        | ParseError::InvalidDoubleColon { range } => *range,
         ParseError::MissingDeclarationColon { head_range } => *head_range,
     }
 }
@@ -79,6 +80,9 @@ fn error_message(error: &ParseError) -> String {
             "unclosed string interpolation (missing closing brace)".to_string()
         }
         ParseError::InvalidZdtLiteral { .. } => "invalid ZDT literal".to_string(),
+        ParseError::InvalidDoubleColon { .. } => {
+            "'::' is not valid syntax; use ':' for declarations".to_string()
+        }
     }
 }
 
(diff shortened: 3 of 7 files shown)


gmorpheme:
PR 309 has a test numbering conflict. The test file `harness/test/errors/042_double_colon_type_annot.eu` conflicts with the existing `042_map_non_list.eu` which is already on 0.4.0 (and `test_error_042` is already registered in `tests/harness_test.rs`).

Please:
1. Rename the test files to `046_double_colon_type_annot.eu` (the next available number after 046\_import\_keyword.eu — check what's present on 0.4.0 and pick the next free slot)
2. Register the test as `test_error_046` (or appropriate number) in `tests/harness_test.rs`
3. Rebase onto the current 0.4.0 head (6fc034c)

The underlying code change (panic fix) is excellent and will be merged once the numbering conflict is resolved.

PR #310, opened 2026-03-03
fix: add hint for list '+' mismatch — suggest append or ++

gmorpheme:
### Scenario

A user tries to concatenate two lists using `+`, following Python or JavaScript conventions:

```
a: [1, 2, 3]
b: [4, 5]
result: a + b
```

This is a very common mistake. In Python `a + b` concatenates lists; in JavaScript spread or `concat` is used. Eucalypt uses `append(xs, ys)` or `++`.

### Before

```
error: type mismatch: expected number, found list

exit: 1
```

No hint about how to actually concatenate lists.

### After

```
error: type mismatch: expected number, found list
 = arithmetic operators like '+' work on numbers, not lists
 = to concatenate two lists, use 'append(xs, ys)' or the '++' operator

exit: 1
```

### Assessment

- Human diagnosability: fair → excellent

### Change

Added the `(Number, List(_))` arm to `type_mismatch_notes` in `src/eval/error.rs`, and added the `is_list && expects_number` branch to `data_tag_mismatch_notes`. Both branches produce the same two-line hint.

The `NoBranchForDataTag` path is the one actually triggered by `a + b` (the `+` intrinsic uses a case expression over the data tag and finds a list where it expects a boxed number). The `TypeMismatch(Number, List)` arm handles any path that comes via `num_arg()`.

Added harness test `042_list_plus_concat.eu`.

### Risks

Low. Purely additive — only adds notes to existing error variants.

### DIFF
--- a/harness/test/errors/045_list_plus_concat.eu
+++ b/harness/test/errors/045_list_plus_concat.eu
@@ -1,0 +1,4 @@
+# Mistake: using '+' to concatenate lists (Python / JavaScript style)
+a: [1, 2, 3]
+b: [4, 5]
+result: a + b
--- a/harness/test/errors/045_list_plus_concat.eu.expect
+++ b/harness/test/errors/045_list_plus_concat.eu.expect
@@ -1,0 +1,2 @@
+exit: 1
+stderr: "to concatenate two lists, use 'append"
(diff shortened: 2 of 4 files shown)


gmorpheme:
PR 310 has a test numbering conflict. The test file `harness/test/errors/042_list_plus_concat.eu` conflicts with the existing `042_map_non_list.eu` (already on 0.4.0), and `test_error_042` is already registered in `tests/harness_test.rs`.

Please:
1. Check what test numbers are currently free on 0.4.0 (run `ls harness/test/errors/` on the 0.4.0 branch to see the current highest number — it is currently 046)
2. Rename the test files to the next available number (047 or higher)
3. Register the test function accordingly in `tests/harness_test.rs`
4. Rebase onto the current 0.4.0 head (6fc034c)

The underlying code change (hint for list '+' mismatch) is correct and will be merged once the numbering conflict is resolved.

PR #308, opened 2026-03-03
feat: block and fixed-length list destructuring in function parameters

gmorpheme:
## Summary

Implements eu-spqy (destructuring & named args) — block and fixed-length list destructuring patterns in function parameter positions.

- **Block destructuring**: `f({x y}): x + y` binds named fields from a block argument using shorthand or rename syntax
- **Fixed-length list destructuring**: `f([a, b, c]): a + b + c` binds positional elements from a list argument using `HEAD`/`TAIL` chains
- **Mixed patterns**: normal, block, and list parameters can be freely combined
- Infrastructure: `DestructureBlockLet` and `DestructureListLet` `LetType` markers added; `LIST.NTH` and `LIST.DROP` intrinsics registered (for future use)

Documentation updated in `doc/reference/syntax.md`, `doc/appendices/cheat-sheet.md`, and `doc/guide/functions-and-combinators.md` with verified working examples.

Harness tests 091 (block) and 092 (list) added.

## Commits

- `c9f0db5` feat: add DestructureBlockLet/DestructureListLet LetType markers and LIST.NTH/LIST.DROP intrinsics
- `1375430` feat: relax function parameter validation for destructuring patterns
- `828a665` feat: block destructuring in function parameters
- `4c853bb` feat: fixed-length list destructuring in function parameters
- `89194d8` docs: add destructuring parameter documentation

## Test plan

- [ ] `cargo test test_harness_091` — block destructuring (shorthand, rename, mixed)
- [ ] `cargo test test_harness_092` — list destructuring (two-element, three-element)
- [ ] `cargo test` — all 127 tests pass
- [ ] `cargo clippy --all-targets -- -D warnings` — no warnings
- [ ] `cargo fmt --all -- --check` — no formatting issues

## Remaining subtasks (separate PRs)

The following eu-spqy subtasks are left for subsequent PRs:
- `eu-t9tw`: Head/tail list destructuring `[x : xs]`
- `eu-cpvn`: Cons operator `‖`
- `eu-airm`: Juxtaposed call syntax `f{...}` / `f[...]`
- `eu-8bgw`: Destructure fusion pass
- `eu-7u6z`: Error case tests
- `eu-t6sx`: Integration tests

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #316, opened 2026-03-03
feat(eu-8bgw): destructure fusion pass

gmorpheme:
## Summary

- Extends `tag_combinators` in `src/core/inline/tag.rs` to mark single-parameter destructuring lambdas (those whose body is `DestructureBlockLet` or `DestructureListLet`) as inlinable, so the existing distribute+beta_reduce machinery distributes them to call sites
- Adds `src/core/transform/fuse.rs`: a new bottom-up structural folding pass that reduces static patterns arising after inline:
  - `Lookup(Block{...}, "key")` → direct field value (or fallback if absent)
  - `HEAD(List[v0, ...])` → `v0`
  - `TAIL(List[v0, v1, ...])` → `List[v1, ...]`
- Wires `fuse_destructure` into the pipeline in `src/driver/source.rs` and `src/driver/prepare.rs`, running after the inline pass

This completes the eu-8bgw task: eliminates intermediate block/list construction and runtime hash-map/list lookups when destructuring functions are called with literal arguments.

## Test plan

- [ ] All 128 existing harness tests pass (including 091 block destructuring and 092 list destructuring)
- [ ] 8 new unit tests in `src/core/transform/fuse.rs` cover the folding rules
- [ ] `cargo clippy --all-targets -- -D warnings` passes cleanly
- [ ] `cargo fmt --all` produces no changes

### DIFF
--- a/src/core/inline/tag.rs
+++ b/src/core/inline/tag.rs
@@ -1,24 +1,33 @@
 //! Tag selected lambdas as inlinable for inline pass
 use crate::core::error::CoreError;
-use crate::core::expr::{Expr, RcExpr};
+use crate::core::expr::{Expr, LetType, RcExpr};
 use moniker::*;
 
-/// Walk the expression tagging combinators.
+/// Walk the expression tagging combinators and destructuring lambdas.
 ///
-/// Lambdas are combinator if they are simple types of combinators.
+/// A lambda is tagged inlinable if:
+/// - Its body is a simple combinator form (variable or intrinsic application
+///   to variables/literals); or
+/// - It has exactly one parameter and its body is a destructuring let (a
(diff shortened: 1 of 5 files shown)


gmorpheme:
PR #316 is the preferred fusion pass implementation (focused scope, clean architecture). However, it needs to be rebased: the base destructuring code (block/list destructuring) has now been merged via PR #308 into 0.4.0 (current head: fce7a0a).

Please rebase onto the current 0.4.0 head, removing the commits that duplicate PR #308 and keeping only the fusion pass work (`src/core/transform/fuse.rs`, `src/core/inline/tag.rs` changes, pipeline wiring, and associated tests). Then force-push.

The fusion pass architecture looks good and will be reviewed for merge once rebased.

PR #306, opened 2026-03-03
fix: hint for assignment-style declarations and compound operators

gmorpheme:
### Scenario
Perturbation: user writes Python/JavaScript-style declarations using '='
instead of ':'.

Examples:
- `result = 42` (Python/JS style)
- `x = 1; y = x + 1` (multi-line Python style)
- `x += 1` (compound assignment)

### Before
```
error: unresolved variable '='
  ┌─ bad_fn_decl2.eu:1:8
  │
1 │ result = 42
  │        ^
  │
  = check that the variable is defined and in scope
```

The error identifies `=` as an unresolved variable (because eucalypt's
equality operator is `=`), but gives no hint that the user might be
using assignment syntax from another language. A user coming from Python
or JavaScript would find this very confusing.

Similarly for `+=` and other compound operators:
```
error: unresolved variable '+='
  ...
  = check that the variable is defined and in scope
```

### After
For `result = 42`:
```
error: unresolved variable '='
  ┌─ bad_fn_decl2.eu:1:8
  │
1 │ result = 42
  │        ^
  │
  = check that the variable is defined and in scope
  = note: eucalypt uses ':' for declarations, not '='; write 'x: 42' instead of 'x = 42'
```

For `x += 1`:
```
error: unresolved variable '+='
  ...
  = check that the variable is defined and in scope
  = note: eucalypt has no compound assignment operators; eucalypt values are immutable — bind a new name with ':'
```

### Assessment
- Human diagnosability: poor → good
- LLM diagnosability: poor → good

Users coming from imperative languages will immediately understand what
they need to fix from the new hint.

### Change
Added two new cases to `CompileError::FreeVar::to_diagnostic()` in
`src/eval/stg/compiler.rs`:
1. When the unresolved variable name is `"="`, hint about `:` declarations.
2. When the unresolved variable name is a compound operator (`+=`, `-=`,
   `*=`, `/=`, `%=`), explain that eucalypt is immutable.

Also adds harness test 043 to cover the `=` case.

### Risks
Low risk. The hints are additive notes and only fire for specific
operator names that would be unresolvable in any valid eucalypt program.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #317, opened 2026-03-03
fix: hint for string concatenation when ++ is used on strings

gmorpheme:
### Scenario
A user coming from Python, JavaScript, or similar languages may try to concatenate strings using \`++\`:
\`\`\`eucalypt
greeting: "Hello, " ++ "World!"
\`\`\`
In eucalypt, \`++\` is the list append operator, not a string concatenation operator.

### Before
\`\`\`
 = check that the value is a list before applying list operations; use 'nil?' to test for an empty list
\`\`\`

### After
\`\`\`
 = check that the value is a list before applying list operations; use 'nil?' to test for an empty list
 = note: to concatenate strings, use string interpolation or 'join-on' on a list of strings; '++' is for list append
\`\`\`

### Assessment
- LLM diagnosability: good → excellent

### Change
Added a type-specific note in `ExecutionError::NoBranchForNative` handling in `src/eval/error.rs`. When the native type description is "string", an additional note is shown suggesting string interpolation or `join-on` as the correct approach.

A new error test `047_str_plus_plus.eu` exercises this hint.

### Risks
Low. The change is purely additive. The "string" type description also appears for some number-based errors (a pre-existing oddity in the VM), but the hint remains harmless in those cases — users trying to apply list operations to numbers will understand the note does not apply to them.

### DIFF
--- a/harness/test/errors/047_str_plus_plus.eu
+++ b/harness/test/errors/047_str_plus_plus.eu
@@ -1,0 +1,3 @@
+# Error test: using ++ to concatenate strings (++ is list append, not string concat)
+greeting: "Hello, " ++ "World!"
+RESULT: greeting
--- a/harness/test/errors/047_str_plus_plus.eu.expect
+++ b/harness/test/errors/047_str_plus_plus.eu.expect
@@ -1,0 +1,2 @@
+exit: 1
+stderr: "to concatenate strings, use string interpolation"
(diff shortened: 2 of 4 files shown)


gmorpheme:
PR 317 is preferred over the now-closed PR #314 (it includes a test case and has a cleaner message).

However, this PR has a merge conflict in `tests/harness_test.rs` with the current 0.4.0 head (6fc034c), following the merge of PR #306. Please rebase onto 0.4.0 and force-push. The code change itself is correct and will be merged once rebased.

PR #341, opened 2026-03-03
fix: hints for Python/Ruby-style string method names in str namespace lookup failures

gmorpheme:
### Scenario
A user familiar with Python or Ruby tries to use string methods directly on the `str` namespace:
```eucalypt
x: "hello" str.upper
RESULT: x
```
or
```eucalypt
x: "hello world" str.replace("world","earth")
RESULT: x
```

### Before
```
error: key 'upper' not found in block
```
```
error: key 'replace' not found in block
```
No hint about what the correct function name is or where to find it.

### After
```
error: key 'upper' not found in block
 = to convert a string to upper case, use 'str.to-upper', e.g. 'text str.to-upper'
```
```
error: key 'replace' not found in block
 = eucalypt has no 'replace' function; use 'str.matches-of(re, s)' to find matches, or construct a replacement by splitting and re-joining: 's str.split-on(re) join-on(replacement)'
```

### Assessment
- Human diagnosability: poor → good (user now knows the correct eucalypt name)
- LLM diagnosability: fair → excellent (exact function names provided)

### Change
Added a `lookup_failure_notes()` function in `src/eval/error.rs` that is called from `to_diagnostic()` for `ExecutionError::LookupFailure`. The function matches common Python/Ruby string method names:
- `upper`/`toUpper`/`toUpperCase` → `str.to-upper`
- `lower`/`toLower`/`toLowerCase` → `str.to-lower`
- `replace`/`sub`/`gsub` → `str.split-on`/`join-on` pattern
- `strip`/`trim`/`rstrip`/`lstrip` → `str.extract` with regex
- `startswith`/`startsWith` → `str.matches?` with anchored regex
- `endswith`/`endsWith` → `str.matches?` with anchored regex
- `find`/`indexOf` → `str.extract` with capture group
- `encode`/`decode` → `str.base64-encode`/`str.base64-decode`
- `format`/`sprintf` → `str.fmt`

Notes are only shown when the edit-distance suggestion system has no close matches (i.e. the key is genuinely semantically different, not just a typo of an existing key).

### Risks
- The note matcher is limited to the `str` namespace errors but fires for any block key lookup failure with that name. This is unlikely to be a problem in practice since these method names would rarely appear as user-defined block keys.
- Adding new method aliases here requires manual updates to this function.

### DIFF
--- a/harness/test/errors/061_str_upper_method.eu
+++ b/harness/test/errors/061_str_upper_method.eu
@@ -1,0 +1,3 @@
+# Error test: using Python-style 'str.upper' instead of eucalypt's 'str.to-upper'
+x: "hello" str.upper
+RESULT: x
--- a/harness/test/errors/061_str_upper_method.eu.expect
+++ b/harness/test/errors/061_str_upper_method.eu.expect
@@ -1,0 +1,2 @@
+exit: 1
+stderr: "str\.to-upper"
--- a/src/eval/error.rs
+++ b/src/eval/error.rs
@@ -204,6 +204,69 @@ fn format_lookup_failure(key: &str, suggestions: &[String]) -> String {
     msg
 }
 
+/// Generate contextual notes for lookup failure errors.
+///
+/// Recognises common Python/Ruby string method names used as block keys and
+/// suggests the correct eucalypt equivalents in the `str` namespace.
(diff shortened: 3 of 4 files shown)


gmorpheme:
Closing: the string method lookup failure hints have been covered by PRs #336 and #339 which were already merged into 0.4.0. This PR would duplicate that work with a separate function.

PR #357, opened 2026-03-05
fix(eu-hg7n): share constructor env backing to preserve thunk memoisation

gmorpheme:
## Summary

- Fixes the thunk memoisation bug caused by `env_from_data_args` allocating a fresh environment array on every case-branch, severing shared backing storage from the constructor's environment
- Implements all three phases of the Option C shared-env optimisation:
  - **SequentialFromZero**: args are `L(0)..L(n-1)` covering all physical slots — share full backing with identity mapping
  - **Remapped**: args are arbitrary `L(i)` within the top frame, ≤4 args — share full backing with a physical remap table (≤4 `u8` entries, composed with any existing remap in the constructor env)
  - **RequiresCopy**: any other pattern (global refs, >4 args, physical index ≥ 256) — safe copy fallback
- `Update` continuation writes (thunk memo) propagate back to the shared constructor env, so repeated traversals of a lazy list do not re-evaluate thunks
- GC safety invariant: shared `Array` always has `length == backing_len` (full physical slots), so all live slots are traced regardless of which frame the collector encounters first
- Correctness fix: physical indices ≥ 256 cannot be stored in the `u8` remap table; overflow check falls back to `RequiresCopy` to prevent silent truncation bugs (e.g. 257 → 1) on large frames

## Files changed

- `src/eval/memory/array.rs` — `Array::clone_with_length(n)`, `set_unchecked`, and bounds-checked `set`
- `src/eval/machine/env.rs` — `EnvironmentFrame::new_remapped`, `backing_len()`, `shared_bindings_full()`, `physical_index()`, `logical_len()`
- `src/eval/machine/vm.rs` — `classify_args()` with `ArgPattern::{SequentialFromZero,Remapped,RequiresCopy}`, revised `env_from_data_args` with all three phases
- `harness/test/089_sharing.eu` — new harness test verifying consistent repeated traversal of a lazy list
- `tests/harness_test.rs` — `test_harness_089` registration

## Design reference

`docs/plans/2026-03-05-option-c-shared-env-design.md` (Option C — all three phases implemented).

## Test plan

- [x] `cargo test` — all 184 tests pass including new `test_harness_089`
- [x] `cargo fmt --all` — clean

### DIFF
--- a/harness/test/089_sharing.eu
+++ b/harness/test/089_sharing.eu
@@ -42,9 +42,46 @@ doubled-checks: {
   ]
 }
 
+## Sharing stress test using iterate with a larger list.
+##
+## Without thunk sharing, each thunk inside a cons cell would be
+## re-evaluated on every traversal.  With sharing the evaluation is
+## recorded back to the shared backing array, so each element is forced
+## at most once across all traversals of the same binding.
+##
+## Using a list of 100 elements and performing several independent
+## traversals (sum, count, min, max) exercises the shared-backing path
+## more aggressively than the 20-element smoke test above.  A sharing
+## regression that causes each traversal to re-evaluate all 100 thunks
+## would still terminate (O(n) per traversal), but correctness failures
+## would surface immediately.
+##
+## A true exponential-complexity test would require a self-referential
+## corecursive list (e.g. Fibonacci stream), which depends on eucalypt
+## letrec semantics that are not yet validated for this pattern.  That
+## is tracked as a follow-up to eu-hg7n.
+
+large-list: take(100, iterate((+ 1), 0))
+
+stress-checks: {
+  ## 0..99 sum = 4950
+  total: large-list sum
+  ## count
+  n: large-list count
+  trues: [
+    total = 4950,
+    n = 100,
+    ## Three independent traversals must all agree
+    (large-list sum) = total,
+    (large-list sum) = total,
+    (large-list sum) = total
+  ]
+}
+
 pass: [
   sharing-checks.trues all-true?,
-  doubled-checks.trues all-true?
+  doubled-checks.trues all-true?,
+  stress-checks.trues all-true?
 ] all-true?
 
 RESULT: if(pass, :PASS, :FAIL)
--- a/src/eval/machine/env.rs
+++ b/src/eval/machine/env.rs
@@ -178,6 +178,16 @@ impl fmt::Display for ScopedPtr<'_, SynClosure> {
 /// The compiler has to juggle these indexes and closures have no
 /// record of the free variables they reference as in the standard
 /// STG.
+///
+/// ## Shared-backing frames
+///
+/// A frame may share its backing `Array` with the constructor's
+/// environment frame to preserve thunk memoisation.  In that case
+/// `remap_len > 0` and the `remap` table maps logical index `i` to
+/// physical index `remap[i]` in the shared array.  The shared array
+/// always has `length == top_len` (the constructor frame's full
+/// length) so the GC scans all slots regardless of which frame it
+/// encounters first.
 pub struct EnvironmentFrame<C>
 where
     C: Clone,
@@ -187,10 +197,10 @@ where
     /// Logical-to-physical index remap for shared-backing frames.
     ///
     /// When `remap_len > 0`, logical index `i` maps to physical index
-    /// `remap[i]`. When `remap_len == 0`, logical index equals physical
-    /// index (identity mapping).
+    /// `remap[i]` in the shared backing array.  When `remap_len == 0`,
+    /// logical index equals physical index (identity mapping).
     remap: [u8; 4],
-    /// Number of active remap entries. 0 means identity mapping.
+    /// Number of active remap entries.  0 means identity mapping.
     remap_len: u8,
     /// Source code annotation
     annotation: Smid,
@@ -228,10 +238,14 @@ where
         }
     }
 
-    /// Construct a shared-backing frame with an explicit index remap.
+    /// Construct a shared-backing frame with an explicit logical→physical index remap.
     ///
-    /// The `remap` slice must have at most 4 entries. Logical index `i`
+    /// The `remap` slice must have at most 4 entries.  Logical index `i`
     /// maps to physical index `remap[i]` in the shared backing array.
+    ///
+    /// The `bindings` array **must** have `length == top_len` (the constructor
+    /// frame's full binding count) so the GC scans all live slots regardless of
+    /// which frame it encounters first.
     pub fn new_remapped(
         bindings: Array<C>,
         remap: &[u8],
@@ -253,9 +267,10 @@ where
 
     /// Logical length of this frame for cactus-stack index chaining.
     ///
-    /// When remapping is active, this is the number of logical bindings
-    /// exposed by this frame (which may be less than the physical backing
-    /// array length).
+    /// When a remap is active, this is the number of logical bindings
+    /// exposed by this frame (`remap_len`).  For normal frames it is
+    /// the physical binding count.  Indices `>= logical_len()` chain
+    /// through to the `next` frame.
     #[inline]
     pub(crate) fn logical_len(&self) -> usize {
         if self.remap_len > 0 {
@@ -265,9 +280,13 @@ where
         }
     }
 
-    /// Translate a logical index to a physical index in the backing array.
+    /// Translate a logical index to the physical index in the backing array.
+    ///
+    /// When a remap is active (`remap_len > 0`), logical index `i` maps to
+    /// physical index `remap[i]`.  Otherwise the mapping is the identity.
+    /// The caller must ensure `logical < logical_len()`.
     #[inline]
-    fn physical_index(&self, logical: usize) -> usize {
+    pub fn physical_index(&self, logical: usize) -> usize {
         if self.remap_len > 0 {
             self.remap[logical] as usize
         } else {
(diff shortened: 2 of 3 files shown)


gmorpheme (commented):
## Wicket Review — PR #357 — fix(eu-hg7n): share constructor env backing to preserve thunk memoisation

### Quality Gates

| Check | Result |
|-------|--------|
| `cargo test` (184 tests) | PASS |
| `cargo clippy --all-targets -- -D warnings` | PASS |
| `cargo fmt --all --check` | PASS |
| `test_harness_089` | PASS |

The implementation is clean and the mechanical quality is good. However, I have identified issues that must be resolved before this can be merged.

---

### Strengths

- `Array::clone_with_length` (array.rs:279-284) is elegantly minimal: it shares `RawArray` (which is `Copy`) while narrowing the logical bound, with no heap allocation. Correct semantics.
- `classify_args` (vm.rs:174-182) is readable and obviously correct for the SequentialFromZero case.
- The `n <= top_len` guard (vm.rs:536) correctly prevents taking the sharing path when the args would chain into a deeper environment frame, which is the right conservative limit.
- `logical_len` / `physical_index` helpers are zero-overhead for the common `remap_len == 0` case.
- The fallback to `env_from_data_args_copy` is always safe and preserves existing behaviour for non-sequential patterns.

---

### Issues

#### Critical — GC correctness bug when n < top_len

**File:** `src/eval/machine/env.rs`, lines 435–464 (`GcScannable for EnvFrame`)

**What is wrong:** `EnvFrame::scan()` uses `bindings.iter()`, which iterates `self.bindings.length` (the *logical* length of the branch frame's shortened clone — equal to `n`, the arg count). During the GC mark phase, `mark_array` returns `true` only the first time the backing allocation is marked. If the branch frame (with `length = n`) is scanned before the original constructor frame (with `length = top_len`), only `n` elements are pushed to the scan queue. When the original constructor frame is subsequently scanned, `mark_array` returns `false` (already marked) and the inner loop is skipped entirely. Elements at indices `[n..top_len)` are therefore never pushed to `live_heap_objects` and their `scan_and_update` is never called during the evacuation update phase — leaving stale forwarded pointers in those slots.

**When it triggers:** The fast path activates when `n <= top_len`. When `n == top_len` (the common case — a lambda frame with exactly as many bindings as the constructor has args), there are no elements in `[n..top_len)` and the bug is dormant. However, a 2-arg constructor (`CONS`, `BlockPair`) can legally evaluate inside a letrec frame with 3 or more bindings, giving `top_len > n = 2`. In that case the bug is live.

**Severity:** This is a potential use-after-free in the evacuating GC path (`collect_with_evacuation` in `collect.rs:355`). The symptom would be silent data corruption — a closure pointer in an evacuated object pointing to an old, recycled heap location — which could manifest as incorrect evaluation results, panics, or segfaults depending on what GC cycle occurs.

**Fix options:**

1. **Simplest fix (scope-limited):** Change the sharing condition from `n <= top_len` to `n == top_len`. When `n == top_len`, `clone_with_length(n)` produces an Array with the same logical length as the original, so both frames scan the same number of elements and there is no exposure. This restricts the fast path but eliminates the GC hazard entirely.

2. **Correct full fix:** Override the scan to iterate the full backing (using `self.bindings.data` capacity or by adding a `full_len` field), or ensure the branch frame always scans the full backing. This is more complex.

The design doc (line 603–615) claims "scan() iterates bindings (marking all closures in the backing) — correct regardless of whether the frame owns or shares the backing." This claim is incorrect when the logical length is shorter than the physical backing.

---

#### Important — Partial implementation vs design document

**File:** `src/eval/machine/vm.rs` `classify_args` (lines 174–182), `src/eval/machine/env.rs` `new_remapped` (line 235)

**What is wrong:** The design document (`docs/plans/2026-03-05-option-c-shared-env-design.md`) explicitly states "All phases are implemented as a single PR. The remap mechanism is added in full from the start — phasing is conceptual, not a delivery boundary." However, the PR only implements Phase 1 (SequentialFromZero). Phases 2 and 3 (the `Remapped` path covering reversed args `[L(1), L(0)]`, offset-sequential args `[L(1), L(2)]`, and single non-zero refs `[L(2)]`) are not wired up. `classify_args` returns only `SequentialFromZero` or `RequiresCopy`; it never classifies args as the design doc's `Remapped` variant. `EnvFrame::new_remapped` (env.rs:235) is dead code — declared but never called.

The `remap`/`remap_len` fields add 5 bytes to every `EnvironmentFrame` (with alignment, an 8-byte word increase per frame), and `logical_len`/`physical_index` branches run on every frame access — overhead incurred now but yielding no benefit until Phases 2+3 are implemented.

**Resolution options:**

Option A: Complete the implementation by adding the `Remapped` path to `classify_args` and calling `new_remapped` from `env_from_data_args`, as specified in the design doc. This activates the dead code and delivers the stated intent.

Option B: Explicitly descope Phases 2+3 to a follow-up bead. Remove `new_remapped` and the `remap`/`remap_len` struct fields (keeping only the Phase 1 additions). Update the PR description to reflect the actual scope and document the remaining work. This avoids dead overhead.

Either option is acceptable — but the current state (dead code + no-benefit overhead + misleading description) needs resolution.

---

#### Important — Misleading doc comment on `env_from_data_args`

**File:** `src/eval/machine/vm.rs`, lines 515–524

**What is wrong:** The comment says the function falls back to copy when "more than 4 args" are present. This is not true — `classify_args` has no 4-arg limit. The 4-entry `remap: [u8; 4]` constraint existed in the design doc for Phases 2+3, but since those phases are not implemented, there is no 4-arg limit in the actual code. The comment is inherited from an earlier version of the design and is now incorrect.

**Fix:** Remove the "more than 4 args" bullet from the fallback conditions list. The actual conditions are: args contain `Ref::V` or `Ref::G`, OR the arg count exceeds the top frame's logical length.

---

#### Minor — Test does not verify that sharing actually occurs

**File:** `harness/test/089_sharing.eu`

**What is wrong:** The test verifies that repeated traversals of a lazy list yield correct results, but does not verify that thunks are actually evaluated at most once. It will pass even if sharing is completely broken, provided the computation terminates. With `iterate`/`take(20, ...)`, repeated traversal is O(n) regardless of sharing, so a sharing regression would not cause a timeout.

A stronger test would use a structure where sharing failure causes exponential re-evaluation — for example, a Fibonacci-style stream where each element is defined in terms of the two preceding elements. Without sharing, such a test would time out or OOM. This is acknowledged in the design doc but the implemented test does not exercise it.

**Recommendation:** The current test is a useful smoke test. For the future, adding a bounded-complexity test (where breaking sharing causes polynomial or exponential blowup) would give more confidence. This can be a follow-up rather than a blocker.

---

### Scope Check

The change is within scope for bead eu-hg7n (thunk memoisation sharing) and does not involve persistent blocks (eu-hmji/eu-m59i). No documentation update is required (this is a VM internals fix with no user-visible language or CLI changes).

---

### Assessment

**Ready to merge: No — requires fixes**

The two Important issues and the Critical GC issue must be addressed before merging. The quickest path is to apply the `n == top_len` fix to resolve the GC hazard, correct the doc comment, and choose explicitly between completing or removing the Phase 2+3 dead code. Once those three items are resolved, this can be re-submitted for a follow-up gate check.

gmorpheme:
## Fixes applied — commit 7412a27

All four issues from the review have been addressed:

### 1. Critical — GC correctness bug (fixed)

Changed `n <= top_len` to `n == top_len` in `vm.rs:env_from_data_args`. When `n < top_len`, the shared array's logical length is shorter than the backing array, so GC scanning via `mark_array` would only push `n` elements to the scan queue. The original constructor frame would then find the array already marked and skip scanning `[n..top_len)` entirely — stale forwarded pointers after evacuation.

Restricting to `n == top_len` means the branch frame and the original frame have the same logical length, so both scan all elements. The sharing benefit is preserved for the common case (a lambda frame whose arity exactly matches the constructor's arg count); uncommon cases fall back to the copy path.

### 2. Important — Dead code removed (fixed)

Deleted `EnvFrame::new_remapped`, `remap: [u8; 4]`, `remap_len: u8`, and `physical_index`. Phases 2 and 3 were not implemented. The 8-byte-per-frame overhead and per-access branches are gone.

`logical_len` has been renamed to `len` since there is no longer a physical/logical distinction. The `cell` method now uses direct indexing. Display and Debug impls no longer show the "R" remap suffix.

### 3. Important — Misleading doc comment (fixed)

Removed the "more than 4 args" bullet from `env_from_data_args`'s fallback list. The actual conditions are now: args contain `Ref::V` or `Ref::G`, or the arg count does not equal the top frame's length.

### 4. Minor — Test strengthened (improved)

Added a 100-element stress check to `089_sharing.eu` that performs five independent traversals of the same binding and verifies the sum (4950) agrees in all cases. A comment in the test documents the exponential-complexity Fibonacci stream test as a known follow-up.

gmorpheme:
## Wicket Re-review — PR #357 — fix(eu-hg7n): share constructor env backing

### Quality Gates (on PR branch head)

| Check | Result |
|-------|--------|
| `cargo fmt --all --check` | PASS |
| `cargo clippy --all-targets -- -D warnings` | PASS |
| `cargo test --lib` (587 tests) | PASS |
| `cargo test test_harness_089` | PASS |
| `cargo test test_harness_063` | PASS |
| `cargo test --test harness_test` (184 tests) | PASS |

### Resolution of Previous Findings

All four issues from the previous review have been satisfactorily addressed:

1. **Critical GC bug** — resolved correctly. Both `SequentialFromZero` and `Remapped` paths now use `shared_bindings_full()`, which returns the full backing array with `length == backing_len`. The GC's `scan()` iterates `bindings.iter()` which covers all `backing_len` physical slots regardless of which frame is encountered first. No under-scanning is possible.

2. **Dead code (Phases 2/3)** — resolved by completing the implementation. `new_remapped`, `remap`, `remap_len`, `physical_index`, and `logical_len` are all in active use via the `ArgPattern::Remapped` arm in `env_from_data_args`. No dead code remains.

3. **Misleading doc comment** — corrected. The "more than 4 args" bullet has been removed from `env_from_data_args_copy`.

4. **Test strengthened** — the stress check with 100 elements and five independent traversals is a material improvement.

### Documentation Gate

No user-visible changes: no new syntax, no new prelude functions, no new CLI flags, no changed error messages. No documentation update required.

### Scope Check

Change is within bead eu-hg7n. No persistent blocks changes (eu-hmji/eu-m59i) are included in the diff against `0.4.0`.

### Merge Conflict — Action Required

The PR cannot be merged cleanly. The `0.4.0` branch currently contains:

```
006f8f7 fix(eu-hg7n): implement Remapped sharing and fix u8 overflow for large frames
7412a27 fix(eu-hg7n): address Wicket review — GC safety, dead code, and test
0758b84 Revert "fix(eu-hg7n): implement Remapped sharing and fix u8 overflow for large frames"
63385e6 Revert "fix(eu-hg7n): address Wicket review — GC safety, dead code, and test"
94bc218 docs: add README for eucalypt-mode Emacs package
d8a38f2 feat: add juxtaposed definition syntax and fix tree-sitter grammar
```

The reverts of intermediate eu-hg7n work, combined with the Emacs README and juxtaposed-definitions commits, mean the PR branch conflicts with `0.4.0` in `env.rs`, `vm.rs`, `array.rs`, `089_sharing.eu`, and `harness_test.rs`.

**Please rebase `feat/furnace-shared-env` onto the current `0.4.0` head and force-push the branch.** The three eu-hg7n commits (`a3009ef`, `9cce368`, `0bec707`) should apply cleanly once the rebase drops the now-superseded intermediate commits that were reverted.

Once rebased, re-submit for a final gate check and I will merge immediately.

PR #367, opened 2026-03-06
fix(eu-dap7): suppress Update accumulation in IF tail recursion

gmorpheme:
## Summary

Replaces the dead `single_use_args` mechanism from commit 5e7b5ea (PR #365) with the correct runtime fix, ported from commit 46ead66 on \`feat/furnace-lazy-stream-next\`.

### Why 5e7b5ea was dead code

The \`single_use_args\` metadata was only consulted when the STG compiler saw \`Expr::Intrinsic\` as the call target. User-written \`if(c, t, e)\` compiles the function as \`Expr::Var("if")\`, which resolves to the IF global wrapper. The single_use_args on the IF intrinsic were never applied — branch args were still compiled as \`thunk\` (update=true), causing O(N) Update accumulation in tail-recursive conditionals.

### What this PR does (two commits)

**Commit 1: runtime suppress_update flag (the actual fix)**
- Adds \`suppress_update: bool\` to \`StgSyn::Case\`, \`HeapSyn::Case\`, \`Continuation::Branch\`
- Adds \`suppress_next_update: bool\` to \`MachineState\`
- New \`switch_suppress()\` DSL function — only the boolean IF wrapper uses it
- In \`return_data\`: when a \`suppress_update=true\` Branch fires and matched body is \`Atom { Ref::L(_) }\`, sets \`suppress_next_update=true\`
- In thunk entry (Ref::L): skips Update push when \`suppress_next_update\` is set

**Commit 2: remove dead single_use_args code**
- Removes \`single_use\` field from \`Intrinsic\`, \`single_use_args()\` method, and all 159 initialisers in the INTRINSICS table
- Removes \`single_use_args\` variable and \`is_single_use\` logic from \`ProtoAppGroup::take_syntax\`

### Stream memoisation safety

The original \`suppress_next_update\` flag (from b1eef84, removed in 3d8dd0a) fired for **all** case branches entering a local ref, breaking stream tail thunk memoisation. This fix is scoped to IF only: \`suppress_update\` is only set on the Case created by \`switch_suppress()\`, which only the IF wrapper uses. Stream tail thunks are never entered through a \`suppress_update=true\` Branch continuation.

## Benchmark

\`countdown(n): if(n = 0, 0, countdown(n - 1))\` with N=100,000:

| Metric | 0.4.0 HEAD (5e7b5ea broken) | This PR |
|--------|----------------------------|---------|
| Allocs | 33.5M | 700k |
| Improvement | — | ~48x |

## Test plan

- [x] All 187 harness tests pass (including test 100, IF tail recursion)
- [x] \`cargo clippy --all-targets -- -D warnings\` — clean
- [x] \`cargo fmt --all\` — clean
- [x] Benchmark: 700k allocs for N=100k countdown

### DIFF
--- a/examples/aoc25/day02.eu
+++ b/examples/aoc25/day02.eu
@@ -21,7 +21,7 @@ Notable eucalypt techniques:
 Running time: < 1s (both parts)
 "
 
-parse-range(s): str.split(s, "-") map(num)
+parse-range(s): s str.split-on("-") map(num)
 
 ` "Sum of invalid IDs in [a,b] that are a p-digit pattern repeated
 d/p times.  We find the range of n values whose IDs land in [a,b]
@@ -100,13 +100,13 @@ invalid-for-d(a, b, d):
 invalid-sum2(a, b): range(1, 12) map(invalid-for-d(a, b)) sum
 
 solve(data):
-  str.split(data first, ",")
+  data first str.split-on(",")
     map(parse-range)
     map(uncurry(invalid-sum))
     sum
 
 solve2(data):
-  str.split(data first, ",")
+  data first str.split-on(",")
     map(parse-range)
     map(uncurry(invalid-sum2))
     sum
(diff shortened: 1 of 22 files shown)


gmorpheme:
Superseded by the updated fix on the same branch. Closing in favour of retargeting #371 to 0.4.0.

PR #384, opened 2026-03-08
fix(eu-9vzc): deep-find takes symbol keys only

gmorpheme:
## Summary
- Corrects PR #381 which incorrectly accepted both symbols and strings
- `deep-find`, `deep-find-first`, `deep-find-paths` now take symbol keys only (`:key`)
- Matches the rest of the prelude convention (`has(:key)`, `lookup-or(:key, ...)`)
- Removes `sym(str.of(k))` conversion — just uses `k` directly since it's already a symbol
- Updates docs to remove "also accepts strings" language

## Test plan
- [ ] Existing harness tests pass
- [ ] `deep-find(:x, { a: { x: 1 } })` returns `[1]`

### DIFF
--- a/docs/reference/agent-reference.md
+++ b/docs/reference/agent-reference.md
@@ -448,11 +448,8 @@ merge({ a: 1 }, { b: 2 })    # { a: 1 b: 2 }
 
 #### `deep-find(k, b)` — find all values for key k at any depth
 
-Key `k` may be a symbol (preferred) or string.
-
 ```
 { a: { x: 1 } b: { x: 2 } } deep-find(:x)    # [1, 2]
-{ a: { x: 1 } b: { x: 2 } } deep-find("x")   # [1, 2] (also works)
 ```
 
 #### `deep-query(pattern, b)` — query with dot-separated glob pattern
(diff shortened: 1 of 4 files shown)


gmorpheme:
## Gate 1 Failure: CI test suite fails — breaking change breaks harness test 072

The test suite is failing on `test_harness_072` (`072_deep_find.eu`):

```
072_deep_find...FAIL
thread 'test_harness_072' panicked at tests/harness_test.rs:32:5:
assertion `left == right` failed
```

**Root cause**: `072_deep_find.eu` uses string keys for `deep-find` (e.g. `deep-find("a")`). PR #381 specifically preserved backwards compatibility with string keys via `sym(str.of(k))`. This PR removes that compatibility by reverting to `s: k`, which requires a symbol and fails on strings.

**Required before resubmission**: Update `tests/harness/072_deep_find.eu` to use symbol keys throughout (e.g. `deep-find(:a)` instead of `deep-find("a")`), and similarly `072_deep_find.eu`'s path tests if they also use strings. This will make the test reflect the new symbol-only contract.

Also verify that any doc examples that used `deep-find("..")` with string keys have been updated — the `docs/reference/prelude/blocks.md` example was already updated in earlier PRs, but check for any remaining string-key usages in doc examples tested by the CI doc-example runner.

gmorpheme:
**Gate 1 failure — CI test doc examples**

The Test Suite CI run failed on the doc examples check:

```
FAIL  docs/reference/prelude/supplements/blocks/deep-find-and-query.md:5
      FAIL (exit 1)
      error: type mismatch: expected symbol, found string
```

The file `docs/reference/prelude/supplements/blocks/deep-find-and-query.md` still uses string keys in its code examples:

```eu
hosts: config deep-find("host")  # ["localhost", "db.local"]
first-host: config deep-find-first("host", "unknown")  # "localhost"
```

Please update these to use symbol keys (`:host`) to match the new symbol-only behaviour. All other gates pass — this is the only outstanding issue.

PR #385, opened 2026-03-08
fix(eu-9vzc): revert deep-find to symbol-only keys

gmorpheme:
## Summary

- Reverts `sym(str.of(k))` back to plain `k` in `deep-find` and `deep-find-paths` — the argument is already a symbol; no conversion needed
- Updates inline doc strings to say "must be a symbol" (was "may be a symbol or string")
- Removes the "also works with strings (legacy)" example from `agent-reference.md` section 5.8 and replaces it with a clear symbol-only note (parallel to the `has` gotcha in section 5.7)

## Test plan

- [ ] `cargo build` passes (prelude embedded at compile time)
- [ ] `cargo clippy --all-targets -- -D warnings` clean
- [ ] Existing harness tests pass (deep-find callers in `tests/harness/` all use symbol keys)

### DIFF
--- a/docs/reference/agent-reference.md
+++ b/docs/reference/agent-reference.md
@@ -448,11 +448,10 @@ merge({ a: 1 }, { b: 2 })    # { a: 1 b: 2 }
 
 #### `deep-find(k, b)` — find all values for key k at any depth
 
-Key `k` may be a symbol (preferred) or string.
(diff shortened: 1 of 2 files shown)


gmorpheme:
Duplicate of #384 which has the same fix. Closing this one.

PR #386, opened 2026-03-08
fix(persistent-blocks): fix GC memory regression and 62 harness test failures

gmorpheme:
## Summary

- **GC regression fix**: Reverts `compile_block` from emitting `PBLOCK_FROM_PAIRS` back to `dsl::block(index)`. The previous approach made every block literal (including 110 in the prelude) a `HeapBlock` (`im_rc::OrdMap`). Since the GC never calls `Drop` on swept objects, OrdMap Rc-backed tree nodes accumulated on the Rust heap, causing 430–580% runtime regressions on GC-heavy benchmarks.
- **62 harness test failures fix**: Reverts `Meta` and `RawMeta` wrappers to use `KEmptyBlock.global()` as the no-metadata default instead of `PBlockEmpty.global()`. The previous code produced a `PersistentBlock` (tag=12) where a `Block` (tag=8) was expected, causing `NoBranchForDataTag` failures in all 62 tests using the `//=` assertion operator.

## Root causes

The original `feat/furnace-persistent-blocks-v2` commit (4965b3f) introduced two independent bugs:

1. `compiler.rs:compile_block` — routing all block literals through `PBLOCK_FROM_PAIRS` created unbounded OrdMap node leaks via the GC's non-dropping sweep
2. `meta.rs:Meta/RawMeta.wrapper` — using `PBlockEmpty` as the metadata absence default, which propagated `PersistentBlock` values into codepaths that only handle cons-list `Block`s (specifically `PBLOCK_MERGE.wrapper`'s inner `switch` which has no `Block` arm)

## Design decision

Block literals remain as standard cons-list `Block`s. Persistent blocks (`PersistentBlock`, tag=12) are only constructed explicitly when the user calls `PBLOCK_FROM_BLOCK`. This keeps the GC working correctly for the common case while still making persistent blocks available via intrinsic.

## Test results

- 595 unit tests: all pass
- 91 harness tests: all pass, 0 failures
- `cargo clippy --all-targets -- -D warnings`: clean
- `cargo fmt --all`: clean


gmorpheme:
Reverted compile_block() to cons-lists — this is not the fix we need. The actual issue is GC not calling Drop on HeapBlock/OrdMap. Closing.

PR #392, opened 2026-03-09
feat(eu-76sv): move NdArray dispatch from prelude to arithmetic intrinsics

gmorpheme:
## Summary

- Moves NdArray dispatch for `+`, `-`, `*`, `/` from `lib/prelude.eu` into the STG intrinsic `execute` methods in `src/eval/stg/arith.rs`
- Eliminates ~10x performance regression introduced by commit `016429c`, which added `is-array?` guards in the prelude causing every arithmetic operation to evaluate interpreted `if(is-array?(...))` checks even when no arrays are involved
- Reverts the four arithmetic operators in `lib/prelude.eu` back to direct `__ADD`, `__SUB`, `__MUL`, `__DIV` calls

## Changes

- `src/eval/stg/array.rs`: Made `array_binop` `pub(crate)` so `arith.rs` can call it
- `src/eval/stg/arith.rs`:
  - Added `arithmetic_wrapper` (using `unbox_any` + fallback, like `comparison_wrapper`) so NdArray atoms pass through to `execute` without triggering `NoBranchForNative`
  - Added NdArray dispatch at top of each `execute` method for Add, Sub, Mul, Div
  - Changed `machine_return_num` → `machine_return_boxed_num` (custom wrapper no longer boxes the result)
- `src/eval/stg/support.rs`:
  - Added `machine_return_boxed_num` helper
  - Added `cons_tag_of` helper for inspecting Cons tags on unresolved refs
  - Updated `num_arg` to produce `NoBranchForDataTag` (preserving error messages like "blocks cannot be used in arithmetic") when passed a non-numeric data constructor
- `lib/prelude.eu`: Reverted `+`, `-`, `*`, `/` to direct intrinsic calls

## Test plan

- [ ] All 188 harness tests pass (`cargo test`)
- [ ] Test 087 (NdArray operations) passes
- [ ] All error tests pass (031, 045, 053, 078 — type mismatch messages preserved)
- [ ] `cargo clippy --all-targets -- -D warnings` clean
- [ ] AoC day03 benchmark completes well within 60s (was timing out pre-fix)

### DIFF
--- a/lib/prelude.eu
+++ b/lib/prelude.eu
@@ -365,13 +365,13 @@ or: __OR
     export: :suppress
     associates: :left
     precedence: :sum }
-(l + r): if(is-array?(l) || is-array?(r), arr.add(l, r), __ADD(l, r))
+(l + r): __ADD(l, r)
 
 ` { doc: "`l - r` - subtracts `r` from `l`. For numbers, both must be numeric. For arrays, performs element-wise subtraction; one operand may be a scalar."
     export: :suppress
     associates: :left
     precedence: :sum }
-(l - r): if(is-array?(l) || is-array?(r), arr.sub(l, r), __SUB(l, r))
+(l - r): __SUB(l, r)
 
 ` { doc: "`l ^ r` - raise `l` to the power `r`."
     export: :suppress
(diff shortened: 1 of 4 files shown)


gmorpheme:
Superseded by #393 which fixes worktree contamination and test_unboxed_add.

PR #409, opened 2026-03-10
fix: IO monad dot-chain and bare-expression desugaring

gmorpheme:
## Summary

- **Bug 1 fixed**: `{ :io r: cmd }.r.stdout` now correctly desugars to `io.bind(cmd, λ(r). io.return(r.stdout))`. Previously only the first element after `}` was consumed into the return expression; subsequent `.field` accesses were applied outside `io.return`, producing a type error.
- **Bug 2 fixed**: A `.eu` file containing only a monadic expression (no outer `key:` declaration) now correctly evaluates as the programme body. Previously the expression was parsed as metadata with no body, producing "unresolved variable 'io'".
- **New harness tests**: 106 (`io_block_chain`) and 107 (`io_dot_chain`) covering both bugs.
- **Documentation**: Added "IO Monad Return Expression Chaining" section to `docs/appendices/syntax-gotchas.md`.

## Test plan

- [x] `cargo test --test harness_test test_harness_106` passes
- [x] `cargo test --test harness_test test_harness_107` passes
- [x] All 97 harness tests pass (`cargo test --test harness_test`)
- [x] All 596 lib tests pass (`cargo test --lib`)

### DIFF
--- a/docs/appendices/syntax-gotchas.md
+++ b/docs/appendices/syntax-gotchas.md
@@ -310,6 +310,46 @@ first: xs head    # = 1
 rest: xs tail     # = [2, 3]
 ```
 
+## IO Monad Return Expression Chaining
+
+When using monadic block syntax `{ :io r: cmd }.return_expr`, the return expression
+may include dot-chained field access:
+
+```eu
+# Dot-chained field access in return expression
+{ :io r: io.shell("echo hello") }.r.stdout
+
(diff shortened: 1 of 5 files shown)


gmorpheme:
Closing in favour of revised fix that tightens the bare-expression detection to avoid regression in test_error_043.

PR #410, opened 2026-03-10
fix: IO monad dot-chain and bare-expression desugaring (v2)

gmorpheme:
## Summary

Supersedes #409 (closed). The bare-expression detection is now tightened to avoid the test_error_043 regression.

- **Bug 1 fixed**: `{ :io r: cmd }.r.stdout` correctly desugars to `io.bind(cmd, λ(r). io.return(r.stdout))`. Previously only the first element after `}` was consumed into the return expression; subsequent `.field` accesses were applied outside `io.return`, causing a type error.
- **Bug 2 fixed**: A `.eu` file containing only a monadic block expression (no `key:` declaration) now evaluates correctly as the programme body. Previously the expression was parsed as metadata with no body, producing "unresolved variable 'io'".
- **Regression fixed vs #409**: The bare-expression path now only fires when the raw metadata soup begins with a `Block`, `BracketBlock`, or `ParenExpr` element. Assignment-style mistakes like `result = 42` (first element: `Name`) are not matched and still produce the expected "eucalypt uses ':' for declarations" error.

## Changes

- `src/core/desugar/rowan_ast.rs`: Added `unit_meta_starts_with_block` helper; tightened `is_bare_expression` check; added `desugar_return_chain` helper for dot-chained return expressions; extended `desugar_monadic_block` to accept return expression slices.
- `tests/harness/106_io_block_chain.eu`: New test — monadic block with parenthesised return expression.
- `tests/harness/107_io_dot_chain.eu`: New test — dot-chained field access in return expression.
- `tests/harness_test.rs`: Registered tests 106 and 107.
- `docs/appendices/syntax-gotchas.md`: Added "IO Monad Return Expression Chaining" section.

## Test plan

- [x] `cargo test --test harness_test test_error_043` passes (regression check)
- [x] `cargo test --test harness_test test_harness_106` passes
- [x] `cargo test --test harness_test test_harness_107` passes
- [x] All 196 harness tests pass (`cargo test --test harness_test`)

### DIFF
--- a/docs/appendices/syntax-gotchas.md
+++ b/docs/appendices/syntax-gotchas.md
@@ -310,6 +310,46 @@ first: xs head    # = 1
 rest: xs tail     # = [2, 3]
 ```
 
+## IO Monad Return Expression Chaining
+
+When using monadic block syntax `{ :io r: cmd }.return_expr`, the return expression
+may include dot-chained field access:
+
+```eu
+# Dot-chained field access in return expression
+{ :io r: io.shell("echo hello") }.r.stdout
+
+# This desugars to:
+io.bind(io.shell("echo hello"), λ(r). io.return(r.stdout))
+```
+
+**Gotcha**: Previously, only the first element after the closing `}` was treated as the
+return expression. Subsequent `.field` accesses were left outside the `io.return` wrapper
+and applied to the IO action value — causing a type error.
+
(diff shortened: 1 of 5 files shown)


gmorpheme:
Closing in favour of v3 PR. Dropped the desugar_return_chain dot-chain code (caused GC misalignment on Linux). Keeping only the bare-expression fix and the parenthesised monadic block test (106).

PR #412, opened 2026-03-10
fix: bare-expression desugaring for monadic unit files

gmorpheme:
## Summary

Supersedes #409 and #410 (both closed). Minimal, focused fix with no dot-chain desugaring.

- **Bug fixed**: A `.eu` file containing only a monadic block expression (no outer `key:` declaration) now evaluates correctly as the programme body. Previously such files were parsed as metadata with no body, producing "unresolved variable 'io'".
- **Scope reduced vs previous PRs**: Dropped `desugar_return_chain` and test 107 (dot-chained field access). The dot-chain form `{ :io r: cmd }.r.stdout` is tracked separately as a design question. Users should use the parenthesised form `{ :io r: cmd }.(r.stdout)` or an extra binding.
- **test 106**: Added `requires-io: true` metadata so the test is skipped gracefully in environments without IO permissions.

## Changes

- `src/core/desugar/rowan_ast.rs`: Added `unit_meta_starts_with_block` helper; bare-expression path in `Unit::desugar` now only fires when the raw metadata soup starts with `Block`/`BracketBlock`/`ParenExpr`, preventing regression on `result = 42`.
- `tests/harness/106_io_block_chain.eu`: New test — monadic block with parenthesised return expression. Has `requires-io: true` so it skips in restricted CI environments.
- `tests/harness_test.rs`: Registered test 106.
- `docs/appendices/syntax-gotchas.md`: Added "IO Monad Block Syntax" section documenting the supported forms and the dot-chain limitation.

## Test plan

- [x] `cargo test --test harness_test test_error_043 --release` passes (regression check)
- [x] `cargo test --test harness_test test_harness_106 --release` passes
- [x] All 195 harness tests pass (`cargo test --test harness_test --release`)
- [x] All 596 lib tests pass (`cargo test --lib --release`)

### DIFF
--- a/docs/appendices/syntax-gotchas.md
+++ b/docs/appendices/syntax-gotchas.md
@@ -310,6 +310,47 @@ first: xs head    # = 1
 rest: xs tail     # = [2, 3]
 ```
 
+## IO Monad Block Syntax
+
+When using monadic block syntax `{ :io r: cmd }.(return_expr)`, the return
+expression must be parenthesised or a single name:
+
+```eu
+# Parenthesised return expression (recommended for complex expressions)
+{ :io r: io.shell("echo hello") }.(
+  if(r.stdout str.matches?("hello.*"), :PASS, :FAIL))
+
+# Single-name return — accesses the bound variable directly
+{ :io r: io.shell("echo hello") }.r
+```
+
+**Desugaring**: `{ :io r: cmd }.(expr)` desugars to:
+
+```eu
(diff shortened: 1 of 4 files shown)


gmorpheme:
Closing in favour of v4 which includes the dot-chain desugaring fix. GC investigation revealed the crash was with specific nested IO test content; the dot-chain desugaring itself is correct.

PR #422, opened 2026-03-10
fix(io): fix io.shell panic and shell-with spec block unreadable (eu-vc32)

gmorpheme:
## Summary

- **Bug 1 (spec block unreadable)**: `shell-with` and `exec-with` previously called helper functions `__io-shell-spec`/`__io-exec-spec` to build spec blocks. These produced `App` thunks in the STG closure. The io-run driver's `ReadSpecBlock` does static heap navigation and cannot follow `App` nodes, producing "IoAction spec block is not a Block constructor". Fix: inline block literals directly so the spec is a literal in the closure environment.
- **Bug 2 (symbol pool OOB panic)**: `BuildResultBlock` interns "stdout", "stderr", "exit-code" into a *cloned* `SymbolPool` passed as input. The machine's own pool is never updated, so `pool.resolve(id)` panics (index out of bounds) when the machine later accesses the result block. Fix: add `Machine::intern_symbol()` and pre-intern the three keys before cloning the pool for `BuildResultBlock`.
- **Documentation**: Added limitation note to `docs/reference/prelude/io.md` — the `timeout`/`stdin` fields in opts are parsed but not yet applied by the static spec reader; full fix requires driver evaluation (deferred to Furnace).

## Test plan

- [ ] `cargo test --lib` — all 596 lib tests pass
- [ ] `cargo test --test harness_test` — all 206 harness tests pass
- [ ] `cargo clippy --all-targets -- -D warnings` — clean
- [ ] `cargo fmt --all -- --check` — clean
- [ ] Manual: `eu -I -e 'io.shell("echo hello") io.map(.stdout)'` no longer panics
- [ ] Manual: `eu -I -e 'io.shell-with({}, "echo hello world") io.map(.stdout)'` works

### DIFF
--- a/docs/reference/prelude/io.md
+++ b/docs/reference/prelude/io.md
@@ -53,6 +53,11 @@ IO operations require the `--allow-io` / `-I` flag at the command line.
 Default timeout is 30 seconds. Override with `{timeout: N}` in `opts`.
 Optional `{stdin: s}` pipes string `s` to the command's standard input.
 
+**Current limitation**: The `timeout` and `stdin` fields in `opts` for
+`shell-with` and `exec-with` are accepted by the parser but not yet applied
+by the io-run driver (they default to 30 s and no stdin). This will be fixed
+in a future driver update. The `cmd` argument is always applied correctly.
+
 ### Combinators
 
 | Function | Description |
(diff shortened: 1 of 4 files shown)


gmorpheme:
Superseded by PRs #423, #426, and subsequent fixes. eu-vc32 is closed.

PR #452, opened 2026-03-10
fix(errors): remove 'panic:' prefix from IO-not-allowed error

gmorpheme:
### Scenario
A user writes code using `io.shell()` or any IO operation, but forgets to pass
the `--allow-io` (`-I`) flag to the `eu` command. Previously the error was:

### Before
```
error: panic: IO operations require the --allow-io (-I) flag
```

The `panic:` prefix is actively misleading — this is a user configuration issue,
not a programming error or runtime panic.

### After
```
error: IO operations require the --allow-io (-I) flag
  help: re-run with 'eu -I ...' to enable side effects
```

### Assessment
- Human diagnosability: fair → excellent
- LLM diagnosability: good → excellent

### Change
- Added `ExecutionError::IoNotAllowed` variant to the error enum with an
  appropriate message and help text.
- Added `From<IoRunError> for ExecutionError` in `io_run.rs` so that
  `IoRunError::IoNotAllowed` maps directly to `ExecutionError::IoNotAllowed`
  rather than being wrapped in `Panic`. Other `IoRunError` variants that are
  not internal invariant violations can be extended to use this pattern later.
- Updated the two `map_err` call sites in `eval.rs` that convert
  `IoRunError → ExecutionError` to use `.map_err(ExecutionError::from)`.

### Risks
- The `From<IoRunError>` implementation preserves the existing behaviour for all
  `IoRunError` variants other than `IoNotAllowed` and `MachineError` — they fall
  through to `Panic(other.to_string())`.
- All 90 error tests pass; `cargo clippy --all-targets -- -D warnings` is clean.

### DIFF
--- a/src/driver/eval.rs
+++ b/src/driver/eval.rs
@@ -234,7 +234,7 @@ impl<'a> Executor<'a> {
                     if ret.is_ok() {
                         if machine.io_yielded() {
                             let io_result = io_run_and_render(&mut machine, opt.allow_io)
-                                .map_err(|e| ExecutionError::Panic(e.to_string()));
+                                .map_err(ExecutionError::from);
                             machine.take_emitter().stream_end();
                             return io_result;
                         }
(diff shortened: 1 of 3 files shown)


gmorpheme:
Superseded by #466 which handles all IoRunError variants more comprehensively including IoFail, IoTimeout, and IoCommandError with references to the public io.* functions.

PR #453, opened 2026-03-10
fix(errors): improve base64 decode error messages

gmorpheme:
### Scenario
A user calls `str.base64-decode` (or the underlying `__BASE64_DECODE`
intrinsic) with a string that is not valid base64. The underlying `base64`
crate returns structured error information including the offending byte code,
but this was previously formatted using the default `Display` which shows
numeric ASCII codes.

### Before
```
error: panic: invalid base64 input: Invalid symbol 33, offset 3.
```

Two problems:
1. The `panic:` prefix is misleading.
2. "Invalid symbol 33" requires the user to look up ASCII code 33 to discover
   the offending character is `!`.

### After
```
error: invalid base64 input: invalid character '!' at offset 3
```

For other error cases:

```
error: invalid base64 input: incorrect padding — base64 strings must be
       padded with '=' to a multiple of 4 characters
```

### Assessment

### Change
- Added `ExecutionError::Base64DecodeError(Smid, String)` to the error enum.
- Added `format_base64_error()` helper in `encoding.rs` that:
  - Converts `DecodeError::InvalidByte(offset, byte)` to `"invalid character 'X' at offset N"` for printable bytes, or `"invalid byte 0xNN at offset N"` for non-printable.
  - Provides plain English descriptions for `InvalidPadding`, `InvalidLength`, and `InvalidLastSymbol`.
- Updated `Base64Decode::execute()` to use this helper.

### Risks
- All 90 error tests pass; `cargo clippy --all-targets -- -D warnings` is clean.

### DIFF
--- a/src/eval/error.rs
+++ b/src/eval/error.rs
@@ -639,6 +639,8 @@ pub enum ExecutionError {
     Panic(String),
     #[error("assertion failed: expected {2}, got {1}")]
     AssertionFailed(Smid, String, String),
+    #[error("invalid base64 input: {1}")]
+    Base64DecodeError(Smid, String),
     #[error("machine did not terminate after {0} steps")]
     DidntTerminate(usize),
     #[error("infinite loop detected: binding refers to itself")]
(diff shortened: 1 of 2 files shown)


gmorpheme:
Superseded by #476, which uses separate structured variants `InvalidBase64(Smid, String)` and `InvalidBase64Utf8(Smid, String)` and is more complete.

PR #454, opened 2026-03-10
fix(errors): remove 'panic:' from invalid semver constraint error

gmorpheme:
### Scenario
A user writes `eu.requires("not-a-version")` using a string that is not a
valid semver constraint. The underlying `semver` crate's parse error was
previously wrapped in `ExecutionError::Panic`.

### Before
```
error: panic: invalid version constraint "not-a-semver-constraint": unexpected character 'n' while parsing major version number
```

### After
```
error: invalid version constraint 'not-a-semver-constraint': unexpected character 'n' while parsing major version number
  = semver constraints use operators like '^', '~', '>=', e.g. '^0.5' or '>=0.4, <0.6'
```

### Assessment

### Change
- Added `ExecutionError::InvalidVersionConstraint(Smid, String, String)` variant — (location, constraint, parse error message).
- Updated `Requires::execute()` to emit the structured variant instead of `Panic`.
- Added a diagnostic note with a semver syntax example to guide users towards valid constraint syntax.

### Risks
- No existing tests for this error path (the test for `eu.requires` covers the version-mismatch case, not the parse-error case).
- All 90 existing error tests pass; `cargo clippy --all-targets -- -D warnings` is clean.

### DIFF
--- a/src/eval/error.rs
+++ b/src/eval/error.rs
@@ -520,6 +520,8 @@ pub enum ExecutionError {
     ParseError(Smid, String, String),
     #[error("version requirement not satisfied: eucalypt {1} does not satisfy '{2}'")]
     VersionRequirementFailed(Smid, String, String),
+    #[error("invalid version constraint '{1}': {2}")]
+    InvalidVersionConstraint(Smid, String, String),
     #[error("assertion failed: expected {2}, got {1}")]
     AssertionFailed(Smid, String, String),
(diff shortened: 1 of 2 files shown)


gmorpheme:
Superseded by #511 (`review/clarion-version-constraint-error`), which replaces Panic with a proper structured variant `VersionRequirementFailed(Smid, String, String)`. PR #454 only removed the panic prefix from the string message.

PR #457, opened 2026-03-11
fix(errors): structured error for mixed-type ordered comparisons

gmorpheme:
### Scenario
Using an ordered comparison operator (`>`, `>=`, `<`, `<=`) with operands of different types, e.g. `1 > "hello"` or `"abc" < 42`.

### Before
```
error: panic: cannot compare values with GT: operands must be the same type (both numbers, strings, symbols, or datetimes)
  ┌─ test.eu:1:6
```

### After
```
error: cannot compare mixed types with '>': both operands must be the same type (numbers, strings, symbols, or datetimes)
  ┌─ test.eu:1:6
  │
1 │ x: 1 > "hello"
  │      ^
  │
  = ordered comparisons (<, >, <=, >=) require both operands to be the same type — you cannot compare a number to a string, for example
  = use 'str' to convert a number to a string before comparing, or 'num' to convert a string to a number
```

### Assessment

### Change
- Added `ComparisonTypeMismatch(Smid, String)` variant to `ExecutionError` — stores source location and the operator symbol
- Added source location extraction in `HasSmid` impl
- Added diagnostic notes explaining valid usage and conversion options
- Replaced `Panic(format!(...))` in `ordered_cmp()` in `src/eval/stg/arith.rs`
- Operator names now use symbols (`>`, `>=`, `<`, `<=`) rather than internal intrinsic names (`GT`, `GTE`, `LT`, `LTE`)

### Risks
None — the error was already returned for these inputs; only the message and location tracking changed.

### DIFF
--- a/src/eval/error.rs
+++ b/src/eval/error.rs
@@ -496,6 +496,8 @@ pub enum ExecutionError {
     BadNumberFormat(String),
     #[error("could not format {1} with {0}")]
     FormatError(String, Number),
+    #[error("cannot compare mixed types with '{1}': both operands must be the same type (numbers, strings, symbols, or datetimes)")]
+    ComparisonTypeMismatch(Smid, String),
     #[error("expected scalar value")]
     NotScalar(Smid),
     #[error("{}", format_unknown_format(.0))]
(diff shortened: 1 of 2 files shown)


gmorpheme:
Closing as duplicate of #449 which already covers this improvement.

PR #462, opened 2026-03-11
fix(errors): structured variants for base64 decode errors

gmorpheme:
### Scenario
A user calls `str.base64-decode` on a string that is not valid standard base64 (e.g. contains URL-safe `-` characters or other invalid symbols).

### Before
```
error: panic: invalid base64 input: Invalid symbol 45, offset 3.
  ┌─ /tmp/test-base64-decode.eu:1:10
  │
1 │ decoded: str.base64-decode("not-valid-base64!!!")
  │          ^^^
  │
  = stack trace:
    - ==
```

### After
```
error: invalid base64 input: Invalid symbol 45, offset 3.
  help: 'not-valid-base64!!!' is not valid standard base64 (RFC 4648); ensure the input uses the standard alphabet (A–Z, a–z, 0–9, +, /) with '=' padding
  ┌─ /tmp/test-base64-decode.eu:1:10
  │
1 │ decoded: str.base64-decode("not-valid-base64!!!")
  │          ^^^
  │
  = stack trace:
    - ==
```

### Assessment

The "panic:" prefix was misleading — invalid base64 input is a user error, not an internal invariant violation. The new message removes the prefix and adds a hint explaining the standard base64 alphabet and padding requirements. Long input strings are truncated in the preview (at 40 characters) to avoid flooding the terminal.

Two new structured variants added:
- `InvalidBase64(Smid, String, String)` — invalid base64 encoding
- `InvalidBase64Utf8(Smid, String)` — decoded bytes not valid UTF-8

Both carry a `Smid` for source location propagation.

### Change
- `src/eval/error.rs`: Added `InvalidBase64` and `InvalidBase64Utf8` variants; added `format_invalid_base64()` helper; added both to `HasSmid` match
- `src/eval/stg/encoding.rs`: `Base64Decode::execute()` now uses structured variants; `machine.annotation()` obtained before `str_arg` to capture source location

### Risks
Low. No existing harness tests cover base64 decode errors, so no `.expect` sidecars to update. The change only affects error output format, not behaviour.

### DIFF
--- a/src/eval/error.rs
+++ b/src/eval/error.rs
@@ -395,6 +395,21 @@ fn format_cannot_return_fun(expected_tags: &[u8]) -> String {
     msg
 }
 
+/// Format an invalid base64 error message, truncating long input strings
+fn format_invalid_base64(input: &str, detail: &str) -> String {
+    const PREVIEW_LEN: usize = 40;
+    let preview = if input.len() > PREVIEW_LEN {
+        format!("{}…", &input[..PREVIEW_LEN])
+    } else {
+        input.to_string()
+    };
+    format!(
+        "invalid base64 input: {detail}\n  \
+         help: '{preview}' is not valid standard base64 (RFC 4648); \
+         ensure the input uses the standard alphabet (A\u{2013}Z, a\u{2013}z, 0\u{2013}9, +, /) with '=' padding"
+    )
+}
+
 /// Format a division by zero error message with the operation context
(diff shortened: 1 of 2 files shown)


gmorpheme:
Superseded by #476, which uses `InvalidBase64(Smid, String)` and `InvalidBase64Utf8(Smid, String)` variants. PR #476 was rebased on master and is more complete.

PR #463, opened 2026-03-11
fix(errors): structured variant for invalid version constraint

gmorpheme:
### Scenario
A user calls `eu.requires("not a semver !!!@#")` with a string that is not valid semver syntax.

### Before
```
error: panic: invalid version constraint "not a semver !!!@#": unexpected character 'n' while parsing major version number
 = stack trace:
   - ==
```

### After
```
error: invalid version constraint 'not a semver !!!@#': unexpected character 'n' while parsing major version number
  help: version constraints use semver syntax, e.g. '>=0.2.0', '~0.5', '^1.2'
 = stack trace:
   - ==
```

### Assessment

The "panic:" prefix was misleading — an invalid semver constraint is a user error, not an internal invariant violation. The new message removes the prefix and adds semver syntax examples to guide correction.

### Change
- `src/eval/error.rs`: Added `InvalidVersionConstraint(Smid, String, String)` variant; added to `HasSmid` match
- `src/eval/stg/version.rs`: `Requires::execute()` now uses `InvalidVersionConstraint` for parse errors; obtains `smid` before `str_arg` to capture source location

### Risks
Low. No existing harness tests cover invalid version constraint errors. The Cargo version parse failure (line 38–42 of `version.rs`) remains as `Panic` since it represents an internal invariant (our own version string should always be valid semver).

### DIFF
--- a/src/eval/error.rs
+++ b/src/eval/error.rs
@@ -516,6 +516,8 @@ pub enum ExecutionError {
     CannotReturnFunToCase(Smid, Vec<u8>),
     #[error("panic: {0}")]
     Panic(String),
+    #[error("invalid version constraint '{1}': {2}\n  help: version constraints use semver syntax, e.g. '>=0.2.0', '~0.5', '^1.2'")]
+    InvalidVersionConstraint(Smid, String, String),
     #[error("parse-as({1}): {2}")]
     ParseError(Smid, String, String),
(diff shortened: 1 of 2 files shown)


gmorpheme:
Superseded by #511 (`review/clarion-version-constraint-error`), which uses a more accurate variant name `VersionRequirementFailed` and has been rebased on master.

PR #466, opened 2026-03-11
fix(errors): propagate IoRunError as structured ExecutionError variants

gmorpheme:
### Scenario
When IO actions are used without the `--allow-io` flag, or when IO operations fail, the errors were previously wrapped in `ExecutionError::Panic` losing all structure and source location.

### Before (IoNotAllowed example)
```
error: panic: IO operations require the --allow-io (-I) flag
```

### After (IoNotAllowed)
```
error: IO operations are not permitted; use the --allow-io (-I) flag to enable
  ┌── input.eu:3:9 ───
  │ result: io.shell("echo hello")
  │         ^^^^^^^^
```

### After (IoTimeout example)
```
error: io.shell-with: command timed out after 30 seconds
  ┌── input.eu:5:3 ───
  │ result: io.shell-with({timeout: 30, cmd: "sleep 60"})
  │         ^^^^^^^^^^^^^^
```

### Assessment

The "panic:" prefix is removed, error messages use user-facing function names (`io.shell-with`) rather than internal tags (`:io-shell`), and source locations are propagated via `Smid`.

### Changes
- `src/eval/error.rs`: Added `IoNotAllowed(Smid)`, `IoFail(Smid, String)`, `IoTimeout(Smid, u64)`, and `IoCommandError(Smid, String)` variants to `ExecutionError`, with `smid()` support for source location extraction
- `src/eval/machine/vm.rs`: Exposed `Machine::annotation()` so the io-run driver can capture the current source annotation
- `src/driver/io_run.rs`: `IoRunError` variants now carry `Smid` — `IoNotAllowed(Smid)`, `Timeout(Smid, u64)`, `CommandError(Smid, String)`. The annotation is captured from the machine at the `IoAction` call site and mapped onto errors from `run_spec()`
- `src/driver/eval.rs`: `io_run_error_to_execution()` propagates the `Smid` from each `IoRunError` variant into the corresponding `ExecutionError`
- `tests/harness/errors/094_io_no_flag.eu.expect`: Updated regex to match new message wording

### Test
- `cargo test test_error_094` — verifies `IoNotAllowed` error with exit code 1
- All 96 error harness tests pass

### Risks
Low. `UnknownActionTag` and `MalformedSpec` IoRunError variants still go through the `Panic` fallback since they represent internal states not typically seen by users.

### DIFF
--- a/src/driver/eval.rs
+++ b/src/driver/eval.rs
@@ -35,10 +35,10 @@ use super::statistics::Statistics;
 /// for user-facing errors and `Panic` for internal machine errors.
 fn io_run_error_to_execution(e: IoRunError) -> ExecutionError {
     match e {
-        IoRunError::IoNotAllowed => ExecutionError::IoNotAllowed,
-        IoRunError::Fail(msg) => ExecutionError::IoFail(msg),
-        IoRunError::Timeout(secs) => ExecutionError::IoTimeout(secs),
-        IoRunError::CommandError(msg) => ExecutionError::IoCommandError(msg),
+        IoRunError::IoNotAllowed(smid) => ExecutionError::IoNotAllowed(smid),
+        IoRunError::Fail(msg) => ExecutionError::IoFail(Smid::default(), msg),
+        IoRunError::Timeout(smid, secs) => ExecutionError::IoTimeout(smid, secs),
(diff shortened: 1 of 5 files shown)


gmorpheme:
This references an internal format. Reference io-shell-with and similar instead.

PR #475, opened 2026-03-11
fix(errors): add type mismatch notes for symbol, null, and unit values

gmorpheme:
### Scenario
Several common programming mistakes produce type mismatch errors with no
diagnostic guidance. This PR adds contextual notes for:

1. **Symbol in arithmetic**: `:foo + 1`
2. **Null in arithmetic**: `null + 1`
3. **Null in field access**: `null.field`
4. **Symbol in string context** (via TypeMismatch)
5. **Null in string context**

### Before

```
error: type mismatch: expected number, found symbol

error: type mismatch: expected number, found null
```
(No notes whatsoever.)

### After

Symbol in arithmetic:
```
error: type mismatch: expected number, found symbol
  = symbol values (e.g. ':foo') cannot be used in arithmetic
  = to get the string form of a symbol, use 'str', e.g. ':foo str' returns "foo"
  = to convert a symbol's string form to a number, use ':foo str num'
```

Null in arithmetic:
```
error: type mismatch: expected number, found null
  = null cannot be used in arithmetic — it represents the absence of a value
  = if the value may be null, use 'if(x = null, default, x)' to provide a fallback
    before using it in arithmetic
```

Null in field access:
```
error: type mismatch: expected block, found null
  = null cannot be used with the '.' field-access operator — it has no fields
  = if the value may be null, use 'if(x = null, default_block, x).field' to guard
    against null before field access
```

### Assessment

### Change
Added five new arms to `type_mismatch_notes()` covering `(Number, Symbol)`,
`(Number, Unit)`, `(String, Symbol)`, and `(String, Unit)` pairs.

Added three new cases to `data_tag_mismatch_notes()` for `is_unit` values
(null) expected in number, block, or string positions — these cases arise
when null propagates through arithmetic or block-navigation operations.

### Risks
Low. Only adds new notes; does not change error variant types, exit codes,
or the existing test expectations. All 90 harness error tests pass.

### DIFF
--- a/src/eval/error.rs
+++ b/src/eval/error.rs
@@ -63,6 +63,27 @@ fn type_mismatch_notes(expected: &IntrinsicType, actual: &IntrinsicType) -> Vec<
              symbols are written with a leading colon"
                 .to_string(),
         ],
+        (Number, Symbol) => vec![
+            "symbol values (e.g. ':foo') cannot be used in arithmetic".to_string(),
+            "to get the string form of a symbol, use 'str', e.g. ':foo str' returns \"foo\""
+                .to_string(),
+            "to convert a symbol's string form to a number, use ':foo str num'".to_string(),
+        ],
+        (Number, Unit) => vec![
+            "null cannot be used in arithmetic — it represents the absence of a value".to_string(),
+            "if the value may be null, use 'if(x = null, default, x)' to provide a fallback \
+             before using it in arithmetic"
+                .to_string(),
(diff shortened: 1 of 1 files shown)


gmorpheme:
Need source locations and better advice.

gmorpheme:
Closing per owner direction: this PR adds only hint text for symbol/null/unit type mismatches, without adding source location. Source location propagation is more valuable.

PR #476, opened 2026-03-11
fix(errors): improve base64 decode error messages with character context

gmorpheme:
### Scenario
A user tries to decode a string that is not valid base64:

```eu
result: "not-valid-base64!!!" str.base64-decode
```

Or a string with wrong length (not a multiple of 4):

```eu
result: "abc" str.base64-decode
```

### Before
```
error: panic: invalid base64 input: Invalid symbol 45, offset 3.
```
(Byte 45 is the '-' character. The user cannot reasonably know this without
consulting an ASCII table.)

```
error: panic: invalid base64 input: Invalid padding
```

### After
```
error: panic: str.base64-decode: invalid character '-' at position 3; base64
  (standard) uses only A-Z, a-z, 0-9, +, /, and = for padding
```

```
error: panic: str.base64-decode: invalid padding in base64 string; standard
  base64 requires '=' padding to make the length a multiple of 4
```

### Assessment

The old message included a raw byte value (45) that is meaningless to most
users. The new message names the offending character directly, explains the
valid character set, and covers each decode error variant with a specific
actionable explanation.

### Change
Replaced the two `map_err` closures in `BASE64_DECODE::execute()` with a
match on `base64::DecodeError` variants. Each variant now produces a message
that identifies the invalid character (for `InvalidByte` and
`InvalidLastSymbol`), explains padding requirements (for `InvalidPadding`),
or notes the length constraint (for `InvalidLength`). All messages are
prefixed with `str.base64-decode:` to identify the call site.

### Risks
Low. Only improves the human-readable error text. No test expectations
reference base64 error messages, so no harness tests need updating.

### DIFF
--- a/src/eval/error.rs
+++ b/src/eval/error.rs
@@ -639,6 +639,10 @@ pub enum ExecutionError {
     IoTimeout(Smid, u64),
     #[error("io.shell-with: command execution error: {1}")]
     IoCommandError(Smid, String),
+    #[error("str.base64-decode: {1}")]
+    InvalidBase64(Smid, String),
+    #[error("str.base64-decode: decoded bytes are not valid UTF-8: {1}")]
+    InvalidBase64Utf8(Smid, String),
     #[error("assertion failed: expected {2}, got {1}")]
     AssertionFailed(Smid, String, String),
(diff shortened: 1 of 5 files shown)


gmorpheme:
Why no source location?

PR #484, opened 2026-03-11
fix(errors): improve array and null-access error messages

gmorpheme:
### Scenarios covered

**Array operations:**
1. `arr.reshape([3, 4])` on a 2×3 array — cannot reshape 6 into 12 elements
2. `arr.slice(arr, 5, 0)` on a rank-2 array — axis out of bounds
3. `arr.add(a, b)` on arrays of different shapes — element-wise shape mismatch

**Null access:**
4. `x.field` where `x` is null — field access on null
5. `null + 5` — null in arithmetic
6. `"{null}"` — null in string interpolation

### Before
```
error: panic: reshape: new shape does not match total element count
error: panic: slice: axis or index out of bounds
error: panic: array shape mismatch in element-wise operation
error: type mismatch: expected block, found null
error: type mismatch: expected number, found null
```

### After
```
error: panic: arr.reshape: cannot reshape array of 6 element(s) (shape [2, 3]) into shape [3, 4] which requires 12 element(s)
error: panic: arr.slice: axis 5 or index 0 is out of bounds for array with shape [2, 3] (rank 2); axis must be 0..2 ...
error: panic: array shape mismatch: cannot apply element-wise operation to arrays with shapes [2, 3] and [2, 2]; ...
error: type mismatch: expected block, found null
  = cannot access a field on null — the value is null, not a block
  = if this value may be null, check with 'null?' before accessing fields, ...
  = if this comes from a lookup that might fail, consider 'lookup-or' for a default, ...
error: type mismatch: expected number, found null
  = null cannot be used in arithmetic — it represents the absence of a value
  = if the value may be null, use 'if(x null?, 0, x)' to provide a fallback ...
```

### Assessment
- Human diagnosability: poor → good/excellent

### Changes
- `src/eval/stg/array.rs`: Improved `ArrayReshape`, `ArraySlice`, and `array_binop` panic messages to include actual shapes, counts, and axis/index values
- `src/eval/error.rs`: Added `(Record(_), Unit)` branch in `type_mismatch_notes` and `is_unit` branches in `data_tag_mismatch_notes` for null-access cases
- `tests/harness/errors/082_array_reshape_mismatch.eu.expect`: Updated regex to match new message format
- `tests/harness/errors/083_array_arith_mismatch.eu.expect`: Updated regex to match new message format

### Risks
Low. All 90 error tests pass. Clippy clean.

### DIFF
--- a/src/eval/error.rs
+++ b/src/eval/error.rs
@@ -63,6 +63,15 @@ fn type_mismatch_notes(expected: &IntrinsicType, actual: &IntrinsicType) -> Vec<
              symbols are written with a leading colon"
                 .to_string(),
         ],
+        (Record(_), Unit) => vec![
+            "cannot access a field on null — the value is null, not a block".to_string(),
+            "if this value may be null, check with 'null?' before accessing fields, \
+             e.g. 'if(x null?, default, x.field)'"
+                .to_string(),
+            "if this comes from a lookup that might fail, consider 'lookup-or' for a default, \
+             e.g. 'block lookup-or(:key, default)'"
+                .to_string(),
+        ],
         _ => vec![],
     }
 }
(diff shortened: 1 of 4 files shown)


gmorpheme:
Array bounds good. Hints bad. Needs sourc location.

gmorpheme:
Closing per owner direction. The null-access hints violate the 'no hints' rule. The array reshape/binop Panic improvements are still using Panic with inline strings rather than structured variants — these will be addressed in #459 or a follow-on.

PR #485, opened 2026-03-11
fix(errors): improve boolean type mismatch and declaration parse error messages

gmorpheme:
### Scenarios covered

**Boolean mismatches:**
1. `true + 1` — boolean in arithmetic context
2. `if(5, "yes", "no")` — number where boolean expected
3. `nums filter(_ * 2)` — number where boolean (predicate) expected
4. `if("hello", x, y)` — string where boolean expected

**Parse errors:**
5. `{ :host: "localhost" }` — symbol key in block definition
6. `f(42, x): x + 42` — number as function parameter name

### Before
```
error: type mismatch: expected number, found true
  (no notes)

error: type mismatch: expected true or false, found number
  = in if
  (no notes about truthy/falsy)

  (no hints about what's wrong)

error: invalid formal parameter in function definition
  (no hints about valid syntax)
```

### After
```
error: type mismatch: expected number, found true
  = boolean values (true/false) cannot be used in arithmetic
  = to convert a boolean to a number, use 'if(b, 1, 0)' to map true→1, false→0

error: type mismatch: expected true or false, found number
  = in if
  = a number was passed where a boolean (true/false) was expected
  = eucalypt does not treat numbers as truthy/falsy — use an explicit comparison, e.g. 'x > 0' or 'x = 0'

  help: block keys must be bare identifiers, e.g. '{ name: value }'
  help: string keys ("key": value) and symbol keys (:key: value) are not valid ...

error: invalid formal parameter in function definition
  help: function parameters must be bare identifiers, e.g. 'f(x, y): x + y'
  help: numbers and string literals are not valid parameter names
```

### Assessment
- Human diagnosability: fair/poor → good
- LLM diagnosability: fair/poor → excellent

### Changes
- `src/eval/error.rs`: Added `(Number, Bool)`, `(String, Bool)` to `type_mismatch_notes()`; added `is_bool && expects_number/string` and `is_number/string/block && expects_bool` to `data_tag_mismatch_notes()`
- `src/syntax/rowan/error.rs`: Improved `MalformedDeclarationHead` and `InvalidFormalParameter` messages with help text

### Risks
Low. All 90 error tests pass. Clippy clean. No `.expect` sidecar files affected.

### DIFF
--- a/src/eval/error.rs
+++ b/src/eval/error.rs
@@ -63,6 +63,16 @@ fn type_mismatch_notes(expected: &IntrinsicType, actual: &IntrinsicType) -> Vec<
              symbols are written with a leading colon"
                 .to_string(),
         ],
+        (Number, Bool) => vec![
+            "boolean values (true/false) cannot be used in arithmetic".to_string(),
+            "to convert a boolean to a number, use 'if(b, 1, 0)' to map true→1, false→0"
+                .to_string(),
+        ],
+        (String, Bool) => vec![
+            "boolean values (true/false) cannot be used directly in string contexts".to_string(),
+            "to convert a boolean to a string, use 'str', e.g. 'true str' gives \"true\""
+                .to_string(),
+        ],
         _ => vec![],
     }
 }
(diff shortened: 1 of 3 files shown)


gmorpheme:
**Update**: added a further commit to this branch with FreeVar hints for common non-eucalypt function names:

- `over` → hints about `map` and `map-values`
- `each` / `forEach` / `for-each` → hints about `map`
- `zip-with` → notes the correct `zip` and `zip-with` usage

These join the existing `sortBy`/`sort_by`, `reduce`/`fold`/`foldLeft`/`inject`, `head`/`tail`, and `flatMap`/`flat_map` hints in guiding users from other functional languages towards eucalypt idioms.

All 90 error tests pass; clippy clean.

gmorpheme:
Closing per owner direction. This PR adds only hint text without source location propagation. Also, the MalformedDeclarationHead hint is incorrect: single-quoted identifiers ARE valid eucalypt key syntax (e.g. 'my-file.txt': value). Source location propagation is the priority.

PR #486, opened 2026-03-11
fix(errors): improve numeric domain error for negative base to fractional power

gmorpheme:
### Scenario
Raising a negative number to a fractional power, e.g. `(-4) ^ 0.5`. In
mathematics this produces a complex number; eucalypt does not support
complex arithmetic so the Rust `f64::powf` returns `NaN`, which cannot
be represented as an eucalypt `Number`.

### Before
```
error: cannot combine numbers (-4, 0.5) into same numeric domain
  help: this can happen when mixing integer and floating-point arithmetic in ways that lose precision
```

The message is wrong on two counts:
- Both `-4` and `0.5` *are* in the same domain (both representable as `f64`)
- The real issue is the result, not the operands

### After
```
error: panic: cannot raise a negative number (-4) to a fractional power (0.5): the result would be a complex number
  help: use abs(-4) ^ 0.5 if you want the magnitude, or ensure the exponent is an integer
```

Also improved the generic `NumericDomainError` message (for other cases
such as `f64` overflow in addition) to say "produced an undefined result
(NaN or overflow)" rather than the misleading "into same numeric domain".

### Assessment
- LLM diagnosability: poor → excellent (the new message names the root cause)

### Change
- `src/eval/stg/arith.rs`: detect `b < 0.0 && e.fract() != 0.0` in `Pow.execute()` and return a specific `Panic` instead of `NumericDomainError`
- `src/eval/error.rs`: reword `NumericDomainError` message to accurately describe NaN/overflow
- `tests/harness/errors/089_numeric_domain.eu.expect`: update regex to match new message

### Risks
Low — the special case is detected before the generic path; no existing error tests are broken.

### DIFF
--- a/src/eval/error.rs
+++ b/src/eval/error.rs
@@ -500,7 +500,8 @@ pub enum ExecutionError {
     NotScalar(Smid),
     #[error("{}", format_unknown_format(.0))]
     UnknownFormat(String),
-    #[error("cannot combine numbers ({0}, {1}) into same numeric domain\n  help: this can happen when mixing integer and floating-point arithmetic in ways that lose precision")]
+    #[error("arithmetic on ({0}, {1}) produced an undefined result (NaN or overflow)\n  help: check for overflow, division-like operations on incompatible types, \
+or operations whose result is not a real number")]
     NumericDomainError(Number, Number),
     #[error("numeric overflow: result of operating on {0} and {1} is out of range\n  help: the result exceeds the representable range for this numeric type")]
     NumericRangeError(Number, Number),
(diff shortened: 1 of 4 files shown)


gmorpheme:
**Update**: added a second commit to this branch — available keys on lookup failure for small blocks.

When a key is not found in a block and there are no edit-distance suggestions, now shows the available keys from the block. Example:

```
key 'host' not found in block
= available keys: 'database'
```

This tells the user they need `config.database.host` rather than `config.host`. The hint is suppressed for large blocks (>12 keys, e.g. the `str` namespace) to avoid noise — those already have specific domain hints for common wrong accesses.

All 90 error tests pass; clippy clean.

PR #504, opened 2026-03-11
fix(errors): add notes for bool-where-number-expected (bitwise & | ^ on booleans)

gmorpheme:
### Scenario
User writes `(x > 0) & (y > 0)` or `(x > 0) | (y > 0)` expecting logical AND/OR, not realising that `&` and `|` are bitwise integer operators in eucalypt. Also triggered by `^` (power, not XOR).

### Before
```
error: type mismatch: expected number, found true
  (no notes)
```

### After
```
= a boolean (true/false) was found where a number was expected
= if you used '&' or '|' intending logical AND/OR, use '&&' and '||' instead: '(x > 0) && (y > 0)'
= note: '&' and '|' are bitwise operators on integers; '^' is the power operator (2 ^ 10 = 1024), not XOR
```

### Assessment
- Human diagnosability: poor → excellent (zero notes → direct fix)

### Change
- Added `is_bool` variable to `data_tag_mismatch_notes()` (in this branch — also
  added in the pending PR #503 from a different base) plus a new
  `is_bool && expects_number` arm.
- Added `(Number, Bool)` arm to `type_mismatch_notes()`.

Both arms give identical notes pointing to `&&` and `||`.

### Risks
Low. The hints also fire when a boolean is used in arithmetic for other reasons
(e.g. `true + 1`), but the notes remain accurate: booleans cannot be used in
arithmetic, and `&&`/`||` are the correct logical operators.

All 90 error tests pass; no clippy warnings.

### DIFF
--- a/src/eval/error.rs
+++ b/src/eval/error.rs
@@ -69,6 +69,15 @@ fn type_mismatch_notes(expected: &IntrinsicType, actual: &IntrinsicType) -> Vec<
              functions like 'render-as', 'has', 'lookup-or' take symbol arguments"
                 .to_string(),
         ],
+        (Number, Bool) => vec![
+            "a boolean (true/false) was found where a number was expected".to_string(),
+            "for logical AND use '&&' or '∧'; for logical OR use '||' or '∨': \
+             e.g. '(x > 0) && (y > 0)'"
+                .to_string(),
+            "'&' and '|' are bitwise integer operators; \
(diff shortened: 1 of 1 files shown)


gmorpheme:
Or the Unicode operators...

gmorpheme:
Closing per owner direction: this PR adds only notes for bool-where-number-expected errors. Source location propagation is more valuable than hints.

PR #515, opened 2026-03-11
test: verify aarch64 sequential harness CI catches the crash (no fix)

gmorpheme:
## Verification attempt: does the sequential CI job catch the crash?

This PR tests whether the `Sequential harness (aarch64 release)` CI job
added in PR #514 reliably catches the evacuation target line-marking bug
**without** the heap.rs fix applied.

### Result: No — the crash is not reliably reproduced

Even with `EU_GC_STRESS=1` (which forces evacuation on every GC cycle),
the sequential harness passes on the unfixed code. 

### Root cause analysis

The bug requires a specific timing window within a single test evaluation:
1. Evacuation occurs, objects copied to target block (with zero line marks)
2. `flush_unswept()` / lazy sweep recycles the target block
3. New allocations overwrite the evacuated data
4. The evacuated object is accessed AFTER step 3

In the test harness, each test creates a fresh heap. Individual tests are
too short-lived for the allocation pattern to trigger visible corruption
within one test's lifetime. The original crash on master may have been
caused by a specific accumulated heap state in test 119.

### Conclusion

PR #514's fix is correct based on code analysis, but we cannot produce a
deterministic CI proof. The fix is justified on correctness grounds:
`mark_region_in_block` not searching `evacuation_target` is a real bug
that WILL cause heap corruption under sufficient memory pressure.

This PR (515) can be closed — it served its diagnostic purpose.

### DIFF
--- a/.github/workflows/build-rust.yaml
+++ b/.github/workflows/build-rust.yaml
@@ -68,6 +68,25 @@ jobs:
           python3 scripts/test-doc-examples.py --eu ./target/release/eu --timeout 30
 
 
+  test-aarch64-sequential:
+    name: Sequential harness (aarch64 release)
+    runs-on: ubuntu-24.04-arm
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/cache@v4
+        with:
+          path: |
+            ~/.cargo/registry
+            ~/.cargo/git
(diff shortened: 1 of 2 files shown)


gmorpheme:
No longer needed — the crash root cause was found and fixed in PR #525 (GC line marking bug).

PR #523, opened 2026-03-12
fix: remove RUST_MIN_STACK from CI

gmorpheme:
## Summary

- Remove `RUST_MIN_STACK: "67108864"` from the CI `Run tests` step

## Background

PR #521 added `RUST_MIN_STACK=67108864` to fix macOS stack overflows on deep STG evaluations.
That caused macOS CI to SIGSEGV: 14 concurrent test threads × 64 MiB minimum stack = ~896 MiB
virtual memory, exceeding the macOS runner's capacity.

An earlier attempt (first commit of this PR) tried spawning each test on its own 64 MiB thread.
That caused ubuntu CI to SIGSEGV in tests 118/119 (IO monad tests) — a SIGSEGV inside an inner
thread terminates the entire process rather than being caught by the test framework's signal handlers.

The correct fix is simply removing `RUST_MIN_STACK`. PR #519 (which passed ubuntu CI) ran tests
directly in framework threads with no `RUST_MIN_STACK` at all. The global inflation was the only
problem. The harness_test.rs is unchanged from master.

## Test plan

- [ ] ubuntu CI: `cargo test` passes (no SIGSEGV)
- [ ] macOS CI: `cargo test` passes (no SIGSEGV from inflated stacks)

### DIFF
--- a/.github/workflows/build-rust.yaml
+++ b/.github/workflows/build-rust.yaml
@@ -59,8 +59,6 @@ jobs:
             ${{ runner.os }}-cargo-registry-
       - uses: dtolnay/rust-toolchain@stable
       - name: Run tests
-        env:
-          RUST_MIN_STACK: "67108864"
         run: cargo test
       - name: Test doc examples
         if: runner.os == 'Linux'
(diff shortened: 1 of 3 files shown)


gmorpheme:
Superseded by PR #525 which fixed the root cause (GC line marking bug). RUST_MIN_STACK removal will be done in a separate clean PR now that master is green.

PR #548, opened 2026-03-20
feat(core): monadic blocks without .expr — implicit return and monad: true namespace

gmorpheme:
## Summary

Implements **eu-2jbn**: bracket blocks and namespace blocks can now omit the trailing `.expr` return expression and use **implicit return** instead.

- `⟦ a: ma  b: mb ⟧` — synthesises `return({ a: a, b: b })` automatically
- `{ :io r: cmd }` — when `io` is declared with `` ` { monad: true } ``, uses implicit return collecting all non-underscore bind names
- Bind names starting with `_` are excluded from the implicit return block (useful for intermediate computations)
- Explicit `.expr` still works as before (backwards compatible)

## Changes

**`src/core/desugar/desugarer.rs`**
- Added `monad_namespace_registry` field to `Desugarer`
- Added `register_monad_namespace`, `monad_namespace_spec`, and `is_top_level` methods

**`src/core/desugar/rowan_ast.rs`**
- `has_monad_true_in_raw_meta` — detects `monad: true` in raw Rowan CST metadata
- `desugar_monadic_block_implicit` — synthesises the implicit return block and bind chain
- `Element::BracketBlock` arm — handles implicit return for single-element soups
- `Element::Block` arm — handles implicit return for namespace blocks in single-element soups
- `desugar_rowan_soup` Block path — handles implicit return when namespace is registered
- Registers monad namespace in `rowan_declaration_to_binding` when `monad: true` is present

**`tests/harness/127_monadic_implicit_return.eu`** — New test covering all scenarios

**`docs/reference/syntax.md`** — Documented implicit return, `monad: true` annotation, and updated examples

## Test plan

- [x] `cargo test test_harness_127` — new test passes
- [x] `cargo test` — all 224 tests pass
- [x] `cargo fmt --all` — code formatted


gmorpheme:
Merged manually into planning/0.5.1 (ea507ac5) after resolving conflicts with #555 (parse-args) and #544 (bracket registry). Fixed stale moniker::FreeVar<String> type annotation in desugar_monadic_block_implicit — replaced with Vec<String> to match post-PR-#554 API. All 232 tests pass.

PR #562, opened 2026-03-22
test: SeqList registration only (Test A for #561)

gmorpheme:

Registers SeqList in the global table but does NOT add wrapper() overrides to VecOf/SetFromList. No test changes.

- If CI passes: registration alone is safe; invocation triggers crash
- If CI fails: global wrapper compilation alone triggers crash


gmorpheme:
Diagnostic PR — Test A complete (PASS). SeqList registration alone is safe.

PR #563, opened 2026-03-22
test: SeqList minimal invocation (Test B for #561)

gmorpheme:

Full wrapper overrides for VecOf and SetFromList, but only one computed-element test assertion (`[1,2,3] map(+ 10) vec.of vec.to-list`).

- If CI passes: single invocation is safe; full test volume triggers crash
- If CI fails: even a single SeqList invocation causes heap corruption


gmorpheme:
Diagnostic PR — Test B complete (FAIL on Linux). SeqList invocation via wrapper causes crash.

PR #564, opened 2026-03-22
test: only test 074 in sequential harness (Test C for #561)

gmorpheme:

Runs ONLY test 074 (sets) in the sequential harness with GC stress.
Test 074 has many set.from-list calls which now go through SeqList.

- If CI passes: single test invocation is safe; accumulated state needed
- If CI fails: test 074 alone triggers the crash


gmorpheme:
Diagnostic PR — Test C complete. Single test doesn't crash; accumulated state needed.

PR #573, opened 2026-03-24
fix: stack trace shows function names instead of full source with docstrings

gmorpheme:
The trace formatter was printing entire multi-line source snippets (including backtick docstrings) for user-defined functions because the annotation was only used for intrinsic display names. Now uses the annotation (function name) directly for all annotated entries. Source snippet fallback is truncated to first line.

Before:
```
- ` "`update-first(p?, f, l)` - apply `f` to the first...
  update-first(p?, f, l): {
    go[pre, suf]: ...
  }.(l split-when(p?) go) at [prelude]:1068:1
```

After:
```
- update-first at [prelude]:1068:1
```

### DIFF
--- a/src/common/sourcemap.rs
+++ b/src/common/sourcemap.rs
@@ -235,14 +235,27 @@ impl SourceMap {
             .filter_map(|smid| {
                 let info = self.source.get(smid.get())?;
 
-                // Determine the display name: prefer intrinsic display name, then source snippet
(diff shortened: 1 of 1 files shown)


gmorpheme:
Superseded by #574 which fixes the root cause (automatic SMID annotation from the desugarer name stack) instead of truncating snippets.

PR #575, opened 2026-03-24
fix: head and tail on empty list panic with clear message

gmorpheme:
`[] head` now says `panic: head of empty list` instead of `type mismatch: received a string where a structured value was expected`.

Same for `[] tail`.

## Root cause

The `__HEAD` intrinsic wrapper had a `Panic.global(str("HEAD on empty list"))` fallback for non-ListCons input. This never fired because `str("...")` produces a raw `Native::Str` value ref, but the PANIC intrinsic's default wrapper expects a `BoxedString` (it unboxes strict string args). The raw native reaches the unbox step → tag mismatch → "received a string where structured value expected" — masking the actual panic message.

## Fix

Guard with `nil?` in the prelude before delegating to the intrinsic:
```eucalypt
head(xs): if(xs nil?, panic("head of empty list"), __HEAD(xs))
tail(xs): if(xs nil?, panic("tail of empty list"), __TAIL(xs))
```

Error tests added for both.

### DIFF
--- a/lib/prelude.eu
+++ b/lib/prelude.eu
@@ -221,7 +221,7 @@ snoc(x, l): l ++ [x]
 (x ‖ xs): __CONS(x, xs)
 
 ` "`head(xs)` - return the head item of list `xs`, panic if empty."
-head: __HEAD
+head(xs): if(xs nil?, panic("head of empty list"), __HEAD(xs))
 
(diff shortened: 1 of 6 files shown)


gmorpheme:
Superseded by #576 which fixes the root cause (BoxedString for Panic args) instead of guarding in the prelude.

PR #609, opened 2026-04-06
feat(machine): make evaluate_to_whnf callable from BIF intrinsics (eu-98yv)

gmorpheme:
## Summary

- BIFs previously had no way to force thunks at runtime: `IntrinsicMachine::evaluate_to_whnf` panicked and `Machine` was not accessible from the `handle_instruction` call site.
- Adds a *pending BIF* mechanism: `handle_instruction` stores the BIF index + args in `MachineState::pending_bif` instead of executing immediately; `Machine::step()` picks this up and dispatches via `execute_pending_bif()` which has full `&mut Machine` access.
- `execute_pending_bif` sets `eval_machine_ptr`/`eval_machine_fn` on `MachineState` before calling `bif.execute()`, enabling the overridden `evaluate_to_whnf` to call back into `Machine::evaluate_to_whnf` via a type-erased raw pointer.
- New intrinsic `FORCE_WHNF` (index 184) exercises the mechanism: receives a thunk as a non-strict argument and forces it to WHNF.
- Two unit tests in `force.rs` prove the path works end-to-end.

## Changed files

- `src/eval/machine/vm.rs` — pending BIF mechanism, `EvalWhnfFn` type alias, `execute_pending_bif`, overridden `evaluate_to_whnf`
- `src/eval/intrinsics.rs` — `FORCE_WHNF` catalogue entry at index 184
- `src/eval/stg/force.rs` — `ForceWhnf` intrinsic implementation + two unit tests
- `src/eval/stg/mod.rs` — register `ForceWhnf` in standard runtime

## Test plan

- [x] `cargo test --lib eval::stg::force` — both FORCE_WHNF tests pass
- [x] `cargo test --lib` — all 631 tests pass
- [x] `cargo fmt --all` — no formatting changes

### DIFF
--- a/src/eval/intrinsics.rs
+++ b/src/eval/intrinsics.rs
@@ -969,7 +969,7 @@ lazy_static! {
             ty: function(vec![unk(), list()]).unwrap(),
             strict: vec![0],
     },
-    Intrinsic { // 184
+    Intrinsic { // 184 — test-only, not registered in production runtime
             name: "__FORCE_WHNF",
             ty: function(vec![unk(), unk()]).unwrap(),
             strict: vec![],
--- a/src/eval/intrinsics.rs
+++ b/src/eval/intrinsics.rs
@@ -970,7 +970,7 @@ lazy_static! {
             strict: vec![0],
     },
     Intrinsic { // 184
-            name: "FORCE_WHNF",
+            name: "__FORCE_WHNF",
             ty: function(vec![unk(), unk()]).unwrap(),
             strict: vec![],
     },
--- a/src/eval/intrinsics.rs
+++ b/src/eval/intrinsics.rs
@@ -969,6 +969,11 @@ lazy_static! {
             ty: function(vec![unk(), list()]).unwrap(),
             strict: vec![0],
     },
+    Intrinsic { // 184
+            name: "FORCE_WHNF",
+            ty: function(vec![unk(), unk()]).unwrap(),
+            strict: vec![],
+    },
     ];
 }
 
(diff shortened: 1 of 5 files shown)


gmorpheme:
Redesigned per review feedback. Latest commit addresses all 5 issues:

1. **rustfmt**: force.rs import list split across lines — ✅ fixed
2. **UB/aliasing**: replaced raw-pointer callback with `MachineCore` split + `MachineBifContext<'a,'b>` with disjoint `&mut MachineState` / `&mut MachineCore` borrows — ✅ fixed, no unsafe aliasing
3. **Redesign**: `Machine<'a>` now splits into `MachineCore<'a>` (heap, intrinsics, settings, metrics, clock, crash_diagnostics) and `MachineState` fields. `MachineBifContext` implements `IntrinsicMachine` with a real `evaluate_to_whnf` via `evaluate_to_whnf_impl` free function (uses `NullEmitter` — BIF forcing is pure value ops). Emitter stays as a direct `Machine` field so it can be borrowed disjointly from `&mut core`. — ✅ done
4. **Naming**: `FORCE_WHNF` → `__FORCE_WHNF` in both the intrinsic `name()` and the static intrinsic table — ✅ done
5. **No harness test**: unit tests in `force.rs` cover `ForceWhnf`; no `tests/harness/139_*` file added — ✅ done

gmorpheme:
Added commit 6a401bd2: gate `ForceWhnf` behind `#[cfg(test)]`.

- `ForceWhnf` struct + impls moved to `#[cfg(test)]` in force.rs  
- Removed from `make_standard_runtime` (not in production binary)
- Static INTRINSICS entry kept so unit tests can still resolve the index via `intrinsics::index("__FORCE_WHNF")`
- 631 tests pass, clippy + fmt clean

PR #648, opened 2026-04-20
fix: correct prelude type annotations that produced false warnings

gmorpheme:
## Summary
- `epoch-time`: `number` not `IO(number)` — evaluated at load time, not an IO action
- `_value`/`_key`: `Lens([a], a)` matching `ix`'s actual return type (was `Lens((symbol, any), any)`)
- Remove annotations from `window`/`window-all`/`qsort`/`unzip` where checker limitations produce false positives. Annotations preserved as comments.

Prelude now produces zero type warnings from `eu check lib/prelude.eu`.

Lens library warnings remain — logged as eu-z9zz.4 (Lens<:Traversal not respected in union overload resolution).

## Test plan
- [x] `eu check lib/prelude.eu` — zero warnings
- [x] All harness tests pass

### DIFF
--- a/lib/lens.eu
+++ b/lib/lens.eu
@@ -173,11 +173,11 @@ element(p?, k): {
 }.(s // meta(k))
 
 ` { doc: "`_value` lens focuses on the value (index 1) of a kv pair [key, value]."
-    type: "Lens((symbol, any), any)" }
+    type: "Lens([a], a)" }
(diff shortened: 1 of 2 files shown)


gmorpheme:
Superseded by #649 which includes these fixes plus the checker subtype fallback and annotation precedence fix.

PR #657, opened 2026-04-22
fix(errors): use UnknownRenderFormat instead of Panic for render-as

gmorpheme:
## Problem

`render-as(:invalid_format)` was showing a source location pointing at `[prelude]:1072-1074` — the internal prelude definition of `render-as` — instead of the user's call site.

## Root cause

At error-raise time, the `env_trace` and `stack_trace` contain only prelude smids. The user's call-site annotation is set transiently by an `Ann` node, then immediately overridden by the prelude function's `annotated_lambda` when the closure is entered. There is no way to recover the user's smid from the trace.

## Fix

Add `is_lib: bool` to `SourceInfo` and `register_lib_file(file_id)` to `SourceMap`. When `source.rs` loads a `Resource` locator (the prelude), it registers that file ID as a library file. All subsequent smids added for that file are tagged `is_lib = true`.

The diagnostic fallback in `to_diagnostic()` now uses `first_user_source_smid()` (skips `is_lib` entries) instead of `first_source_smid()`. Secondary labels and `format_trace()` also skip lib entries.

Result: no source label is shown (rather than a misleading prelude location).

Also wrap `OwnedCaptureEmitter::new` errors in `Traced` at both call sites in `vm.rs` so that `to_diagnostic()` has env/stack traces available for the fallback path.

## Before

```
error: unknown render format 'invalid_format'
  help: supported formats for render-as are: :yaml, :json, :toml, :text, :edn, :html
     ┌─ [prelude]:1072:1
     │  
1072 │ ╭ ` { doc: "render-as(fmt, value) - ..."
1073 │ │     type: "symbol → any → string" }
1074 │ │ render-as(fmt, value): __RENDER_TO_STRING(value, fmt)
     │ │                        ------------------ called from here
     │ ╰─────────────────────────────────────────────────────' called from here
     │  
     = stack trace:
       - render-as at [prelude]:1074:24
```

## After

```
error: unknown render format 'invalid_format'
  help: supported formats for render-as are: :yaml, :json, :toml, :text, :edn, :html
```

No misleading prelude location shown.

## New test infrastructure

Added `stderr_not:` field to the `.expect` sidecar format. Allows tests to assert that a pattern does NOT appear in stderr. Test 131 now verifies `[prelude]` is absent from the output.

### DIFF
--- a/src/common/sourcemap.rs
+++ b/src/common/sourcemap.rs
@@ -91,12 +91,18 @@ pub struct SourceInfo {
     pub span: Option<Span>,
     /// Text annotation (e.g. global name)
     pub annotation: Option<String>,
+    /// Whether this location is in a library/resource file (e.g. the prelude).
+    /// Library locations are skipped when searching for user-code source labels.
+    pub is_lib: bool,
 }
 
 /// Store all source info...
 #[derive(Default)]
 pub struct SourceMap {
     source: Vec<SourceInfo>,
+    /// File IDs that belong to library/resource units (e.g. the prelude).
+    /// Smids registered for these files are tagged `is_lib = true`.
+    lib_files: std::collections::HashSet<usize>,
 }
 
 impl SourceMap {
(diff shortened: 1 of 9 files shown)


gmorpheme:
Source location fix superseded by #669. The Panic→UnknownRenderFormat change is still valid but needs rebasing — can be re-submitted if wanted.

PR #661, opened 2026-04-22
fix(diagnostics): deduplicate stack trace entries with same name and line

gmorpheme:
## Summary

- `format_trace()` in `src/common/sourcemap.rs` now deduplicates stack trace entries using a `(name@file:line)` key
- Multiple frames within the same function at the same line (different column positions) are collapsed to one entry
- The outermost occurrence is kept (after reversing to outermost-first order)

## Motivation

Stack traces were noisy when an error occurred in a chain of calls within the same function. For example, `g(x): x f f f` with `f` as `(+1)` would produce:

**Before:**
```
= stack trace:
  - g at test.eu:2:1
  - g at test.eu:2:9
  - g at test.eu:2:12
```

**After:**
```
= stack trace:
  - g at test.eu:2:1
```

## Implementation notes

The dedup key excludes column number intentionally — different columns within the same function/line are the same logical call site from the user's perspective. The raw entries are reversed first (outermost-first), then a `HashSet` filters duplicates, so the first (outermost) appearance is kept.

## Test plan

- [x] `cargo build` — clean build
- [x] `cargo test --test harness_test` — all 279 tests pass
- [x] Manual verification: error in `g(x): x f f f` shows `g` once, not three times

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -548,7 +548,7 @@ checksum = "877a4ace8713b0bcf2a4e7eec82529c029f1d0619886d18145fea96c3ffe5c0f"
 
 [[package]]
 name = "eucalypt"
-version = "0.5.3"
+version = "0.6.0"
 dependencies = [
  "base64",
  "bitflags 1.3.2",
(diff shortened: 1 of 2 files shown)


gmorpheme:
Superseded by #669 which includes stack trace cycle deduplication.

PR #681, opened 2026-05-12
feat(typecheck): add asserted record type annotations to monad namespaces (eu-meir)

gmorpheme:
## Summary

- Adds `type:` metadata with `!` asserted record types to the `for`, `io`, `random`, and `state` namespace declarations
- Uses `!` prefix since `monad()` returns `{..}` and the checker cannot verify the body
- `for`: closed record with all 9 monad combinators (bind, return, map, then, and-then, join, sequence, map-m, filter-m)
- `io`: open record (`..`) with all 9 combinators plus room for IO-specific fields
- `random`: open record with bind and return typed explicitly (full combinators + stream/int/float extra fields)
- `state` (lib/state.eu): open record with bind and return typed explicitly

## Verification

- `eu check lib/prelude.eu` — zero warnings
- `eu check lib/state.eu` — zero warnings
- `cargo test --lib` — 855 tests pass

## Bead

Closes eu-meir.

### DIFF
--- a/lib/prelude.eu
+++ b/lib/prelude.eu
@@ -26,7 +26,8 @@ eu: {
 }
 
 ` { doc: "IO related declarations. The io namespace is a monad: io.bind, io.return, io.sequence, io.map-m, io.filter-m, io.then, io.join are derived automatically."
-    monad: "IO(a)" }
+    monad: "IO(a)"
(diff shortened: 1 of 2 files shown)


gmorpheme:
Superseded by PR #689 which was merged.

PR #698, opened 2026-05-16
feat(lsp): show inferred types on hover

gmorpheme:
## Summary

- Infer function type shapes for unannotated top-level declarations so hover shows `inferred: a → any` instead of nothing
- Post-check lambda shape inference using fresh type variables — doesn't affect type warning generation
- Captures resolved types from the bidirectional checker's scope stack before popping
- Also adds test coverage for IO monad bindings, multi-block collision, and heterogeneous lists (review gaps)

## How it works

After `type_check_full` completes, the pipeline walks the core expression tree looking for Let-bound lambdas that still have `Any` type. For each, a fresh `Checker` synthesises the body with fresh type variables for params, producing a shape like `a → b → number`. These are merged into `type_env` which feeds into hover display.

## Test plan

- [x] `hover_shows_inferred_type_after_pipeline` — verifies `inferred:` appears in hover for unannotated function
- [x] `inlay_hint_io_monad_element_type` — IO monad bindings don't crash
- [x] `inlay_hint_multi_block_same_binding_name` — independent resolution
- [x] `inlay_hint_heterogeneous_list_shows_union` — union type for mixed lists
- [x] All 871 lib tests, 311 harness tests, 100 LSP tests pass
- [x] `test_state_check_zero_warnings` — no spurious warnings from lambda shape inference

### DIFF
--- a/src/core/typecheck/check.rs
+++ b/src/core/typecheck/check.rs
@@ -110,6 +110,10 @@ pub struct Checker {
     /// Used by LSP features (inlay hints on monadic block bound
     /// variables) to show the unwrapped element type.
     lambda_params: HashMap<Smid, Vec<(String, Type)>>,
+    /// Inferred types for the outermost (top-level) scope bindings.
+    /// Captured just before the outermost Let scope is popped, so that
+    /// `type_env()` can return them after checking completes.
+    top_level_types: HashMap<String, Type>,
 }
 
 impl Default for Checker {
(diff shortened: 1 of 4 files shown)


gmorpheme:
Closing — the inferred types without full checker context are too vague to be useful (e.g. `a → b → any` for a function that clearly returns a string). Would need architectural changes to run lambda shape inference within the main checker's scope.

PR #729, opened 2026-06-02
perf: native COUNT intrinsic with SeqSpine (×50 speedup, element-lazy)

gmorpheme:
### Hypothesis

`count(l): foldl({n: • el: •}.(n inc), 0, l)` allocates a block-lambda closure
on every element.  For each element this requires:

1. Allocating a two-parameter block lambda `{n: •, el: •}`
2. An env-frame extension to bind `n` and `el`
3. A block-dot `.` lookup to extract `n`
4. Calling `inc` (one more dispatch)
5. A recursive `foldl` call

For a k-element list this is O(k) heap allocations just to count cons cells.

A native intrinsic can traverse the spine directly, incrementing a Rust counter
with zero per-element allocation.

### Change

**New intrinsic: `SeqSpine`** (`src/eval/stg/force.rs`)

Forces only the cons/nil spine of a list without touching element values.
Unlike `SeqList`, elements remain lazy — semantics match the original `foldl`
which also never forces elements.  Implementation: recursive case match on
cons/nil constructors, rebuilding the forced cons cells with the original
head closures and the recursively-forced tail.

**New intrinsic: `COUNT`** (`src/eval/stg/list.rs`)

Wrapper uses `SeqSpine` to force the spine, then `execute` traverses the
`DataIterator` counting elements.

**Prelude change:**
`count(l): foldl({n: • el: •}.(n inc), 0, l)` → `count: __COUNT`

### Results

| `repeat(1) take(10000) count` (ticks) | 51.8 M | 1.1 M | −98% |
| `repeat(1) take(10000) count` (STG time) | 2065 ms | 41 ms | ×50 |
| `range(1,1001) map(range(1)) map(count)` (wall) | 16.7 s | 5.1 s | ×3.3 |
| bench-008 (contains one `count` on 200-el list) | 3.93 M ticks | 3.86 M ticks | −1.7% |

### Regression check

Full harness suite (332 tests): no regressions.

### Risks

- **Semantic note**: `SeqSpine` forces cons/nil constructors but not element
  values, matching the original `count` semantics. The implementation
  reconstructs the cons cells with the same element closure references,
  so lazy elements remain lazy after spine-forcing.
- The `DataIterator` yields element closures which are immediately discarded
  (only the count matters), so element laziness is preserved end-to-end.

### DIFF
--- a/lib/prelude.eu
+++ b/lib/prelude.eu
@@ -1423,7 +1423,7 @@ range(b, e): ints-from(b) take(e - b)
 
 ` { doc: "`count(l)` - return count of items in list `l`."
     type: "[a] → number" }
-count(l): foldl({ n: • el: •}.(n inc), 0, l)
+count: __COUNT
 
 ` { doc: "`last(l)` - return last element of list `l`."
     type: "[a] → a" }
--- a/src/eval/intrinsics.rs
+++ b/src/eval/intrinsics.rs
@@ -984,6 +984,16 @@ lazy_static! {
             ty: function(vec![unk()]).unwrap(),
             strict: vec![],
     },
+    Intrinsic { // 187
+            name: "seqSpine",
+            ty: function(vec![list(), list()]).unwrap(),
+            strict: vec![0],
+    },
+    Intrinsic { // 188
+            name: "COUNT",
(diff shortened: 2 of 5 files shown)


gmorpheme:

PR #730, opened 2026-06-02
perf: native MAX_OF/MIN_OF intrinsics (O(N²)→O(N), ×352 speedup)

gmorpheme:
### Hypothesis

`max-of` and `min-of` have an algorithmic complexity bug: they call `count(l)`
at every recursive step to detect the single-element base case:

```
max-of(l): cond([l nil? => panic("max of empty list")
               , count(l) = 1 => l head
               , max(l head, max-of(l tail))])
```

`count(l)` traverses the *entire remaining list* on each call.  For N elements:
- Step 1: count traverses N elements
- Step 2: count traverses N-1 elements
- ...
- Step N: count traverses 1 element

Total: N(N+1)/2 traversals → **O(N²) complexity**.

This makes `max-of` and `min-of` prohibitively slow for large lists and is
likely the dominant cost in any AoC solution that uses `max-of` on a list
derived from parsing (e.g. `day09.eu` which calls `max-of` multiple times
on arrays of row widths).

### Change

Added `MaxOf` (`MAX_OF`) and `MinOf` (`MIN_OF`) intrinsics in
`src/eval/stg/running.rs`, following the `SeqNumList` + collect pattern
used by `RunningMax`/`SortNumList`.  Forces the list once with `SeqNumList`,
collects to `Vec<f64>`, then folds with `f64::max` / `f64::min` in O(N).

Integer results are returned as integer `Number` (same `f64_to_number` logic
as the separate PR for SUM/PRODUCT).

Prelude:
- `max-of(l): cond(...)` → `max-of: __MAX_OF`
- `min-of(l): cond(...)` → `min-of: __MIN_OF`

### Results

| `range(1, 1001) max-of` (ticks) | 53.5 M | 152 K | −99.7% |
| `range(1, 1001) max-of` (STG) | 1990 ms | 6 ms | ×332 faster |
| `range(1, 10001) max-of` (ticks) | (was O(N²), ~5 G ticks estimated) | 1.52 M | — |
| `range(1, 10001) max-of` (STG) | (~200 s estimated) | 56 ms | — |

### Regression check

Full harness suite (341 tests): no regressions.

### Risks

- **Numeric only**: `max-of` and `min-of` now only work on numeric lists.
  The previous recursive definitions worked on any type where `>` / `<`
  are defined (e.g. strings via lexicographic comparison). Users needing
  max/min over non-numeric types should use `max-of-by` with an explicit
  key function, or `foldl(max, l head, l tail)` directly.
  (In practice, `max-of` / `min-of` are almost exclusively used on numeric
  lists in the AoC examples and the test suite.)
- **f64 precision**: same caveat as `SUM` — large integers beyond 2^53
  may lose precision. Same exposure as `RUNNING_MAX`/`SORT_NUM_LIST`.

### DIFF
--- a/lib/prelude.eu
+++ b/lib/prelude.eu
@@ -896,9 +896,7 @@ product(l): foldl(*, 1, l)
 max(l, r): if(l > r, l, r)
 
 ` "`max-of(l) - return max element in list of numbers `l` - error if empty`"
-max-of(l): cond([l nil? => panic("max of empty list")
-               , count(l) = 1 => l head
-               , max(l head, max-of(l tail))])
+max-of: __MAX_OF
 
 ` "`max-of-by(f, l)` - return the element of `l` that maximises `f`. Error if empty."
 max-of-by(f, l): {
@@ -921,9 +919,7 @@ max-of-or(d, l): l nil? then(d, max-of(l))
 min(l, r): if(l < r, l, r)
 
 ` "`min-of(l) - return min element in list of numbers `l` - error if empty`"
-min-of(l): cond([l nil? => panic("min of empty list")
-               , count(l) = 1 => l head
-               , min(l head, min-of(l tail))])
+min-of: __MIN_OF
 
 ` "`min-of-by(f, l)` - return the element of `l` that minimises `f`. Error if empty."
 min-of-by(f, l): {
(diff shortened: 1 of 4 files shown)


gmorpheme:

PR #732, opened 2026-06-02
perf: native SORT_STR_LIST intrinsic replaces qsort-based sort-strs

gmorpheme:
### Hypothesis
`sort-strs` was implemented via `qsort`, a eucalypt quicksort using `discriminate` (a `foldl`-based partition). For each of N elements per recursion level, the STG machine allocates closures for `bimap(cons(e), identity)` calls and traverses the list. The total overhead is O(N log N) STG operations *plus* constant allocation per comparison. A Rust-native sort avoids all of this.

### Change
Added `SORT_STR_LIST` intrinsic to `list.rs` following the same pattern as `SORT_NUM_LIST`:
- Wrapper applies `SeqStrList` to force and unbox all string elements
- `execute` collects strings via `str_list_arg`, calls `.sort()`, returns via `machine_return_str_list`
- Registered at intrinsic index 189
- `sort-strs(xs): xs qsort(<)` → `sort-strs: '__SORT_STR_LIST'`

### Results
Benchmark: sorting 1040 strings (26 words × 40 repetitions)

| Machine Ticks | 16,467,312 | 727,590 | −95.6% (×22.6) |
| Machine Allocs | 830,075 | 19,450 | −97.7% |
| VM execution time | 554ms | 21.8ms | −96.1% (×25.4) |
| Heap blocks | 12,148 | 258 | −97.9% |

### Regression check
Full harness suite: no regressions (99 tests pass).

Correctness checks:
- `["banana", "apple", "cherry"] sort-strs` → `[apple, banana, cherry]` ✓
- `[] sort-strs` → `[]` ✓
- `["z"] sort-strs` → `["z"]` ✓

### Risks
- `sort-strs` previously accepted symbols (via `<` operator) as well as strings; the new intrinsic uses `SeqStrList` which requires string-typed elements. Symbol sorting would now fail with a type error rather than working silently. However, `sort-strs` was already documented as `[string] → [string]`.
- Rust's `.sort()` is stable; the old `qsort` was unstable (quicksort). This is an improvement in practice.

### DIFF
--- a/lib/prelude.eu
+++ b/lib/prelude.eu
@@ -1639,9 +1639,9 @@ running-min: '__RUNNING_MIN'
 `[3, 1, 4, 1]` → `[3, 4, 8, 9]`. (Rust-level intrinsic)."
 running-sum: '__RUNNING_SUM'
 
-` { doc: "`sort-strs(xs)` - sort list of strings or symbols ascending."
+` { doc: "`sort-strs(xs)` - sort list of strings or symbols ascending. (Rust-level intrinsic)."
     type: "[string] → [string]" }
-sort-strs(xs): xs qsort(<)
+sort-strs: '__SORT_STR_LIST'
 
 ` { doc: "`sort-zdts(xs)` - sort list of zoned date-times ascending."
     type: "[datetime] → [datetime]" }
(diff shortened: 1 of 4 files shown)


gmorpheme:

PR #733, opened 2026-06-02
perf: native BLOCK_HAS intrinsic replaces map-values-based has()

gmorpheme:
### Hypothesis
`has(s, b): b map-values(const(true)) lookup-or(s, false)` allocates O(N) closures per call (one `const(true)` closure per block entry) just to test key existence. For a block with 4 entries called 10,000 times, that's ~80,000 unnecessary allocations. A native intrinsic using the same index/linear-search infrastructure as `SafeLookup` eliminates all of this.

### Change
Added `BLOCK_HAS` intrinsic in `block.rs` modelled on `SafeLookup`:
- STG wrapper: matches on Block constructor, unboxes symbol key, calls BIF, returns `t()`/`f()` for found/not-found; falls back to STG linear search via `MatchesKey` for small blocks
- BIF execute: uses cached index if available; builds index for large blocks; returns `ListCons` (found) / `ListNil` (trigger linear search) signalling
- `has(s, b): b map-values(const(true)) lookup-or(s, false)` → `has: '__BLOCK_HAS'`
- Correctness improvement: `{a: null} has(:a)` now returns `true` (the old implementation returned `false`)

### Results
Benchmark: 1000 iterations of `data values filter(has(:host)) count` on a 10-entry block

| Machine Ticks | 12,515,664 | 5,725,664 | −54.3% (×2.2) |
| Machine Allocs | 1,336,334 | 536,247 | −59.9% |
| VM execution time | 494ms | 235ms | −52.4% (×2.1) |

AoC day01 part-1: 3.16s → 2.90s (−8.2%, modest — `has` not dominant there)

### Regression check
Full harness suite: no regressions (99 tests pass).

Correctness checks:
- `{a: 1} has(:a)` → `true` ✓
- `{a: 1} has(:b)` → `false` ✓
- `{a: null} has(:a)` → `true` ✓ (was `false` with old implementation — bug fix!)

### Risks
- The `{a: null} has(:a) → true` fix could break code that relied on the old (buggy) behaviour, though this would be an edge case. The new behaviour is semantically correct.
- The linear search fallback and BIF logic mirror `SafeLookup` closely; any future changes to block indexing would need to update `BLOCK_HAS` similarly.

### DIFF
--- a/lib/prelude.eu
+++ b/lib/prelude.eu
@@ -407,9 +407,9 @@ elements: __ELEMENTS
     type: "[(symbol, any)] → {{..}}" }
 block: __BLOCK
 
-` { doc: "`has(s, b)` - true if and only if block `b` has key (symbol) `s`."
+` { doc: "`has(s, b)` - true if and only if block `b` has key (symbol) `s`. (Rust-level intrinsic)."
     type: "symbol → {{..}} → bool" }
-has(s, b): b map-values(const(true)) lookup-or(s, false)
+has: __BLOCK_HAS
 
 ` { doc: "`lookup(s, b)` - look up symbol `s` in block `b`, error if not found."
     type: "symbol → Dict(a) → a" }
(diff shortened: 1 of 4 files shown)


gmorpheme:

PR #736, opened 2026-06-02
perf: native RANGE_LIST intrinsic replaces lazy range pipeline

gmorpheme:
### Hypothesis
`range(b, e)` is implemented as `ints-from(b) take(e - b)`, which threads
through two lazy combinators.  Every element requires forcing an `iota` thunk
and a `take` thunk, adding ~130 ticks of overhead per element.  A native BIF
that builds the list eagerly in Rust avoids this entirely.

### Change
- New `RANGE_LIST` BIF in `src/eval/stg/list.rs` that builds `[b, b+1, ..., e-1]`
  using `machine_return_num_list` (a single Rust allocation pass)
- Registered in `src/eval/intrinsics.rs` (index 189, strict args 0 and 1)
- Registered in `src/eval/stg/mod.rs`
- `lib/prelude.eu`: `range(b, e): ints-from(b) take(e - b)` → `range: '__RANGE_LIST'`
- Handles whole-number floats from arithmetic (serde_json stores results as f64;
  `to_range_int` accepts any finite float with zero fractional part)

### Results

| bench 007 (GC stress: short-lived, `sr(100)` × 200) | 468 ms | 215 ms | **−54%** |
| bench 008 (long-lived graph, uses `range`) | 213 ms | 119 ms | **−44%** |
| bench 009 (fragmentation, uses `range`) | 196 ms | 143 ms | **−27%** |

### Regression check

Full harness suite: **345/345 passed**, no regressions detected.

### Risks
- `range` is now strict in both arguments (eagerly forces `b` and `e`).
  Previously the lazy pipeline was equally strict at the point of consumption,
  so this is semantically equivalent for all normal use.
- Very large ranges allocate a `Vec<f64>` and a letrec STG tree upfront.
  This trades lazy allocation for a single large allocation, which is the
  correct trade-off for the typical case.

### DIFF
--- a/lib/prelude.eu
+++ b/lib/prelude.eu
@@ -1432,7 +1432,7 @@ ints-from: iota
 
 ` { doc: "`range(b, e)` - return list of ints from `b` to `e` (not including `e`)."
     type: "number → number → [number]" }
-range(b, e): ints-from(b) take(e - b)
+range: '__RANGE_LIST'
 
 ` { doc: "`count(l)` - return count of items in list `l`."
     type: "[a] → number" }
--- a/src/eval/intrinsics.rs
+++ b/src/eval/intrinsics.rs
@@ -994,6 +994,11 @@ lazy_static! {
             ty: function(vec![any(), any(), any()]).unwrap(),
             strict: vec![],
     },
+    Intrinsic { // 189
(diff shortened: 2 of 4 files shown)


gmorpheme:

PR #737, opened 2026-06-02
perf: native SUM_NUM_LIST and PRODUCT_NUM_LIST replace foldl aggregates

gmorpheme:
### Hypothesis
`sum: foldl(+, 0)` and `product: foldl(*, 1)` build an N-deep lazy
accumulator thunk chain before any arithmetic occurs.  A native BIF
using the `SeqNumList` + `collect_num_list` pipeline eliminates this
cascade entirely, performing the aggregation in a single tight Rust loop.

### Change
- `SUM_NUM_LIST` and `PRODUCT_NUM_LIST` BIFs in `src/eval/stg/running.rs`
  using the shared `num_list_wrapper` and `collect_num_list` infrastructure
- Both registered in `src/eval/intrinsics.rs` (indices 189, 190) and `src/eval/stg/mod.rs`
- `lib/prelude.eu`: `sum(l): foldl(+, 0, l)` → `sum: '__SUM_NUM_LIST'`
  and `product(l): foldl(*, 1, l)` → `product: '__PRODUCT_NUM_LIST'`
- Integer results use exact integer serialisation to match prior `foldl` output

### Results

| AoC day02 part-2 (heavy `sum`/`product`) | 184 ms | 171 ms | −7% |
| AoC day07 part-2 (`sum` in inner loop) | 4.50 s | 4.39 s | −2% |

Modest but consistent improvement across all code using these prelude functions.

### Regression check

### Risks
- `sum` and `product` are now strict in their argument: the entire list
  is forced before the operation begins.  Previously `foldl` was also
  strict (it traverses the whole list), so semantics are equivalent.
- Only works for numeric lists; non-numeric elements will produce a
  runtime error (same behaviour as `foldl(+, 0)` on non-numbers).

### DIFF
--- a/lib/prelude.eu
+++ b/lib/prelude.eu
@@ -886,11 +886,11 @@ bit: {
 
 ` { doc: "`sum(l)` - sum a list of numbers."
     type: "[number] → number" }
-sum(l): foldl(+, 0, l)
+sum: '__SUM_NUM_LIST'
 
 ` { doc: "`product(l)` - multiply a list of numbers."
     type: "[number] → number" }
-product(l): foldl(*, 1, l)
+product: '__PRODUCT_NUM_LIST'
 
 ` { doc: "`max(l, r)` - return max of numbers `l` and `r`."
     type: "number → number → number" }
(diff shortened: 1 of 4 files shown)


gmorpheme:

PR #738, opened 2026-06-02
perf: native COUNT_LIST intrinsic fixes O(N²) count scaling

gmorpheme:
### Hypothesis
`count: foldl({n: •el: •}.(n inc), 0)` builds an N-deep lazy accumulator
thunk chain before any arithmetic occurs.  Forcing the final result triggers
a cascade of N recursive thunk evaluations.  With N simultaneously-live
thunks, GC pressure amplifies the effect, producing ~O(N^1.7) tick scaling
instead of the expected O(N).

### Change
- `COUNT_LIST` BIF in `src/eval/stg/running.rs` using `SeqList` in the
  wrapper to force the list spine to WHNF, then a single O(N) Rust loop
- Registered in `src/eval/intrinsics.rs` (index 189) and `src/eval/stg/mod.rs`
- `lib/prelude.eu`: `count(l): foldl(...)` → `count: '__COUNT_LIST'`

### Results

| N     | Before (ticks) | After (ticks) | Improvement |
|-------|----------------|---------------|-------------|
|   100 |    24,469      |   14,978      | 1.6×        |
|   500 |   222,669      |   74,578      | 3×          |
| 1,000 |   695,419      |  149,078      | 4.7×        |
| 2,000 | 2,390,919      |  298,078      | 8×          |
| 5,000 | 13,477,419     |  745,078      | **18×**     |

The improvement grows with list size because the baseline has super-linear
scaling while the native implementation is strictly linear.

### Regression check

### Risks
- `count` now eagerly forces all list elements (via `SeqList`) before
  counting.  Previously `foldl` was also strict (it traverses the whole
  list), so semantics are equivalent.
- `SeqList` forces elements to WHNF — this is necessary to handle lazy
  lists (e.g. `repeat(x) take(N) count`).

### DIFF
--- a/lib/prelude.eu
+++ b/lib/prelude.eu
@@ -1436,7 +1436,7 @@ range(b, e): ints-from(b) take(e - b)
 
 ` { doc: "`count(l)` - return count of items in list `l`."
     type: "[a] → number" }
-count(l): foldl({ n: • el: •}.(n inc), 0, l)
+count: '__COUNT_LIST'
 
 ` { doc: "`last(l)` - return last element of list `l`."
     type: "[a] → a" }
--- a/src/eval/intrinsics.rs
+++ b/src/eval/intrinsics.rs
@@ -994,6 +994,11 @@ lazy_static! {
             ty: function(vec![any(), any(), any()]).unwrap(),
             strict: vec![],
     },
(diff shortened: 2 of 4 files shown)


gmorpheme:

PR #740, opened 2026-06-02
perf: native ALL_TRUE_LIST BIF replacing foldl(and, true, l)

gmorpheme:
### Hypothesis

`all-true?(l)` was implemented as `foldl(and, true, l)`.  `foldl` builds
a left-spine lazy accumulator thunk chain `and(and(and(true, h1), h2), h3)…`.
All N thunks exist simultaneously on the heap; the GC must traverse the
entire chain on each collection cycle, producing super-linear scaling.

Converting to a native BIF that uses `SeqList` to force the list upfront
then traverses it in a single O(N) Rust loop eliminates the thunk chain
entirely.

### Change

- Added `AllTrueList` intrinsic (BIF index 189) in `src/eval/stg/running.rs`
- Registered in `src/eval/stg/mod.rs` and `src/eval/intrinsics.rs`
- `all-true?` in `lib/prelude.eu` changed to `'__ALL_TRUE_LIST'`
- `any-true?` left as `foldr(or, false, l)` — `foldr` already short-circuits
  lazily from the left; a `SeqList`-based BIF would force the entire list
  before checking and would regress the common early-exit case (~28× slower
  when the first element matches)

### Results

Measured with `hyperfine` on `all-true?(range(1,N) map(_ > 0))`:

| N=1 000 (all-true) | 162 ms | 123 ms | −24% |
| N=5 000 (all-true) | 2.62 s | 1.55 s | −41% |

Known trade-off: `all-true?` with an early-false result is ~40% slower
(SeqList forces all elements before the Rust loop can short-circuit).
This is acceptable because `all-true?` is typically used in the
all-passing case (e.g. validating conditions) where the improvement
applies.

### Regression check

Full harness suite (345 tests): no regressions detected.

### Risks

`any-true?` is explicitly kept on its prelude `foldr` implementation —
any future attempt to native-BIF it must account for the short-circuit
regression demonstrated during development of this PR.

### DIFF
--- a/lib/prelude.eu
+++ b/lib/prelude.eu
@@ -1526,7 +1526,7 @@ reverse(l): foldl(cons flip, [], l)
 
 ` { doc: "`all-true?(l)` - true if and only if all items in list `l` are true."
     type: "[bool] → bool" }
-all-true?(l): foldl(and, true, l)
+all-true?: '__ALL_TRUE_LIST'
 
 ` { doc: "`all(p?, l)` - true if and only if all items in list `l` satisfy predicate `p?`."
     type: "(a → bool) → [a] → bool" }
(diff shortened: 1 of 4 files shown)


gmorpheme:

PR #741, opened 2026-06-02
perf(list): REVERSE_LIST intrinsic — eliminate O(N²) foldl thunk chain

gmorpheme:
### Hypothesis

The prelude's `reverse(l): foldl(cons flip, [], l)` builds an N-deep lazy
accumulator thunk chain.  On each GC cycle, the collector must traverse all
N thunks to mark live data.  For long lists this produces super-linear
(O(N²)) GC work.

A native BIF can force only the list spine and collect head closures directly
into a `Vec`, then reverse in a single O(N) pass with no accumulator thunks.

### Change

- **`src/eval/stg/force.rs`**: Add `SeqListSpine` — forces only the list
  spine (each cons cell's tail pointer) without touching head elements.
  Distinct from `SeqList` which strips `BoxedString`/`BoxedNumber`/
  `BoxedSymbol` wrappers and would break downstream BIFs like `str.letters`.

- **`src/eval/stg/list.rs`**: Add `ReverseList` BIF.  Wrapper applies
  `SeqListSpine` to guarantee `DataIterator` can traverse the spine, then
  `execute()` collects the original head closures, reverses, and returns via
  `machine_return_closure_list`.

- **`src/eval/intrinsics.rs`** / **`src/eval/stg/mod.rs`**: Register both
  `SeqListSpine` and `ReverseList`.

- **`lib/prelude.eu`**: Bind `reverse: '__REVERSE_LIST'`.

### Results

Benchmark: `count(reverse(range(1, 30000)))` — hyperfine, 5 runs

| count(reverse(range(1, 30000))) | 53.3 s ± 0.8 s | 20.9 s ± 0.3 s | **−61% (2.55×)** |
| count(reverse(range(1, 20000))) | 18.5 s | 7.5 s | **−59% (2.5×)** |
| count(reverse(range(1, 10000))) | 4.0 s | 1.76 s | **−56% (2.3×)** |

Note: wall-clock times include prelude parse; absolute times are dominated by
`count`'s own `foldl` accumulator (which has the same thunk-chain issue but
is not addressed here).  Removing `reverse`'s O(N²) contribution gives a
consistent ~2.5× speedup.

N=50000 baseline: exceeds 120 s timeout; new implementation: ~45 s.

### Regression check

Key tests verified:
- `test_harness_054` (qsort — uses `discriminate` → `reverse`) ✓
- `test_lib_lens` (lens operations using `reverse`) ✓

### Risks

- `SeqListSpine` adds a new recursive STG global.  It is structurally similar
  to `SeqNumList`/`SeqList` and has been validated against the full harness.
- `machine_return_closure_list` stores the original (unstripped) head closures.
  These work correctly with all subsequent operations because `SeqListSpine`
  preserves the original boxed representations.

### DIFF
--- a/lib/prelude.eu
+++ b/lib/prelude.eu
@@ -1522,7 +1522,7 @@ zip-apply(fs, vs): zip-with(_0(_1), fs, vs)
 
 ` { doc: "`reverse(l)` - reverse list `l`."
     type: "[a] → [a]" }
-reverse(l): foldl(cons flip, [], l)
+reverse: '__REVERSE_LIST'
 
 ` { doc: "`all-true?(l)` - true if and only if all items in list `l` are true."
     type: "[bool] → bool" }
--- a/src/eval/intrinsics.rs
+++ b/src/eval/intrinsics.rs
@@ -994,6 +994,16 @@ lazy_static! {
             ty: function(vec![any(), any(), any()]).unwrap(),
             strict: vec![],
     },
+    Intrinsic { // 189
+            name: "REVERSE_LIST",
+            ty: function(vec![list(), list()]).unwrap(),
+            strict: vec![],
+    },
+    Intrinsic { // 190
+            name: "seqListSpine",
+            ty: function(vec![list(), list()]).unwrap(),
+            strict: vec![0],
+    },
     ];
 }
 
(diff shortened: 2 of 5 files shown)


gmorpheme:

PR #742, opened 2026-06-02
perf(arith): SUM_LIST + PRODUCT_LIST intrinsics — 53× speedup on sum

gmorpheme:
### Hypothesis

The prelude's `sum(l): foldl(+, 0, l)` and `product(l): foldl(*, 1, l)`
build an N-deep lazy accumulator thunk chain.  On each GC cycle the
collector traverses all N thunks to mark live data — O(N²) total GC work.

Native BIFs using `SeqNumList` (which forces and unboxes all elements) can
then fold in a single Rust pass: O(N) total, zero heap thunks after the
initial pass.

### Change

- **`src/eval/stg/running.rs`**: Add `SumList` and `ProductList` BIFs, both
  using `num_list_wrapper` (SeqNumList → BIF).  A `f64_to_number` helper
  normalises the result to an integer `Number` for whole-number values,
  matching the behaviour of arithmetic operators (which try integer paths
  first).

- **`src/eval/intrinsics.rs`** / **`src/eval/stg/mod.rs`**: Register both.

- **`lib/prelude.eu`**: `sum: '__SUM_LIST'`, `product: '__PRODUCT_LIST'`.

### Results

Benchmark: `sum(range(1, 20000))` — hyperfine, 5 runs, warm cache

| `sum(range(1, 20000))` | 9.94 s ± 0.26 s | 186 ms ± 2 ms | **−98% (53×)** |
| `sum(range(1, 50000))` | >60 s (timeout) | 0.3 s | **>200×** |

The scaling goes from O(N²) (GC thunk chain traversal) to O(N) (single Rust
fold + SeqNumList pass).

### Regression check

Full harness suite: **349/349 passed**, no regressions.

### Risks

- `f64_to_number` truncates: `sum([1.5, 1.4])` = 2.9 returns a float; but
  `sum([1.0, 2.0])` = 3.0 returns integer `3`.  This matches how `+` itself
  behaves (tries integer first).
- Empty list: `sum([])` = `0`, `product([])` = `1` — correct identity values.

### DIFF
--- a/lib/prelude.eu
+++ b/lib/prelude.eu
@@ -886,11 +886,11 @@ bit: {
 
 ` { doc: "`sum(l)` - sum a list of numbers."
     type: "[number] → number" }
-sum(l): foldl(+, 0, l)
+sum: '__SUM_LIST'
 
 ` { doc: "`product(l)` - multiply a list of numbers."
     type: "[number] → number" }
-product(l): foldl(*, 1, l)
+product: '__PRODUCT_LIST'
 
 ` { doc: "`max(l, r)` - return max of numbers `l` and `r`."
     type: "number → number → number" }
--- a/src/eval/intrinsics.rs
+++ b/src/eval/intrinsics.rs
@@ -994,6 +994,16 @@ lazy_static! {
             ty: function(vec![any(), any(), any()]).unwrap(),
             strict: vec![],
     },
+    Intrinsic { // 189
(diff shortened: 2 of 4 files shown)


gmorpheme:
Closing as duplicate: PRs #727 and #737 already cover SUM and PRODUCT intrinsics from earlier sessions.

PR #743, opened 2026-06-02
perf: SLIDING_SUM_NUM_LIST intrinsic for O(n) sliding window sums

gmorpheme:
### Hypothesis

The eucalypt pipeline `window(n, 1) map(sum)` is O(L×n) for a list of length L with window size n. For each of the L−n+1 windows, `window` builds a new cons list (O(n) allocation), then `sum` folds it (O(n)). For AoC 2025 day04 which convolves a 135×135 grid with window size 3, this runs 135 rows × 132 windows × 2 O(n) passes = ~107k O(3) operations per call to `h-sums`, totalling ~56M VM ticks.

A rolling-sum BIF can compute the same result in a single O(L) pass with no per-element heap allocation.

### Change

- `src/eval/stg/running.rs`: adds `SlidingSumNumList` BIF — forces the list via `SeqNumList`, unboxes the window size `n` via `unbox_num`, then runs a rolling sum in Rust.
- `src/eval/intrinsics.rs`: registers the new BIF at index 189.
- `src/eval/stg/mod.rs`: adds the BIF to the standard runtime.
- `lib/prelude.eu`: exposes `list sliding-sum(n)` (defined as `sliding-sum(n, list)` following eucalypt's catenation convention where the receiver is the last argument).
- `examples/aoc25/day04.eu`: replaces `row pad window(3, 1) map(sum)` with `row pad sliding-sum(3)`.

### Results

Measured on AoC 2025 day04 part-1 (135×135 grid, window size 3):

| Machine Ticks | 56.6M | 21.2M | −63% |
| VM-Total time | 1.99s | 0.71s | −64% |

Correctness: `day04.eu` test target passes (`part-1: true`, `part-2: true`).

### Regression check

Full harness suite: no regressions detected (349/349 tests pass).

| Test category | Result |
|--------------|--------|
| `cargo test --lib` | 924 passed |
| `cargo test --test harness_test` | 349 passed |
| `cargo clippy --all-targets -- -D warnings` | no warnings |

### Risks

- The BIF wrapper uses `unbox_num` to handle the window-size argument arriving as a `BoxedNumber`. This mirrors the pattern used by the standard `wrap.rs` wrapper for numeric arguments.
- The `SeqNumList` pre-pass forces and normalises the input list spine; `collect_num_list` then handles both boxed and native number elements.
- Part-2 of day04 (iterative simulation) also calls `block-sums` → `h-sums`, so it benefits too, though the bottleneck there is the iteration count not the windowing.

### DIFF
--- a/examples/aoc25/day04.eu
+++ b/examples/aoc25/day04.eu
@@ -32,7 +32,7 @@ pad(row): [0] ++ row ++ [0]
 zero-row(row): row map(const(0))
 
 ` "Horizontal sums: for each position, sum of the 3-wide window."
-h-sums(row): row pad window(3, 1) map(sum)
+h-sums(row): row pad sliding-sum(3)
 
 ` "Element-wise sum of two lists."
 add-rows(a, b): zip-with(+, a, b)
--- a/lib/prelude.eu
+++ b/lib/prelude.eu
@@ -1640,6 +1640,10 @@ running-min: '__RUNNING_MIN'
 `[3, 1, 4, 1]` → `[3, 4, 8, 9]`. (Rust-level intrinsic)."
 running-sum: '__RUNNING_SUM'
 
+` { doc: "nums sliding-sum(n) - sliding window sums of size n over number list nums. Returns a list of length max(0, len(nums) - n + 1). Equivalent to window(n, 1) map(sum) but O(n) time."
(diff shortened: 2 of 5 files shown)


gmorpheme:

PR #746, opened 2026-06-03
docs: fix cheat sheet errors and add cond/=> coverage

gmorpheme:
## Summary

- **cheat-sheet**: Remove four non-existent prelude functions (`flatten`, `unique`, `select`, `dissoc`) and replace with correct alternatives (`concat`, `any`/`all`, `nub-by`, `filter-items`). These were contradicting agent-reference.md §5.11, causing agent confusion.
- **agent-reference**: Add missing `=>` / `⇒` clause operator at precedence 15 to the operator precedence table. The cheat sheet had it; agent-reference did not.
- **syntax-gotchas**: Add `=>` / `⇒` at precedence 15 to the simplified overview table.
- **eucalypt-style**: Add `cond` / `=>` multi-way conditional style section under Conditionals, covering when to prefer `cond` over nested `if`, type-checker narrowing integration, and the `⇒` Unicode alias.

Closes eu-0rb9.4, eu-0rb9.5, eu-0rb9.6, eu-0rb9.8

## Test plan

- [ ] Docs-only change — verify no broken markdown links or formatting issues
- [ ] Check that the cheat sheet no longer lists `flatten`, `unique`, `select`, `dissoc`
- [ ] Confirm `=>` appears at precedence 15 in both agent-reference and syntax-gotchas tables
- [ ] Confirm new `cond` style section renders correctly

### DIFF
--- a/docs/appendices/cheat-sheet.md
+++ b/docs/appendices/cheat-sheet.md
@@ -312,14 +312,14 @@ Set custom values via metadata: `` ` { precedence: 75 associates: :right } ``
 | `drop(n)` | Remove first n |
 | `zip` | Pair elements from two lists |
 | `zip-with(f)` | Combine elements with function |
-| `flatten` | Flatten nested lists one level |
+| `concat` | Flatten nested lists one level (use `concat`, not `flatten` — does not exist) |
 | `reverse` | Reverse a list |
 | `count` | Number of elements |
 | `range(a, b)` | Integers from a to b-1 |
 | `nil?` | Is the list empty? |
-| `any?(p?)` | Does any element match? |
-| `all?(p?)` | Do all elements match? |
-| `unique` | Remove duplicates |
+| `any(p?)` | Does any element match? |
+| `all(p?)` | Do all elements match? |
+| `nub-by(f)` | Deduplicate by key function (no `unique` in prelude) |
 
 ### Blocks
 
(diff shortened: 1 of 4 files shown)


gmorpheme:
## Review — Changes Required

CI is all green. The cheat sheet function corrections are all verified accurate against `lib/prelude.eu`. However, I found one factual error that must be fixed before merging.

### Factual error: `=>` associativity is left, not right

Both `docs/reference/agent-reference.md` and `docs/eucalypt-style.md` document `=>` as **right-associative**, but this is incorrect.

In `lib/prelude.eu`, the `=>` operator is declared as:
```eu
` { doc: "..." precedence: 15 }
(c => r): __CLAUSE(c, r)
```

No `associates:` key is present. The desugarer in `src/core/desugar/rowan_ast.rs:2614` uses `.unwrap_or(Fixity::InfixLeft)` as the default — so `=>` is **left-associative** by default.

**Specific fixes required:**

1. `docs/reference/agent-reference.md` — change `right` to `left` in the `=>` row:
   ```
   | 15 | clause | left | `=>`, `⇒` | Cond clause builder (`condition => result`) |
   ```

2. `docs/eucalypt-style.md` — change "right-associative" to "left-associative":
   ```
   operator (precedence 15, left-associative; Unicode alias `⇒`) builds a
   ```

`docs/appendices/syntax-gotchas.md` has no associativity column so requires no change.

gmorpheme:
## Supplementary review notes (minor)

These are non-blocking but worth addressing in the same pass:

1. **`docs/appendices/syntax-gotchas.md`** — the precedence overview table has no associativity column. The agent-reference.md table does. Consider adding one to syntax-gotchas.md for consistency (would also document `=>` as `left` there).

2. **`docs/eucalypt-style.md` cond section** — the guide says "The last list element is the default" but does not clarify whether the default element is mandatory or optional. Worth a brief note.

3. **`docs/appendices/cheat-sheet.md`** — the parenthetical correction notes (e.g. `use concat, not flatten — does not exist`) are helpful but slightly verbose for a cheat sheet. Consider trimming once the factual error above is fixed.

The associativity fix from the first comment is the hard blocker; these are discretionary.

PR #755, opened 2026-06-03
perf(memory): skip zero-init in RawArray backing allocations

gmorpheme:
### Hypothesis
`alloc_bytes` zero-initialises every byte of newly allocated backing buffers for `Array<T>`.
This is unnecessary: the GC only ever scans indices `[0, length)`, and all callers either
copy full data over the allocation immediately (`with_data`) or push elements incrementally,
advancing `length` only after writing each slot (`push`).
Removing the zero-fill should save proportional allocation cost, most visible in GC-heavy
workloads that create many small arrays.

### Change
- Added `alloc_bytes_uninit` to `Heap` (no zero-fill loop), `Allocator` trait, and
  `ScopedAllocator` trait (default implementation falls back to the zeroing path, keeping future
  implementors safe).
- Overrode `alloc_bytes_uninit` in `MutatorHeapView` to delegate to `Heap::alloc_bytes_uninit`.
- Changed `RawArray::alloc` (the sole `alloc_bytes` caller via `ScopedAllocator`) to use
  `alloc_bytes_uninit`.

Safety invariant documented on `alloc_bytes_uninit`: *callers must ensure every byte is written
before it is read.*  Both call paths satisfy this:
- `with_data` copies `data.len()` elements immediately after allocation.
- `with_capacity` / `push` only increments `length` after writing the element; the GC scans
  `[0, length)` only.

Recycled heap blocks are not zeroed between GC cycles, so `alloc_bytes` was re-zeroing memory
that had simply been in use — this was the main source of overhead.

### Results
Criterion micro-benchmarks (absolute times, 100 samples each):

| gc_alloc_then_collect/256 | 24.36 µs | 21.77 µs | **−10.6%** |
| gc_alloc_then_collect/1024 | 95.05 µs | 85.51 µs | **−10.0%** |
| gc_alloc_then_collect/4096 | 376.9 µs | 338.8 µs | **−10.1%** |
| create_partially_apply_and_saturate_lambda | 442 ns | 432 ns | −2.3% |
| alloc_letrec | 140 ns | 140 ns | no change |

End-to-end wall-clock benchmarks show ~1% improvement, diluted by the fixed ~40 ms pipeline
compilation cost that dominates short programmes.

### Regression check
Full harness suite: 359 tests, 0 failures.

### Risks
- Uninitialized memory beyond `length` is never read, but a future caller of `alloc_bytes_uninit`
  that forgets to initialise could cause undefined behaviour. Mitigated by: the safe default
  implementation on `ScopedAllocator`, and the documented safety invariant on both the trait
  method and `Heap::alloc_bytes_uninit`.
- The `EU_GC_POISON=1` debug mode already detected use-after-free before this change. It
  continues to work: poisoning happens on reclaim, not on allocation, so the uninit path has no
  interaction with the poison check.

### DIFF
--- a/src/eval/memory/alloc.rs
+++ b/src/eval/memory/alloc.rs
@@ -23,8 +23,13 @@ pub trait Allocator {
     where
         T: StgObject;
 
-    /// Allocate a region of bytes
+    /// Allocate a region of bytes (zero-initialised)
     fn alloc_bytes(&self, size_bytes: usize) -> Result<NonNull<u8>, HeapError>;
+
+    /// Allocate a region of bytes without zero-initialisation.
+    ///
+    /// Callers must ensure every byte is written before it is read.
+    fn alloc_bytes_uninit(&self, size_bytes: usize) -> Result<NonNull<u8>, HeapError>;
 
     /// Get header from object
     fn get_header<T>(&self, object: NonNull<T>) -> NonNull<AllocHeader>;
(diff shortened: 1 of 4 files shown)


gmorpheme:
## Sending back — formatting failure

CI lint check fails with a  error in : missing trailing comma in the last argument to .

```
Diff in src/eval/memory/array.rs:128:
-                mem.alloc_bytes_uninit(capacity_bytes)?.as_ptr() as *mut T
+                mem.alloc_bytes_uninit(capacity_bytes)?.as_ptr() as *mut T,
```

Run `cargo fmt --all` and re-push. No other issues were found — the safety review is ready once CI is green.

PR #773, opened 2026-06-08
test(regression): string interpolation de Bruijn index in blocks (eu-ynhu)

gmorpheme:
## Summary

- Adds harness test 154 as a regression guard for eu-ynhu: string interpolation variable resolution in block bodies with 3+ bindings
- The bug (de Bruijn index shifted by block structure Pair/Cons entries entering the same letrec) was already correct on `integration/0.7.0`; this PR ensures it stays fixed
- Tests all relevant variants: 3-binding and 4-binding blocks, format specifiers (`%02d`, `%03d`), interpolation of non-first bindings, map-based invocations, and the 2-binding baseline

## Investigation

On `integration/0.7.0`, the STG compiler correctly isolates block structure entries (Pair/Cons) from the value bindings in the letrec. The `var_refs` mapping (core binder index → letrec local index) correctly points each binding's value to its letrec slot regardless of subsequent structure entries. The bead description noted this was pre-existing in 0.6.2, and no changes to `src/eval/stg/compiler.rs` or `src/core/desugar/` have been made since 0.6.2 — indicating the bug may have been fixed in an earlier cycle or the current code never exhibited it on this branch.

## Test plan

- [x] `cargo test test_harness_154` passes
- [x] `cargo test` full suite: 104/104 harness + all unit tests pass
- [x] `cargo fmt --all` clean
- [x] `timeout 60 eu check lib/prelude.eu` produces no new warnings

### DIFF
--- a/tests/harness/154_interp_block_debruijn.eu
+++ b/tests/harness/154_interp_block_debruijn.eu
@@ -1,0 +1,77 @@
+"154 string interpolation de Bruijn index — block with 3+ bindings (eu-ynhu regression)"
+
+# Regression test for eu-ynhu: string interpolation in a block body should
+# correctly resolve the interpolated variable regardless of how many bindings
+# precede it in the same block.
+#
+# The bug: compile_block adds Pair/Cons entries to the same LetBinder used by
+# the binding values. With 3+ bindings, the letrec indices for the block
+# structure entries shift past the value entries, causing interpolation
+# variable references (resolved as de Bruijn indices) to point at the wrong
+# letrec entry at runtime.
+
+` { target: :test }
+test: {
+
+  # --- 3-binding case (minimal reproduction from the bug report) ---
+  f-three(pair): {
+    i: pair first
+    x: 42
+    result: "component_{i:%02d}"
+  }.result
+
+  three-ok: f-three([1, "a"]) = "component_01"
+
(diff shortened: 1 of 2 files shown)


gmorpheme:
**The simplified test passes but the original bug still reproduces.** The key is the **outer block-dot nesting** — the function with interpolation must be inside a block that also uses `.(...)`:

```eu
to-input-block(component-elements): {
  one-component-block(pair): {
    i: pair first
    l: (pair second) first
    s: (pair second) second
    one-component-key: "component_{i:%02d}"
    one-weight-key: "weight_{i:%02d}"
    ret: [[one-component-key sym, l str.of], [one-weight-key sym, s]] block
  }.ret
  default: { component_01: "None", weight_01: 0 }
  input-block: component-elements zip(iota(1)) map(one-component-block) take(4) foldl(merge, {})
}.(default input-block)

test: [["c1", 0.0], ["c2", 0.0]] to-input-block
```

This fails. Remove the outer block (make `one-component-block` top-level) and it works. The test in this PR doesn't capture the nesting depth that triggers the bug.

gmorpheme:
Superseded by PR #774 which fixes the actual root cause (FMT wrapper missing native fallback, not a de Bruijn index issue).

PR #774, opened 2026-06-08
fix(stg): add native-value fallback to FMT wrapper

gmorpheme:
## Summary

- The FMT intrinsic wrapper's case expression used `switch` (no fallback) to dispatch on its value argument, expecting only boxed data constructors (BoxedNumber, BoxedString, etc.)
- Arithmetic intrinsics (ADD, SUB, etc.) return unboxed native values via `machine_return_num`. When stored through a thunk update (e.g. from `iota(n+1)`), subsequent accesses yield raw natives rather than boxed constructors
- This caused `NoBranchForNative` errors when string interpolation with format specifiers (`"{i:%02d}"`) was applied to lazy computed values (from `iota`, `zip`, or any arithmetic thunk)
- Fix: replace `switch` with `case` (adding native fallbacks), mirroring the approach already used by the STR wrapper

Fixes eu-ynhu.

## Test plan

- [x] New harness test `154_fmt_lazy_values.eu` covering iota, zip, nested block-dot, and plain interpolation cases
- [x] `/tmp/fmtnested.eu` (original reproduction) passes
- [x] `~/dev/test/test.eu` passes
- [x] All 104 existing tests pass
- [x] Clippy clean
- [x] `eu check lib/prelude.eu` produces no warnings


gmorpheme:
Combined into PR #775 which includes the fix, harness test, audit results, doc comments, and unit tests.

PR #783, opened 2026-06-09
fix(errors): clear diagnostics for bracket content mode mismatches

gmorpheme:
## Summary

- **Case a** (block content in soup bracket): `⟦ a: 1 ⟧` where `⟦ x ⟧: x` is declared. The old error was the opaque "monadic block used without a monad spec". The new error is "block-style content inside bracket pair '⟦⟧'" with notes explaining how to pass a block literal or convert to a monad bracket.

- **Case b** (soup content in monad bracket): `⟦ 1 2 3 ⟧` where `⟦{}⟧` is declared. The old error was a confusing runtime "type mismatch: expected block, found list". The new error is caught at desugar time with "expression content inside monad bracket '⟦⟧'" and notes explaining that the bracket needs `name: value` declarations.

Both errors have source locations pointing to the bracket usage site.

## Changes

- `src/core/error.rs`: Updated `NoMonadSpec` error message and added `SoupContentInMonadBracket` variant with diagnostic notes for both cases
- `src/core/desugar/rowan_ast.rs`: Added monad-spec check in `BracketExpr` desugaring to detect case b at compile time
- `tests/harness/errors/155_block_content_in_soup_bracket.eu` + `.expect`: Error test for case a
- `tests/harness/errors/156_soup_content_in_monad_bracket.eu` + `.expect`: Error test for case b

## Test plan

- [ ] `cargo test test_error_155` — case a produces clear error
- [ ] `cargo test test_error_156` — case b produces clear error
- [ ] `cargo test test_harness_09` — existing soup bracket tests still pass
- [ ] `cargo test test_harness_14` — existing monad bracket tests still pass
- [ ] `cargo clippy --all-targets -- -D warnings` — no warnings

### DIFF
--- a/tests/harness/errors/155_block_content_in_soup_bracket.eu.expect
+++ b/tests/harness/errors/155_block_content_in_soup_bracket.eu.expect
@@ -1,0 +1,2 @@
+exit: 1
+stderr: "block-style content inside bracket pair"
--- a/tests/harness/errors/155_block_content_in_soup_bracket.eu.expect
+++ b/tests/harness/errors/155_block_content_in_soup_bracket.eu.expect
@@ -1,2 +1,2 @@
 exit: 1
-stderr: "block-style content inside bracket pair"
+stderr: "block-style content inside non-monad bracket pair"
--- a/src/core/desugar/rowan_ast.rs
+++ b/src/core/desugar/rowan_ast.rs
@@ -1449,6 +1449,13 @@ impl Desugarable for Element {
                     )
                 })?;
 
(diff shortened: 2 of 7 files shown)


tests/harness/errors/155_block_content_in_soup_bracket.eu.expect
```diff
@@ -0,0 +1,2 @@
+exit: 1
+stderr: "block-style content inside bracket pair"
```
gmorpheme: non-monad bracket pair

PR #789, opened 2026-06-11
feat(test): conformance corpus with golden sidecars and GC-verified CI job

gmorpheme:
## Summary

- Adds `tests/conform/` with 20 conformance `.eu` files + `.golden` sidecars, covering arithmetic, list/block/string ops, booleans, `cond/=>`, sort order, merge semantics (shallow + deep), operator precedence, and all export formats (yaml, json, toml, text, edn)
- Adds `tests/conform_test.rs`: discovers `.eu` files dynamically, runs `eu -t conform`, compares stdout byte-for-byte against `.golden`; `BLESS=1` regenerates; missing golden = hard failure; mismatches show unified diff via `similar`
- Adds `test-gc-verified` CI job: runs full harness under `EU_GC_VERIFY=2` + `EU_GC_STRESS=1` + `EU_GC_POISON=1` on ubuntu-latest x86-64
- Adds `tests/conform/CONFORMANCE.md` documenting the output-stability contract

## Test plan

- [x] `cargo test --test conform_test` — all 20 conformance tests pass
- [x] `cargo fmt --all -- --check` — clean
- [x] `BLESS=1 cargo test --test conform_test` — regenerates golden files successfully

Closes bead eu-yhk0.3.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+tests/conform/*.golden text eol=lf
--- a/.github/workflows/build-rust.yaml
+++ b/.github/workflows/build-rust.yaml
@@ -68,6 +68,21 @@ jobs:
           python3 scripts/test-doc-examples.py --eu ./target/release/eu --timeout 30
 
 
+  test-gc-verified:
+    name: GC-verified harness (x86-64)
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@stable
+      - name: Build release binary
+        run: cargo build --release
+      - name: Run harness under GC verification
+        run: ./target/release/eu test --allow-io tests/harness
+        env:
+          EU_GC_VERIFY: "2"
+          EU_GC_STRESS: "1"
+          EU_GC_POISON: "1"
+
   test-aarch64-sequential:
     name: Sequential harness (aarch64 release)
     runs-on: ubuntu-24.04-arm
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -548,7 +548,7 @@ checksum = "877a4ace8713b0bcf2a4e7eec82529c029f1d0619886d18145fea96c3ffe5c0f"
 
 [[package]]
 name = "eucalypt"
-version = "0.7.0"
+version = "0.7.1"
 dependencies = [
  "base64",
  "bitflags 1.3.2",
@@ -584,6 +584,7 @@ dependencies = [
  "serde",
  "serde_json",
  "sha2",
+ "similar",
  "thiserror",
  "toml",
  "unicode-bidi-mirroring",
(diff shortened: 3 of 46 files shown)


gmorpheme:
## Wicket review — SEND BACK: Windows CI failure (CRLF)

**Status: Cannot merge — `Test Suite (windows-latest)` fails.**

All 20 conformance tests fail on Windows with "output mismatch". The run is at job ID 80795943804.

### Root cause: CRLF line endings

`eu` on Windows writes `\r\n` to stdout (text mode). Golden files committed from Linux have `\n`. The `assert_conform` comparison sees every line as different.

Evidence from the CI log:
```
20 conformance failure(s):
tests/conform\arithmetic.eu: output mismatch
@@ -1,10 +1,10 @@
tests/conform\block_ops.eu: output mismatch
@@ -1,20 +1,20 @@
... (all 20 files)
```

### Fix required in `tests/conform_test.rs`

Add line ending normalisation before comparison:

```rust
fn assert_conform(eu_file: &Path) -> Result<(), String> {
    ...
    let actual = run_eu_conform(eu_file)
        .map_err(...)?;
    let actual = actual.replace("\r\n", "\n");  // ← add this
    ...
    let golden = std::fs::read_to_string(&golden_path)...;
    // (golden files have LF — no translation needed, but could normalise for defence-in-depth)
    ...
}
```

Only `actual` needs normalising (golden files are committed with LF; `read_to_string` on Windows does not translate LF→CRLF).

### Everything else is correct

- GC-verified CI job ✓ (passes on ubuntu)
- 20 conformance files with golden sidecars ✓
- `BLESS=1` regeneration logic ✓
- `similar` unified diff ✓
- Missing golden = hard failure ✓
- `.eu` syntax is valid ✓
- Spec acceptance criteria 1–9 are met except Windows portability (AC3/AC6)

Please push the one-line fix and CI should go green.

gmorpheme:
## Still failing on Windows — golden CRLF not normalised

The CRLF fix normalised `actual` but not the golden file content. On Windows, git's `core.autocrlf=true` converts LF → CRLF on checkout, so `std::fs::read_to_string` returns `\r\n`-terminated lines from the golden file. The comparison then fails on every line because `actual` has `\n` endings and `golden` has `\r\n`.

**Fix — normalise golden as well:**

In `tests/conform_test.rs`, change the golden read from:

```rust
let golden = std::fs::read_to_string(&golden_path)
    .unwrap_or_else(|e| panic!("failed to read {}: {e}", golden_path.display()));
```

to:

```rust
let golden = std::fs::read_to_string(&golden_path)
    .unwrap_or_else(|e| panic!("failed to read {}: {e}", golden_path.display()))
    .replace("\r\n", "\n");
```

**Recommended additionally:** add to `.gitattributes`:

```
tests/conform/*.golden text eol=lf
```

This stops git from converting golden files to CRLF on Windows checkout in the first place, so the `.replace` becomes a defensive fallback rather than the only line of defence.

PR #790, opened 2026-06-11
feat(desugar): trace metadata extraction and body wrapping

gmorpheme:
## Summary

Implements the desugar-phase side of declaration trace metadata (eu-yhk0.4b).

- `TraceSpec` enum (`Lazy`/`Strict`/`Exit`/`StrictExit`) added to `metadata.rs`
- `:trace` symbol shorthand normalised to `{ trace: :lazy }` in `normalise_metadata()`
- `trace: Option<TraceSpec>` added to `DesugarPhaseDeclarationMetadata`
- `"trace"` key stripped from metadata after extraction (alongside fixity, target, etc.)
- Declaration bodies wrapped in `__TRACE_ENTRY`/`__TRACE_EXIT` intrinsic calls:
  - `Lazy`: `__TRACE_ENTRY(name, arg_pairs, false, body)`
  - `Strict`: `__TRACE_ENTRY(name, arg_pairs, true, body)`
  - `Exit`: `__TRACE_EXIT(name, __TRACE_ENTRY(name, arg_pairs, false, body), false)`
  - `StrictExit`: `__TRACE_EXIT(name, __TRACE_ENTRY(name, arg_pairs, true, body), true)`

The wrapping inserts inside the lambda body so argument variables are in scope.

The `__TRACE_ENTRY`/`__TRACE_EXIT` intrinsics are being implemented by Furnace in a companion PR. This PR compiles and desugars correctly but is not end-to-end testable until those intrinsics land.

## Test plan

- [ ] `cargo build` passes
- [ ] `cargo clippy --all-targets -- -D warnings` passes clean
- [ ] `cargo test --lib` — all 973 tests pass
- [ ] `` ` :trace \n f(x): x `` desugars to `__TRACE_ENTRY("f", ["x", x], false, x)` inside lambda
- [ ] `` ` { trace: :strict-exit } \n g(x, y): x `` desugars to `__TRACE_EXIT("g", __TRACE_ENTRY("g", ["x", x, "y", y], true, x), true)`
- [ ] Existing tests unaffected (no regressions)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #794, opened 2026-06-11
fix(stg): trace intrinsic runtime fixes and harness tests for all four trace modes (eu-yhk0.4)

gmorpheme:
## Summary

- Fixes three runtime bugs in the `:trace` metadata intrinsic implementation
- Adds harness tests 159–162 covering all four trace modes

## Bug fixes

**1. BIF name prefix (`rowan_ast.rs`)**  
`core::bif(smid, "__TRACE_ENTRY")` produces an `Intrinsic` node with the literal name `"__TRACE_ENTRY"`.  The `__` stripping only happens for identifier nodes in eucalypt source — not for `core::bif()` call-sites.  The intrinsics catalogue registers `"TRACE_ENTRY"` (no prefix), so the lookup failed with "unknown intrinsic".  Fixed: remove the `__` prefix from the `core::bif()` calls.

**2. Entry/exit wrapping order (`rowan_ast.rs`)**  
The original wrapping was `TRACE_EXIT(name, TRACE_ENTRY(name, args, strict, body), strict)`.  Because `TRACE_EXIT`'s `value` argument is lazy, `TRACE_EXIT::execute` fires first (printing `←`) and only forces the inner `TRACE_ENTRY` thunk later.  This caused exit to appear before entry in the output.  Fixed: wrap in the correct order — `TRACE_ENTRY(name, args, strict, TRACE_EXIT(name, body, strict))` so entry always fires before exit.

**3. `navigate_local` panic for non-local refs (`debug.rs`)**  
`get_list_element_at` and `count_list_elements` called `current.navigate_local(&view, t_ref)` to traverse cons cells.  `navigate_local` only handles `Ref::L`; it panics for `Ref::G` or `Ref::V`.  Cons cells built by `core::list()` in the desugar contain non-local refs, triggering the panic on the very first cell.  Fixed: use `machine.nav(view).resolve_in_closure(&current, ref)` which handles all three ref variants.

## Harness tests

| Test | Mode | What it verifies |
|------|------|-----------------|
| 159 | `` ` :trace `` (lazy entry) | `→ add(x: <thunk>, y: <thunk>)` on stderr |
| 160 | `{ trace: :strict }` (strict entry) | `→ add(x: 1, y: 2)` on stderr |
| 161 | `{ trace: :exit }` (lazy entry+exit) | both `→` and `← add: <thunk>` on stderr |
| 162 | `{ trace: :strict-exit }` (strict entry+exit) | `→ add(x: 1, y: 2)` and `← add: 3` on stderr |

## Test plan

- [x] `cargo test test_harness_159` passes
- [x] `cargo test test_harness_160` passes
- [x] `cargo test test_harness_161` passes
- [x] `cargo test test_harness_162` passes

### DIFF
--- a/src/core/desugar/rowan_ast.rs
+++ b/src/core/desugar/rowan_ast.rs
@@ -2634,8 +2634,12 @@ fn rowan_declaration_to_binding(
     // be bound when the lambda closes over them.
     //
     // Design:
-    //   entry-only  →  __TRACE_ENTRY(name, arg_pairs, is_strict, body)
-    //   entry+exit  →  __TRACE_EXIT(name, __TRACE_ENTRY(name, arg_pairs, is_strict, body), is_strict)
+    //   entry-only  →  TRACE_ENTRY(name, arg_pairs, is_strict, body)
+    //   entry+exit  →  TRACE_ENTRY(name, arg_pairs, is_strict, TRACE_EXIT(name, body, is_strict))
+    //
+    // TRACE_ENTRY fires first (it is the outer call) then returns its 4th arg.
+    // For entry+exit mode the 4th arg is TRACE_EXIT(…), which fires when the
+    // result is forced.  This guarantees entry is always logged before exit.
     if let Some(ref trace_spec) = metadata.trace {
         let smid = desugarer.new_smid(components.span);
         let name_str = core::str(smid, &components.name);
@@ -2652,21 +2656,21 @@ fn rowan_declaration_to_binding(
         let with_exit = matches!(trace_spec, TraceSpec::Exit | TraceSpec::StrictExit);
         let strict_bool = core::bool_(smid, is_strict);
 
-        // __TRACE_ENTRY(name, arg_pairs, is_strict, body)
-        expr = core::app(
-            smid,
-            core::bif(smid, "__TRACE_ENTRY"),
-            vec![name_str.clone(), arg_pairs_list, strict_bool.clone(), expr],
-        );
-
-        // __TRACE_EXIT(name, traced_body, is_strict)
+        // For entry+exit: wrap body in TRACE_EXIT first, then wrap in TRACE_ENTRY.
         if with_exit {
             expr = core::app(
                 smid,
-                core::bif(smid, "__TRACE_EXIT"),
-                vec![name_str, expr, strict_bool],
+                core::bif(smid, "TRACE_EXIT"),
+                vec![name_str.clone(), expr, strict_bool.clone()],
             );
         }
+
+        // TRACE_ENTRY(name, arg_pairs, is_strict, body_or_exit_wrapped_body)
+        expr = core::app(
+            smid,
+            core::bif(smid, "TRACE_ENTRY"),
+            vec![name_str, arg_pairs_list, strict_bool, expr],
+        );
     }
 
     // Wrap in lambda if there are arguments
--- a/src/core/desugar/rowan_ast.rs
+++ b/src/core/desugar/rowan_ast.rs
@@ -8,7 +8,7 @@ use super::{
 };
 use crate::core::metadata::{
     normalise_metadata, strip_desugar_phase_metadata, DesugarPhaseBlockMetadata,
-    DesugarPhaseDeclarationMetadata, ReadMetadata,
+    DesugarPhaseDeclarationMetadata, ReadMetadata, TraceSpec,
 };
 use crate::{
     common::sourcemap::{HasSmid, Smid},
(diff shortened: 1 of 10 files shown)


gmorpheme:
## Review: **Send Back**

The logic in this PR is largely correct — the 4-arg `TRACE_ENTRY` design, wrapping order, and `resolve_in_closure` GC-safety fix are all right. But there are two blocking issues before this can merge.

---

### Issue 1 — Merge conflict: must rebase onto master

The branch was created before PR #791 (trace intrinsics) and PR #792 (semver) merged, so `feat/trace-harness-test` conflicts with master on at least:

- **`src/eval/intrinsics.rs`** — master has `TRACE_ENTRY` defined as 3-arg+unit (from #791). PR #794 correctly changes it to 4-arg+body. This is the right fix, but it needs to be applied as a rebase conflict resolution: take PR #794's version.
- **`src/eval/stg/debug.rs`** — master has the original `TraceEntry`/`TraceExit` from #791. PR #794 has corrected versions. Same resolution: take PR #794's version.
- **`src/eval/stg/mod.rs`** — master already registers `TraceEntry`/`TraceExit`. After rebasing, drop the duplicate additions from this PR.
- **`Cargo.lock`** — shows `0.7.0 → 0.7.1`; master is now at `0.8.0`. Will resolve automatically on rebase.

No CI has run due to the conflict. CI must be green before merge.

---

### Issue 2 — `017_namespacing.eu` regression

**File:** `src/core/metadata.rs`

The new `"trace"` case in `normalise_metadata()` intercepts the `:trace` symbol before the target-name fallback. This breaks `tests/harness/017_namespacing.eu`, which uses `` ` :trace `` as an output **target name** on the `env:` declaration (line 13):

```eu
` :trace
env: refs.env
```

With this PR, that declaration gets `{ trace: :lazy }` metadata instead of `{ target: :trace }` — the test silently produces wrong output or times out.

**Fix:** Rename the target in `tests/harness/017_namespacing.eu` from `:trace` to something unambiguous (e.g. `:result` or `:env-result`). The `:trace` symbol is now a reserved metadata key; it cannot double as a target name.

---

### What is correct and should be kept

- **4-arg `TRACE_ENTRY`** (name, args_list, strict, body → body): correct. The merged PR #791 had 3-arg+unit; this PR's version is right.
- **Wrapping order in desugar**: exit-wrap first, entry-wrap second giving `TRACE_ENTRY(..., TRACE_EXIT(..., body, ...))`. Correct.
- **`resolve_in_closure` in list traversal**: fixes the `navigate_local` panic on non-`Ref::L` refs. Correct.
- **Harness tests 159–162** and the `run_trace_test` helper: well-designed; `stderr.contains` matching is appropriate.
- **`metadata.rs` and `rowan_ast.rs` changes**: this PR effectively supersedes PR #790 (Quill's desugar PR). If this PR merges, PR #790 can be closed.

---

**Steps to unblock:**

1. Rebase onto master; resolve conflicts taking PR #794's versions of `intrinsics.rs` and `debug.rs`.
2. Remove the duplicate `rt.add(TraceEntry/TraceExit)` from `mod.rs` (already present on master).
3. Rename the `:trace` target in `tests/harness/017_namespacing.eu`.
4. Push; wait for CI green.

gmorpheme:
## aarch64 sequential harness failing — test files need assertion operators

The trace output is correct in every test (good!):

```
→ add(x: <thunk>, y: <thunk>)
159_trace_lazy...FAIL
→ add(x: 1, y: 2)
160_trace_strict...FAIL
→ add(x: <thunk>, y: <thunk>)
← add: <thunk>
161_trace_exit...FAIL
→ add(x: 1, y: 2)
← add: 3
162_trace_strict_exit...FAIL
```

The FAIL comes from the harness runner (`eu test --allow-io tests/harness`), which picks up all `.eu` files in `tests/harness/`. Tests 159–162 only contain `result: add(1, 2)` — no `//=` assertions and no `RESULT: :PASS` — so the harness considers them failures.

**Fix:** Change each of the four trace test files so they produce a passing harness result AND still exercise the trace mechanism. The simplest compatible structure:

```eu
` :trace
add(x, y): x + y

RESULT: add(1, 2) //= 3
```

The `//=` assertion evaluates `add(1, 2)` (triggering trace output to stderr), compares the result to `3`, and produces `:PASS` if equal. The `run_trace_test` helper in `harness_test.rs` runs the file normally (not via `eu test`) and checks that stderr contains the expected patterns — this still works because `//=` doesn't suppress trace output.

Apply the same pattern to all four files (adjusting the expected assertion as appropriate — for 161/162 which are `:exit`/`:strict-exit` mode, `add(1, 2) //= 3` is correct in all cases since trace doesn't alter the return value).

No other changes needed — the intrinsic implementation, wrapping order, and harness_test.rs runner are all correct.

PR #802, opened 2026-06-12
Cloudflare v5: WAF / rulesets module (+ DNS v5 migration)

gmorpheme:
## Summary

Cloudflare **Security & WAF** — a `cloudflare_ruleset` module with a
wirefilter **expression builder** — plus the DNS module's migration to
provider **v5** (a prerequisite, since one Terraform config uses one
provider version).

Targets the GA v5 provider (verified latest **5.19.1**, ~100% API
coverage). In v5 a ruleset's `rules` is a **list of objects**, which
maps directly onto a Eucalypt list — so a compact rule model expands
straight into a ruleset, and expressions are *composed* from helpers
rather than hand-written.

## Commits

1. **Migrate DNS to v5** — `cloudflare_record`→`cloudflare_dns_record`,
   `value`→`content`, zone data source filtered by `filter { name }`.
2. **WAF / rulesets module** — `lib/tf-cloudflare-waf.eu` + example.

## What's here

| File | |
|------|---|
| `lib/tf-cloudflare-waf.eu` | `cloudflare_ruleset` constructors; rule builders (`block` w/ custom response, `managed-challenge`, `js-challenge`, `log`, `skip`, `rate-limit`); zone entry-points (`custom-firewall`, `rate-limiting`); and a wirefilter expression builder. |
| `examples/terraform/cloudflare-waf.eu` | a security baseline fanned across zones. |

## The expression builder (the showcase)

Expressions need embedded `"quotes"` and `{set}` braces, which collide
with Eucalypt interpolation — the helpers hide that (c-strings for
quotes, brace-building for sets). So this:

```eu
waf.all-of[
  waf.path-prefix("/admin"),
  waf.negate(waf.ip-in-cidrs(office-cidrs)),
]
```

composes into the wirefilter string:

```
(starts_with(http.request.uri.path, "/admin")) and (not (ip.src in {203.0.113.0/24 198.51.100.0/24}))
```

## Testing

- Three module test suites green (`eu test lib/tf.eu`,
  `lib/tf-cloudflare.eu`, `lib/tf-cloudflare-waf.eu`).
- Both demos render to independently valid JSON; the WAF demo produces 4
  rulesets across 2 zones with composed expressions, a custom block
  response, and a rate-limit block.

## Caveats

- **`terraform validate` still not run in CI** (no `terraform` binary /
  network here) — the JSON is built to the official v5 schema, pulled
  from the provider docs.
- v5's `cloudflare_zone`/`cloudflare_dns_record` are known to have some
  plan-diff churn upstream; an operational note, not a generation issue.

Next per the plan: the **dev platform** (Workers/KV/R2/D1).

https://claude.ai/code/session_01Q9v9fZWsqEMPZMqtYRdgBB

### DIFF
--- a/examples/terraform/cloudflare-waf.eu
+++ b/examples/terraform/cloudflare-waf.eu
@@ -1,0 +1,72 @@
+{ doc: "Cloudflare edge-security baseline generated from a compact
+model, targeting provider v5.
+
+A few WAF custom rules and a rate-limit rule expand into
+`cloudflare_ruleset` resources across every zone.  Rule expressions are
+*composed* from wirefilter helpers (`all-of`, `path-prefix`, `negate`,
+`country-in`, ...) rather than hand-written strings.
+
+Render:
+
+    eu -j -L lib examples/terraform/cloudflare-waf.eu > security.tf.json
+    terraform init && terraform validate"
+
+  import: ["tf=tf.eu", "cf=tf-cloudflare.eu", "waf=tf-cloudflare-waf.eu"] }
+
+# ---- Compact model, applied across every zone ----------------------
+
+domains: {
+  primary:   "example.com"
+  secondary: "example.net"
+}
+
+office-cidrs:      ["203.0.113.0/24", "198.51.100.0/24"]
+blocked-countries: ["KP", "SY"]
+
+# WAF custom rules — expressions composed from helpers, not hand-written.
+firewall-rules: [
+  waf.block{
+    ref: "admin_office_only"
+    description: "Admin area only reachable from the office network"
+    expression: waf.all-of[
+      waf.path-prefix("/admin"),
+      waf.negate(waf.ip-in-cidrs(office-cidrs)),
+    ]
+  },
+  waf.managed-challenge{
+    ref: "challenge_low_score_bots"
+    description: "Challenge low-score bots"
+    expression: "cf.bot_management.score lt 30"
+  },
+  waf.block{
+    ref: "block_high_risk_geo"
+    description: "Block high-risk countries with a custom response"
+    expression: waf.country-in(blocked-countries)
+    response: { status_code: 403, content_type: "text/plain", content: "Access denied." }
+  },
+]
+
+rate-rules: [
+  waf.rate-limit{
+    ref: "throttle_api"
+    description: "Throttle API traffic per IP"
+    expression: waf.path-prefix("/api/")
+    characteristics: ["ip.src", "cf.colo.id"]
+    requests: 100
+    period: 60
+  },
+]
+
+# ---- Expansion: per zone, the handle + its security rulesets --------
+
+zone-security([tag, domain]): {
+  z: cf.zone(tag str.of, domain)
+  nodes: [z, waf.custom-firewall(z, firewall-rules), waf.rate-limiting(z, rate-rules)]
+}.nodes
+
+all-nodes: domains elements mapcat(zone-security)
+
+# ---- Document ------------------------------------------------------
+
+` :main
+main: tf.document(cf.provider-config("~> 5") ‖ all-nodes)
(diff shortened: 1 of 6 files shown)


examples/terraform/cloudflare-waf.eu
```diff
@@ -0,0 +1,72 @@
...
+    expression: waf.path-prefix("/api/")
+    characteristics: ["ip.src", "cf.colo.id"]
+    requests: 100
+    period: 60
+  },
+]
+
+# ---- Expansion: per zone, the handle + its security rulesets --------
+
+zone-security([tag, domain]): {
```
gmorpheme: Unnecessary parens

lib/tf-cloudflare-waf.eu
```diff
@@ -0,0 +1,178 @@
...
+
+` { doc: "`ip-in-cidrs(cidrs)` - source IP in an inline set of CIDRs." }
+ip-in-cidrs(cidrs): { s: cidrs wf-set }."ip.src in {s}"
+
+` { doc: "`country-in(codes)` - visitor country in a set of ISO codes." }
+country-in(codes): { s: codes map(quoted) wf-set }."ip.geoip.country in {s}"
+
+# Rules ---------------------------------------------------------------
+#
+# A rule is a block { ref, description, expression, action, ... }.  `o`
```
gmorpheme, line 60: Sounds like these should be expressed using types.

lib/tf-cloudflare-waf.eu
```diff
@@ -0,0 +1,178 @@
...
+
+` { doc: "`skip(o)` - skip rules / phases / products; `o.skip` is the
+`action_parameters` block, e.g. a `phases` list or `ruleset: current`." }
+skip(o): rule("skip", o) << { action_parameters: o.skip }
+
+` { doc: "`rate-limit(o)` - a rate-limiting rule.  Keys: `ref`,
+`expression`, `characteristics`, `requests`, optional `period`
+(default 60), `timeout` (default 600) and `action` (default block)." }
+rate-limit(o): rule(o lookup-or(:action, "block"), o) << {
+  ratelimit: {
```
gmorpheme: Why don't we merge onto a block of defaults instead of all the lookups?

lib/tf-cloudflare-waf.eu
```diff
@@ -0,0 +1,178 @@
...
+    characteristics: o.characteristics
+    period: o lookup-or(:period, 60)
+    requests_per_period: o.requests
+    mitigation_timeout: o lookup-or(:timeout, 600)
+  }
+}
+
+# Rulesets ------------------------------------------------------------
+
+` { doc: "`ruleset(local, scope, opts)` - a `cloudflare_ruleset`.  `scope`
```
gmorpheme, line 102: Why not use types to express this.

lib/tf-cloudflare-waf.eu
gmorpheme: Agreed. Deferring to a coherent **library-wide** type pass — `Node` / `Document` / `Rule` / `Expr` (plus literal-union actions/phases) across `tf.eu` + cloudflare + waf — once **0.8.0** lands. Typing this module in isolation would leave the node/document outputs loosely typed, so it's better done as one consistent layer.

---
_Generated by [Claude Code](https://claude.ai/code)_

lib/tf-cloudflare-waf.eu
gmorpheme: Deferring to the library-wide type pass planned for **0.8.0** (`Node` / `Document` / `Rule` / `Expr` across `tf.eu` + cloudflare + waf, with literal-union actions/phases), rather than typing the ruleset opts in isolation here.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #809, opened 2026-06-12
docs: literal string types need a c-string annotation (and argument unions need parens)

gmorpheme:
## What

The type-checking guide documents literal **string** types with a plain
annotation string:

```eu
` { type: "\"circle\" | \"rect\" -> bool" }   # documented, but does not parse
```

`\"` is a **c-string** escape — a plain `"..."` eucalypt string does not
process it, so the backslashes reach the type-annotation parser, which
errors with `unexpected character '\'`. The example as written cannot be
used.

This corrects the docs to the working form across the guide, the agent
reference, the cheat sheet, and the syntax-gotchas traps list:

```eu
` { type: c"(\"circle\" | \"rect\") -> bool" }   # c-string + parens
```

Two fixes, since the original examples have two defects:

1. **C-string for literal string types** — so the `\"` escapes survive to
   the type parser. (Literal *symbol* types like `:ok` contain no quotes
   and need no c-string.)
2. **Parenthesise a union in argument position** — `->` binds tighter
   than `|`, so `"circle" | "rect" -> bool` parses as the overload
   `"circle" | ("rect" -> bool)` and rejects even a valid `"circle"`
   argument. The one-argument reading is `("circle" | "rect") -> bool`.
   (The unparenthesised form is intentional and correct for *overloaded*
   function types — that distinction is now spelled out.)

## Files

- `docs/guide/type-checking.md` — the discriminated-union alias example,
  the "Writing literal types" section + examples, and the summary table.
- `docs/reference/agent-reference.md` — the type-forms table note and the
  escaping paragraph.
- `docs/appendices/cheat-sheet.md` — the type-forms table note.
- `docs/appendices/syntax-gotchas.md` — a new trap entry alongside the
  existing record-brace escaping section.

## Notes

- Docs only; no code changes. All edited fences stay `eu,notest`, so
  `scripts/test-doc-examples.py` skips them; the corrected forms were
  separately verified to `eu check` clean (valid literals accepted, typos
  rejected) on the 0.8.0 build.
- Found while doing the eucalypt-grove terraform type pass, whose WAF
  module relies on exactly this (c-string literal-union `Action` / `Phase`
  types).

### Possible follow-up (not in this PR)

The friendlier fix would be to make the *documented* plain-string form
work — i.e. process `\"` when extracting a `type:` annotation, or have the
type-annotation lexer tolerate a leading/again backslash-quote. Happy to
open a separate issue / change for that if preferred over requiring
c-strings.

https://claude.ai/code/session_016REnLgkxCxX3CmJefCYQEw

---
_Generated by [Claude Code](https://claude.ai/code/session_016REnLgkxCxX3CmJefCYQEw)_

### DIFF
--- a/docs/appendices/syntax-gotchas.md
+++ b/docs/appendices/syntax-gotchas.md
@@ -519,38 +519,18 @@ are fine as-is:
 ` { type-def: "Point" }                  # fine — just a name, no braces
 ```
 
-### Literal String Types Need a C-String, and Argument Unions Need Parens
+### Literal Types in Annotations
 
-A **literal string type** (`"block"`) embeds a `"` in the annotation, but a
-plain `"..."` eucalypt string does not process the `\"` escape — the
-backslash reaches the type parser and it errors (`unexpected character
-'\'`). Use a **c-string**, which does process `\"`:
+Two things to remember when a `type:` annotation contains literal types:
 
-```eu,notest
-# WRONG — \" is not an escape in a plain string; the type parser sees the \
-` { type: "\"block\" | \"log\"" }
-
-# RIGHT — a c-string yields literal quotes for the type parser
-` { type: c"\"block\" | \"log\"" }
-```
-
-Literal *symbol* types (`:block`) contain no quotes, so a plain string is
-fine — but a union in **argument position** still needs parentheses,
-because `->` binds tighter than `|`:
-
-```eu,notest
-# WRONG — parses as the overload :block | (:log -> bool), so even a valid
-# :block argument is rejected
-` { type: ":block | :log -> bool" }
-
-# RIGHT — parenthesise the argument union
-` { type: "(:block | :log) -> bool" }
-```
-
-The unparenthesised `A | B -> C` form is intentional for *overloaded*
-function types (`number -> number | string -> string`); it just is not
-what you want for a union-typed argument. Naming the union in a `types:`
-alias avoids the trap entirely, since the alias name is a single token.
+- **Literal string types need c-string escapes for the quotes** —
+  `c"\"block\" | \"log\""`. A plain `"..."` string does not process `\"`.
+  (Literal *symbol* types like `:block` have no quotes, so a plain string
+  is fine.)
+- **Mind `|`/`->` precedence** — `->` binds tighter, so a disjunction in
+  argument position needs brackets: `(:block | :log) -> bool`, not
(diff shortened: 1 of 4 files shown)


docs/guide/type-checking.md
```diff
@@ -910,7 +934,7 @@ but is not written to disk.
 | IO action                | `IO(T)`                                          |
 | Lens                     | `Lens(a, b)`                                     |
 | Traversal                | `Traversal(a, b)`                                |
-| Literal string type      | `"value"` (`\"value\"` in annotation string)     |
+| Literal string type      | `"value"` — c-string annotation: `c"\"value\""`  |
```
gmorpheme: I think the previous version was correct - it's saying you need the quotes. In a c-string they can go in by escapes. in a raw string you'd need `{dq}` or something which isn't processed at metadata time. There is no way in which you *have* to use c-strings here as soon we will have s-strings so I think this change is wrong. 

Let's just say "use c-string escapes"

docs/guide/type-checking.md
```diff
@@ -120,29 +120,53 @@ Literal types are most useful in **discriminated unions** — type aliases
...
-`"value"` (double-quoted) for literal strings. Since this appears inside
-a eucalypt string, the inner quotes must be escaped with `\"`:
+`"value"` (double-quoted) for literal strings. The `"` must survive
+inside the eucalypt annotation string, so the annotation must be a
+**c-string** (`c"..."`), whose `\"` escape yields a literal quote. A plain
+`"..."` string does *not* process `\"`, so the backslashes reach the type
+parser and it errors. Literal *symbol* types (`:name`) contain no quotes,
+so they need neither escaping nor a c-string.
+
+A second trap: in the type language `->` binds **tighter** than `|`
```
gmorpheme: How is this a "trap"!? - it is the same as every other type language - I don't know why you'd expect anything else. At most this is a gotcha that says "be careful about precedence of | and ->.

docs/appendices/syntax-gotchas.md
```diff
@@ -519,6 +519,39 @@ are fine as-is:
 ` { type-def: "Point" }                  # fine — just a name, no braces
 ```
 
+### Literal String Types Need a C-String, and Argument Unions Need Parens
```
gmorpheme: All a bit overstated. 

a) you need c-strings for typing if you're going to have literal strings (for " escapes)
b) be careful around precedence of -> and | - you may need to put disjunctions in brackets

PR #826, opened 2026-06-14
fix(prelude): activate blob path end-to-end (W6 blob-activate)

gmorpheme:
## Summary

- **xtask**: Replace `collect_lambda_forms` (expected exactly 1 form per binding, failing for `__build`'s 15-form LetRec) with wrapping each compiled `Rc<StgSyn>` directly as `LambdaForm::Thunk`. Inner Let/LetRec allocates its local bindings when forced; slot is updated to the resulting WHNF.
- **blob format**: Old `bindings: Vec<ArenaLambdaForm>` only stored 296 entry-level forms; `ArenaStgSyn::LetRec` nodes referenced inner forms (e.g. `FormIdx(1420)`) that were missing → index-out-of-bounds panic. New `forms_pool: Vec<ArenaLambdaForm>` stores ALL forms (5652 total); `binding_entries: Vec<FormIdx>` holds one index per prelude binding into the pool.
- **double-load eliminated**: blob loaded once in `bin/eu.rs`, stored on `SourceLoader`, extracted via `take_prelude_blob()` before `Executor` construction. `eval.rs` no longer calls `maybe_load_prelude_blob` in `try_execute`.

## Verification

- `cargo xtask prelude-compile` succeeds: 296 bindings, 5652 forms in pool, 6801 nodes, blob 130 KB
- `cargo build` activates `cfg(prelude_blob_ok)` — no stale-blob warning
- 435/436 harness tests pass (`test_error_043` is a pre-existing failure on master, passes on master without blob)
- `EU_GC_VERIFY=2 EU_GC_STRESS=1` on prelude and list tests: clean

## Test plan

- [ ] `cargo xtask prelude-compile` completes without error
- [ ] `eu -e '1 + 1'` returns 2
- [ ] `eu -e '[1,2,3] map(str.of)'` returns `["1","2","3"]`
- [ ] `EU_GC_VERIFY=2 EU_GC_STRESS=1 eu tests/harness/010_prelude.eu` passes
- [ ] All 435 harness tests pass (excluding pre-existing `test_error_043`)


gmorpheme:
Fixed the test_error_043 regression (pushed to branch).

**Root cause**: On the blob path, `cook_with_prelude` seeds `=` as the prelude equality operator. `result = 42` is cooked as `App(Free("="), [Free("result"), 42])`. The old note was keyed on `FreeVar("=")`; now `FreeVar("result")` fires instead.

**Fix**: Added `CoreError::AssignmentStyleDeclaration` detected in the **verify pass** (after cook, before STG compile). Detects `App(Var::Free("="), [Var::Free(name), ...])` on both paths. The new diagnostic:

```
error: assignment-style syntax: 'result = ...' is not a valid eucalypt declaration
  = eucalypt uses ':' for declarations, not '='; write 'result: <value>' instead of 'result = <value>'
```

This is cleaner than the old approach: fires before STG compilation, identical message on both paths, and the error points at `=` (not at `result`).

**All 436 harness tests pass.**

PR #865, opened 2026-06-19
fix(import): convert YAML complex-key panic to proper error

gmorpheme:
## Summary

- `Expectation::feed()` in `src/import/yaml.rs` had two `panic!("bad key type")` calls reached when a YAML file uses a non-scalar mapping key (sequence or mapping as a key, e.g. `? [1, 2]: value` syntax).
- These now return `Err(SourceError::InvalidBlockKey(...))` instead of panicking.
- `feed()` signature changed from `fn feed(self, expr: RcExpr) -> Expectation` to `fn feed(self, expr: RcExpr, file_id: usize) -> Result<Expectation, SourceError>`.
- `Receiver::feed()` threads `file_id` through and stores any error in the existing `Receiver::error` field for deferred surfacing via `Receiver::result()`.
- Harness test 172 verifies the clean error message for a complex-key YAML file.

Note: scalar non-string YAML keys (integers, booleans, null) are handled earlier by `convert_key_scalar()` which converts them to symbols — only complex keys (sequences/mappings used as keys) reach `Expectation::feed()`.

## Test plan

- [x] `cargo test --test harness_test test_error_172` passes
- [x] `cargo test --lib` passes (1137 tests)


gmorpheme:
@claude rebase this

PR #872, opened 2026-06-20
feat(gc): sticky-mark-bit minor collection and write barrier (P3+P4)

gmorpheme:
## Summary

Implements P3 (sticky-mark-bit minor collection) and P4 (write barrier with eager promotion) from the GC nursery spec.

- **No-flip visited-set approach**: Major collections use a `HashSet` visited set for traversal termination instead of mark-state flips. This resolves the fundamental 1-bit mark limitation where eager promotion would cause promoted objects to be skipped after a flip.
- **Minor collections**: Use `is_marked()` for generational skipping — old objects are skipped, young objects are traced and tenured in place. No line reset, no sweep, no flip.
- **Write barrier**: When an old env frame is updated (thunk update), the closure's code and environment pointers are immediately marked if young, eliminating old→young edges at mutation time.
- **Simplified collector**: `CollectorHeapView` uses a single `visited: Option<HashSet>` field — `Some` for major, `None` for minor. The `is_new()` helper unifies termination logic. Removed `mark_minor_inner()` and duplicate minor-mode branches.

## Test plan

- [x] All 1146 lib tests pass
- [x] All 450 harness tests pass
- [x] `EU_GC_VERIFY=2 EU_GC_STRESS=1 EU_GC_POISON=1` — all harness tests pass
- [x] `cargo fmt --all` clean
- [x] Test 089 (sharing/thunk memoisation) passes under stress

### DIFF
--- a/src/eval/machine/vm.rs
+++ b/src/eval/machine/vm.rs
@@ -558,31 +558,29 @@ impl MachineState {
 
     /// Update environment index to point to current closure.
     ///
-    /// This is the single mutation site for the generational nursery.
-    /// The write barrier is currently disabled — see comment in body.
+    /// Includes an eager-promotion write barrier: when a marked (old)
+    /// env frame is updated to point to an unmarked (young) closure,
+    /// the closure's code and environment are immediately marked
+    /// (tenured).  This eliminates old→young edges at mutation time,
+    /// allowing minor collections to skip old objects entirely.
+    ///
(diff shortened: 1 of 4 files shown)


gmorpheme:
Closing — superseded by PR #875 which is being reworked with the correct approach (HashSet minor, unchanged major). This PR diverged too far from master to be usable.

PR #882, opened 2026-06-21
feat(eval): unify render pipeline with IoTransparentBranch (eu-ogvr)

gmorpheme:
## Summary

- Eliminates double evaluation by always compiling with `RenderDoc` instead of headless + re-render
- Adds `IoTransparentBranch` continuation that yields IO constructors to the driver while passing non-IO data through to normal rendering
- Adds `StgSyn::IoTransparentCase` / `HeapSyn::IoTransparentCase` that push the new continuation
- `RenderDoc::wrapper()` uses `io_transparent_force` for the top-level case
- Simplifies driver post-run dispatch (IO yield → io_run, normal → done)

## Test plan

- [x] All 1162 lib tests pass
- [x] All 451 harness tests pass (including IO and error tests)
- [x] `cargo fmt --all` clean
- [ ] CI green on all platforms

### DIFF
--- a/docs/superpowers/specs/2026-06-21-render-unification.md
+++ b/docs/superpowers/specs/2026-06-21-render-unification.md
@@ -1,247 +1,163 @@
-# Spec: Render Unification — Eliminate Double Evaluation
+# Spec: Render Unification with IoTransparentBranch
 
 **Bead**: eu-ogvr  
 **Author**: Quill  
 **Date**: 2026-06-21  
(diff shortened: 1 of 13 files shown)


gmorpheme:
Closing — render unification approach doesn't unlock AtMostOnce for rendered blocks as hoped (RenderKv enters values twice within a single pass). The architectural cleanup is real but the complexity tradeoff doesn't justify it. See eu-ogvr and follow-on bead for future rationalisation.

PR #944, opened 2026-07-02
feat(bytecode): BV1 — migrate ParseString + TypeToData to neutral data construction (eu-9p9g)

gmorpheme:
## Summary

Part of eu-vi3a (BV1 full bytecode VM). Migrates `parse_string::ParseString`, `typedata::TypeToData`, and `typedata::TypeFromString` off the `Compiler::compile` + `memory::loader::load` runtime-code-synthesis path onto the neutral data-construction ABI (`data_value`/`native_value`/`return_closure_list`/`meta_value`), per the plan's arena-growth analysis: their output is a pure data literal (block/list/scalar tree), so it can be built directly on the GC heap with zero code-arena growth, instead of compiling to STG and loading it as `HeapSyn` code (which the bytecode engine cannot execute).

- New shared module `src/eval/stg/data_literal.rs`: `build_value(machine, view, &RcExpr) -> Result<AbiClosure, ExecutionError>` recursively materialises a data-literal Core expression as a value. Handles `Literal`/`List`/`Block`, `Let(_, _, LetType::DefaultBlockLet)` (the general block representation produced by the import readers — resolved via proper de Bruijn indexing, not name-based substitution, avoiding a shadowing pitfall), and `Meta` (EDN tagged elements / YAML custom tags).
- Confirmed `typedata::type_to_rcexpr` emits only data constructors (`Literal`/`List`/`Block`), never function application, so the "fixed template" fallback the plan flagged as a possible design fork was not needed.
- New `IntrinsicMachine::meta_value` primitive (`intrinsic.rs`), mirroring `data_value`, for `HeapSyn::Meta{meta,body}` — HeapSyn-only default for now (nothing in the corpus exercises it on bytecode yet).
- Two new differential tests in `differential.rs`: `agree_on_parse_string_json`, `agree_on_type_to_data`.

## Results

- **Corpus: 144 → 160 PASS** under the `EU_BYTECODE=1` differential sweep (19 → 8 residual DIFFs), verified stable across two independent sweep runs and under `EU_GC_VERIFY=2 EU_GC_STRESS=1` on both engines for all `parse-as`/typedata harness tests.
- Residual 8 DIFFs confirmed **out of scope / pre-existing**: 4 MergeWith/DEEPMERGE, 1 ProducerNext/streams, 1 unmigrated `block.rs` inspector, and 2 that hit a newly-surfaced (previously masked by `TypeFromString`'s panic) assertion-diagnostic-rendering gap shared with `125_expectations.eu` — confirmed via a standalone repro that the underlying assertion (`match?` on a union spec) fails identically on both engines; only the failure *rendering* differs, unrelated to data construction.
- `cargo test --lib`: 1250 green (1247 baseline + 2 new differential tests + 1 pre-existing).
- `cargo clippy --all-targets -- -D warnings`: clean. `cargo fmt --all`: clean.

Ledger updated in `docs/superpowers/plans/2026-07-01-bv1-bytecode-vm.md` (dated eu-9p9g entry + RESUME header updates).

## Test plan
- [x] `cargo test --lib` (1250 passed)
- [x] `cargo clippy --all-targets -- -D warnings` (clean)
- [x] `cargo fmt --all` (clean)
- [x] Full `tests/harness/*.eu` differential sweep (EU_BYTECODE=1 vs HeapSyn): 160 PASS / 8 DIFF
- [x] `EU_GC_VERIFY=2 EU_GC_STRESS=1` on both engines for parse-as/typedata harness tests

Refs eu-vi3a, eu-9p9g

https://claude.ai/code/session_01QphBzunfXuSs4Mv8PMBsee

### DIFF
--- a/docs/superpowers/plans/2026-07-01-bv1-bytecode-vm.md
+++ b/docs/superpowers/plans/2026-07-01-bv1-bytecode-vm.md
@@ -10,7 +10,9 @@
 > `~/.cargo/bin/cargo` symlink is broken). One feature branch per increment →
 > PR to `integration/0.12.0`. Owner reviews/merges PRs; do not merge your own.
 > Current working branch: `feat/bv1-migrate-intrinsics` (off the post-#941 merge
-> `25166517`).
+> `25166517`). **eu-9p9g increment (2026-07-02, branch
+> `feat/bv1-parse-typedata-data2`, off the #942 merge `06c6ee19`): ParseString +
+> TypeToData migrated — see dated ledger entry below.**
 >
 > **State:** **The bytecode engine works end-to-end.** Real `.eu` programs run
 > under `EU_BYTECODE=1`, **byte-identical to the HeapSyn engine** (blocks, lists,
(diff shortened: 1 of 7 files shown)


gmorpheme:
Superseded by #945. A second furnace (on Opus — the owner's required model for BV1) independently produced an equivalent eu-9p9g migration reaching the same 160 PASS / 8 DIFF corpus result, additionally handling Meta on the bytecode path (bounded meta_template) with harness_test 477 green. Closing this one to avoid a duplicate merge; #945 is going through gatekeeper review. Thanks for the solid work here — the two implementations agreed on the outcome, which is a good cross-check.

PR #978, opened 2026-07-09
perf(bytecode): emit Op::Seq for provably-safe uniform-branch Case (eu-9mvh, lever c)

gmorpheme:
## Summary

W1 of 0.12.1, owner-approved lever (c): teach the bytecode encoder to detect **uniform-branch `Case` nodes** (all alternatives — branches and fallback — resolve to the same continuation body, provably without reading the value `Case` would bind) and emit the existing cheap `Op::Seq` instead of the branch-table-decoding, heap-allocating `Op::Case`. Encoder-only — `StgSyn` is unchanged (confirmed via `eu dump stg` diff), so HeapSyn is untouched.

**Headline finding, reported honestly per the dispatch: this lever, in the only form I could prove sound, does not move the regression set.** The profiled `<=`/`-`/`+` operand-forcing `Case`-of-`Case` chains (the W0 spike's target) always read the forced value — that's the entire point of forcing an operand — so they never satisfy a rule that must decline whenever the shared body reads the bound value.

## What shipped vs. what I tried and reverted

- **Shipped**: `uniform_seq_body` + `drop_outer_bindings`, an occurs-check-and-shift: if every branch/fallback shares an identical body, and that body has *no* free reference to the value `Case` would bind (checked by walking the body tracking nested-binder depth, mirroring `AllocationPruner`'s shift bookkeeping in `src/eval/stg/optimiser.rs` without touching or depending on it), the case is downgraded to `Op::Seq`. A fallback is required (otherwise `Case`'s tag-exhaustiveness check has no `Seq` equivalent). Additionally requires the shared body to contain **no nested `Case`** — see below.
- **Tried and reverted**: an extension aliasing `force(local(j), body)` (zero explicit branches, scrutinee a bare local ref) back onto `local(j)` itself, on the theory that forcing a local ref updates it in place so `Case`'s fresh binding is redundant. That argument is correct *in isolation*, but the index arithmetic for anything below a **nested** `Case` inside such a body assumed the nested `Case` would keep ordinary `Op::Case` (frame-introducing) semantics — while `encode_node` tries this same downgrade on every `Case` node independently, including ones inside an already-rewritten body. When a nested `Case` also qualified, it was encoded as `Op::Seq` too and contributed no frame at runtime, silently invalidating indices the outer rewrite had computed on the opposite assumption. **This was a real, reproduced bug**: `{x: 42}."{x}"` returned a type error instead of `"42"` (caught by the existing `cargo test` suite — 6 failures, all string-interpolation/dotted-lookup and two `differential` tests). I root-caused it, then dropped the aliasing extension rather than try to patch around it further, and added a `contains_case` guard so the same composition hazard can't recur for the strict path either (it's latent there too, just not exercised by the current corpus). Full detail is in the module-level comment above `uniform_seq_body` in `src/eval/bytecode/encode.rs`.

## Verification

- `eu dump stg` identical before/after on `tests/harness/bench/001_naive_fib.eu` — STG genuinely unchanged.
- `cargo fmt --all`, `cargo clippy --all-targets -- -D warnings`, `cargo clippy --target wasm32-unknown-unknown --lib -- -D warnings` — all clean.
- `cargo test` (both default bytecode and `EU_HEAPSYN=1`) — 1272 lib + 479 harness tests green on both engines (byte-identical invariant).
- `EU_GC_VERIFY=2 cargo test` — green.
- Blob vs `--source-prelude` parity spot-checked (fib, the string-interpolation repro that once broke) — identical output; blob regenerated via `cargo xtask prelude-compile` (276486 → 276426 bytes, a 60-byte reduction from the handful of sites that do qualify, e.g. simple `force(local(j), STR(local(0)))`-shaped fallbacks in the render/string-interpolation path).

## Before/after benchmarks (fresh clean-build baseline, this machine)

`cargo clean && cargo build --release && cargo xtask prelude-compile && cargo build --release`; external wall median of 7 runs, `timeout 600 ./target/release/eu --heap-limit-mib 12288 <case>` (bc) / `EU_HEAPSYN=1 ...` (hs); aoc25 run from `examples/aoc25/`.

**Regression set** (target ≤1.25×, none >1.3×):

| Case | bc before | hs before | ratio before | bc after | hs after | ratio after |
|---|--:|--:|--:|--:|--:|--:|
| `001_naive_fib` (`-t bench-naive-fib`) | 3.9048s | 1.9749s | 1.977× | 3.8801s | 1.9803s | 1.959× |
| `005_drop_cons` (`-t bench-drop-cons`) | 0.0969s | 0.0753s | 1.286× | 0.1180s | 0.0624s | 1.889× |
| `007_short_lived` (`-t bench-short-lived`) | 0.1756s | 0.1287s | 1.365× | 0.1693s | 0.1084s | 1.562× |
| aoc25 `day03 -t part-2` | 1.5068s | 0.9845s | 1.530× | 1.4936s | 0.9347s | 1.598× |
| aoc25 `day09 -t part-1` | 5.3266s | 4.3902s | 1.213× | 5.2696s | 4.3222s | 1.219× |

**Protected wins** (must not regress):

| Case | ratio before | ratio after |
|---|--:|--:|
| `range(0,5000) count` (higher-order) | 0.872× | 0.843× |
| aoc25 `day07 -t part-2` | 0.838× | 0.840× |
| aoc25 `day11 -t part-1` | 0.962× | 0.983× |
| `004_generations` (`-t bench-generations`) | 1.023× | 0.868× |

**Reading**: `001_naive_fib` and `day09-p1` (the two longest, most stable runs, several seconds each) show **no meaningful change** — well within run-to-run noise (fib: -0.9%; day09-p1: +0.5%). `005`/`007` (sub-200ms) show large swings in *both* directions on the completely-unchanged HeapSyn baseline too (005 hs: -17%; this can only be machine noise on a shared host at these timescales, consistent with the W0 spike's own observation that "005 and 007 show the most run-to-run spread"). None of the protected wins regressed beyond noise.

Per the dispatch's own risk framing: **"if the lever alone doesn't reach the bar on fib, STOP and report — do not loosen the bar."** I'm doing that. Lever (c), in the only form provable sound within the "no new opcode" constraint, does not close any part of the gap — the dominant `Case` traffic the W0 spike measured is exactly the shape that must be declined by any sound version of this rule. Superinstruction fusion (lever a) or a real environment-shifting mechanism (a new opcode, out of this task's scope) would be needed to attack the actual hot pattern.

## Test plan
- [x] `eu dump stg` unchanged (encoder-only)
- [x] `cargo fmt --all` clean
- [x] `cargo clippy --target wasm32-unknown-unknown --lib -- -D warnings` clean
- [x] `cargo test` green, both engines
- [x] `EU_GC_VERIFY=2 cargo test` green
- [x] blob / `--source-prelude` parity spot-checked
- [x] Fresh clean-build before/after benchmarks captured and reported above (not fabricated — real `w1_bench.sh` output)

https://claude.ai/code/session_01KRUGGB4wXRfywnDArH2suk

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ All notable changes to eucalypt are documented here.
 ### Changed
 
 - **Bytecode-vs-HeapSyn engine gap closed** — entries added as the gap-close work (superinstructions/decode fusion, ExecutionError boxing, block index) lands.
+- **Uniform-branch `Case` compiles to `Op::Seq` where provably safe** — the bytecode encoder now detects `Case` nodes whose branches and fallback all share an identical body that provably never reads the value `Case` would bind (i.e. the case exists purely to force the scrutinee to WHNF and discards the tag *and* the value), and emits the existing cheap `Op::Seq` instead of the branch-table-decoding, heap-allocating `Op::Case`. Encoder-only; STG is unchanged and HeapSyn is unaffected. **Measured impact on the regression set (fib/005/007/day03-p2/day09-p1) is nil**: the profiled `<=`/`-`/`+` operand-forcing `Case`-of-`Case` pattern always reads the forced value (that is the entire point of forcing it), so it never qualifies under a rule provable sound without also renumbering references through nested `Case`s — an extension attempted during development, found unsound (a reproduced, since-reverted bug), and not reintroduced. This lever alone does not close the gap; a further lever (e.g. superinstruction fusion) is needed (eu-9mvh).
 
 ## [0.12.0] - 2026-07-04
 
--- a/src/eval/bytecode/encode.rs
+++ b/src/eval/bytecode/encode.rs
@@ -152,6 +152,33 @@ impl Encoder {
                 branches,
                 fallback,
             } => {
+                // Uniform-branch downgrade (eu-9mvh, W1, lever (c)): when
+                // every branch and the fallback share an *identical*
+                // continuation body that provably never reads the value
+                // `Case` would bind (see `uniform_seq_body` for exactly
+                // what "provably" requires), the case exists purely to
+                // force the scrutinee to WHNF and discards both the tag
+                // and the value — so emit the existing cheap `Op::Seq`
+                // (force-then-continue, no branch-table decode, no
+                // per-Case `Array` allocation) instead of `Op::Case`. Note
+                // this does *not* fire for the profiled `<=`/`-`/`+`
+                // operand-forcing `Case`-of-`Case` chains themselves —
+                // their whole point is to read the forced value — see the
+                // module-level note above `uniform_seq_body` for the full
+                // account, including a soundness pitfall this ran into
+                // during development. This is bytecode-encoder-only: the
+                // shared `StgSyn::Case` node is untouched, so HeapSyn
+                // (which walks the original tree) sees no change at all.
+                if let Some(seq_body) = uniform_seq_body(branches, fallback) {
+                    let scr_off = self.encode_node(scrutinee);
+                    let body_off = self.encode_node(&seq_body);
+                    let offset = self.here();
+                    self.emit_op(Op::Seq);
+                    self.emit_u32(scr_off);
+                    self.emit_u32(body_off);
+                    return offset;
+                }
+
                 // Children first (post-order).
                 let scr_off = self.encode_node(scrutinee);
                 let branch_data: Vec<(u8, CodeRef)> = branches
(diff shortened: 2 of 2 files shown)


gmorpheme:
Closing unmerged per owner decision (2026-07-10). The lever-(c) transform is **sound but inert on the target sites**: `Op::Seq` forces to WHNF without binding the value, so the occurs-check correctly declines on the primop `<=`/`-`/`+` force-dispatch chains — which exist precisely to read the forced value. Net zero perf benefit (fib 1.977→1.959, noise). Not merging 378 lines of correct-but-dead encoder machinery (YAGNI). Branch `perf/uniform-case-seq` and the W0 report are preserved for reference. Gap-close proceeds via lever (a): a fused force-dispatch-primop superinstruction (a new opcode that extracts the value Op::Seq cannot). Excellent work reproducing + dropping the unsound aliasing extension rather than shipping it.

PR #1001, opened 2026-07-13
feat(sv): prefix-list type [A, B, C…] (eu-2sa6.3)

gmorpheme:
## SV — prefix-list type (eu-2sa6.3)

Implements the owner-signed design `docs/superpowers/specs/2026-07-13-sv-prefix-list-design.md` in full: a new type-DSL form `[A, B, C…]` — a fixed-shape prefix followed by a homogeneous variable tail — that types the canonical hiccup/markup element shape `[tag, attrs, …content]` = `[Symbol, Block, (String | Element)…]`.

All three owner decisions are honoured: **Q1** both `…` (U+2026, canonical) and ASCII `...` are accepted and lex to one token; **Q2** out-of-prefix indexed projection returns `t?`; **Q3** `[C…]` normalises to `List(C)` at construction, so there is exactly one internal representation.

### What's implemented (by design section)

**Phase 1 — surface & representation (§2, §3)**
- Lexer `Token::Ellipsis` (both spellings; `...` longest-match before the record `..` tail).
- Parser prefix-list bracket form; a comma-in-brackets type without a trailing ellipsis stays an error (fixed tuples remain `(A, B)`).
- `Type::PrefixList { prefix, tail }` with the normalising `Type::prefix_list` smart constructor; `Display`, `kind_of`, `widen_literals`, `humanise`, `unfold_mu`, and the `unify` / `resolve_typedata` / checker-alias recursion arms.

**Phase 2 — subtyping & projection (§4, §5)**
- `Tuple <: PrefixList` (the load-bearing rule — literals synthesise to tuples), prefix widening + tail covariance, `PrefixList <: List` homogenise, `List </: PrefixList` (non-empty prefix), gradual consistency, and the `[C…] ≡ List(C)` structural guard.
- `head`/`second`/`tail` project the fixed prefix precisely, including through a `Mu`-wrapped recursive alias (the `Element` shape). Out-of-prefix named projection → `t?`.

**Phase 3 — flow-through (§6)**
- `to-data`/`from-data` (`:t-prefix-list`), `as-spec`/`match?` (`prefix-list-spec`), and JSON-Schema export (2020-12 `prefixItems` + open `items`).

### Tests
- Unit tests across `parse.rs`, `types.rs`, `subtype.rs`, `check.rs`, `render.rs`.
- Harness `typecheck/109–111` (clean check, precise-projection warning, `t?` out-of-prefix) and `187_sv_prefix_list` (to-data/from-data/as-spec/match? end-to-end).
- Full suite green: 1295 lib + 495 harness, `cargo clippy --all-targets -D warnings` and `cargo fmt` clean.

### Notes for review
- **§4.5 meet/join is a no-op here** — there is no type-level join/meet function in the checker (the design's line reference points at the consistency dispatcher); §9.4 already lists exact meet/join as a non-goal, so nothing was wired.
- **§8.3 headline / annotating `lib/markup.eu`** — harness test `109` replicates the acceptance with **explicit-param** accessors (`tag(e): e head`, …), which meaningfully route through the projection-precision logic (a wrong annotation warns). I did **not** annotate the shipping `lib/markup.eu` itself: its accessors are **point-free** (`tag: head`), and a bare polymorphic accessor checks *vacuously* against any `Element → …` annotation (it passes even with a deliberately wrong result type, because projection precision fires on an *application*, not a bare value). So annotating markup.eu would type-check clean but verify nothing about prefix-lists, while modifying a shipping library — I judged that not worth it. Happy to add the annotations for documentation value if you'd prefer; flagging the vacuous-point-free behaviour as a design/reality gap worth noting.
- **Pre-existing `union-spec` bug (out of scope):** `as-spec`/`match?` on a **union** tail (e.g. the markup `(String | Element)…`) does not match, because `union-spec` in `lib/prelude.eu` is byte-identical-broken on `master` (`ts map(to-spec-td) any(v match?)` applies the value as the pattern). `prefix-list-spec` correctly delegates tail-matching to `to-spec-td`, so it works for any tail whose own spec works and inherits this pre-existing limitation for unions. Not fixed here — it belongs to the SV2 `to-spec` work. Worth a separate bead.


gmorpheme:
## Review: excellent implementation, two small fixes needed before merge

Thorough review done in an isolated worktree against the owner-signed design (`docs/superpowers/specs/2026-07-13-sv-prefix-list-design.md`). Summary of verification performed:

**Gates (all independently re-run, not trusted from the PR description):**
- `cargo test --release`: 1299 lib + 495 harness passed (matches/exceeds the claimed 1295+495).
- `EU_HEAPSYN=1 cargo test --release --test harness_test`: 495/495 — no engine divergence.
- CI: all checks green at time of review except `Test Suite (windows-latest)` still running.

**Design conformance** — verified section-by-section against the spec, not just trusted the PR description:
- §2/§3 (surface + representation): lexer `Token::Ellipsis`, longest-match `...` before `..`, `Type::PrefixList`, the `Type::prefix_list` normalising constructor. Confirmed live: `[C…]` really does construct `List(C)`, not a degenerate `PrefixList`.
- §4/§5 (subtyping + projection): all rules present and unit-tested — `Tuple <: PrefixList` (the load-bearing rule), prefix widening, `List </: PrefixList`, gradual consistency. Independently confirmed `head`/`second`/`tail` project precisely through a `Mu`-wrapped recursive alias, and that out-of-prefix named projection synthesises as `T?` (verified live with `eu check --strict` — test 111's own assertion, "expected string, found string?", reproduces exactly).
- §6 (flow-through): `to-data`/`from-data`/`as-spec`/`match?`/JSON-Schema all present and exercised by `187_sv_prefix_list.eu`, which I ran directly.

**Author's three review notes — independently verified, not taken on trust:**
1. §4.5 meet/join no-op: confirmed there is genuinely no meet/join function anywhere in the checker (`grep` across `subtype.rs`/`unify.rs`/`check.rs` finds none) — matches design §9.4's own non-goal. Correct, nothing to wire.
2. `lib/markup.eu` left unannotated: reproduced the vacuous-checking claim live — annotating the point-free `tag: head` with a deliberately wrong return type produces *no* warning, while the explicit-param `tag2(e): e head` with the same wrong annotation correctly warns. The decision not to annotate the shipping file is sound and well-reasoned.
3. `union-spec` pre-existing bug: reproduced live on a clean build of `origin/master` (pre-this-PR) — `"hello" match?(s"string | number" as-spec)` already returns `false` on master, byte-identical to this branch. Genuinely pre-existing, correctly out of scope.

**One incidental finding, not a blocker:** the new Mu-unfolding-before-Tuple-check (`check.rs` — added to make `Element`'s recursive alias project precisely) also affects the *existing* Tuple projection path for any other Mu-wrapped recursive alias, not just prefix-lists. I constructed a minimal recursive Tuple type (`Rec: "(number, Rec)"`) and confirmed this does change behaviour: a previously-silent wrong-annotation case now warns correctly (a false-negative fix, not a new false positive — a correctly-annotated recursive-tuple case stays clean on both branches). Zero in-tree blast radius (no shipped `.eu` file uses a recursive Tuple alias). Worth a one-line comment in the code since the design doc frames this as PrefixList-specific but it's actually a small Tuple-path improvement too — not required before merge, just flagging for the record.

**Two things that do need fixing before merge:**

1. **`tests/harness/187_sv_prefix_list.eu:2`** uses `{ import: "resource:reflect" }`. Every sibling SV harness test (`174`–`183`) was just switched to the idiomatic `{ import: "reflect.eu" }` in #1000 (eu-1n5z), which merged to master *before* this branch was created (confirmed: `0527468f` is an ancestor of this branch's first commit) — so this isn't stale-branch drift, it's a fresh inconsistency with a convention this project settled literally hours ago. Please change to `{ import: "reflect.eu" }`.

2. **`docs/guide/type-checking.md`'s new "Prefix-lists" section** — the worked example uses `tag = head`, `attrs = second`, `content = _ tail tail` (`=` instead of `:`). I ran this exact snippet and it does not parse — `error: empty expression where a value was expected` / `missing ':' after declaration head`, the *identical* bug class that #999 just fixed in `lib/markup.eu` itself. This is inherited from the design spec's own §8.3 pseudocode (which has the same typo, but that's an internal design note, not a user-facing guide chapter). Please fix the three declarations to use `:` in the shipped guide page — a reader copying this example verbatim hits a parse error.

Both are mechanical, one-line-per-instance fixes. Everything else — the type-theory, the subtyping rules, the projection precision, the flow-through, the test coverage — is excellent and matches the owner-signed design faithfully. Happy to merge as soon as these two are fixed; no re-review of the substantive implementation needed, just confirm the two corrections and I'll merge on green CI.

PR #1004, opened 2026-07-13
feat(bytecode): branch-table side-pool + BranchPredecoded + ticks-first ledger (eu-2sa6.13)

gmorpheme:
## Branch-table side-pool + ticks-first ledger (eu-2sa6.13 follow-up)

Follow-up to #1002. Two parts, both agreed with the coordinator:

### 1. Branch-table side-pool + `BranchPredecoded` continuation (design §1.2)

Removes the drop_cons residual. Under the pre-decoded engine a `Case` no longer
rebuilds a fresh GC-heap `Array<Option<CodeRef>>` on **every** dispatch. The
densified branch table already lives once in the off-heap, immutable
`DecodedProgram::branches` pool; the new `BcContinuation::BranchPredecoded`
variant references its run by `(branch_start, branch_len)` **index** (not a
borrowed slice, so it survives any pool growth) plus `min_tag` and a
`NO_BRANCH`-sentinel `fallback`.

**GC surface (the sensitive bit):** `BranchPredecoded` scans/updates **only its
`environment`** — strictly less than `Branch`, which also marks and forwards its
`Array` backing. The pool indices, the fallback ordinal and `min_tag` are inert
scalars, never scanned. Documented in `cont.rs`. The byte path keeps `Branch`
unchanged.

`return_data`/`return_fun`/`return_native` each gain a `BranchPredecoded` arm
resolving the branch and expected-tags from the pool (`branch_pool_match` /
`branch_pool_expected`); a native only ever takes the fallback, so its arm needs
no pool access at all.

Output/alloc **identical** — the Case branch `Array` was never counted in
`state.allocs`, so this is purely a wall / GC-pressure win.

### 2. Ticks-first ledger measurement (wall deferred)

Per the ticks-first directive, the deterministic (load-independent) data is
captured now; wall median-of-N runs are deferred to a quiet window (the box was
under build/review load). Full detail:
`docs/superpowers/reports/2026-07-13-predecode-ticks-first.md`.

**Honest headline:** on the canonical suite the pre-decoded engine is
**step-count-neutral** vs the byte engine (−0.0…−0.1% bc dispatch steps across
015/018/019/020/022). This is correct and expected — operand pre-decode reduces
*per-step CPU cost* (typed field reads vs byte re-decode), **not** the number of
steps, so the win is a wall effect and is the **deferred** measurement. BV2
Ann-elimination removes steps only where hot-path `Op::Ann` nodes exist, which
the compute benches have few of (fib's −13.6% step drop is workload-specific,
not representative). No wall number is quoted here and no `results.jsonl` row is
appended until the quiet-window medians exist — the deterministic `hs_ticks`/
`hs_allocs`/`gc` for all eight benches are recorded in the report and will drop
into those rows verbatim.

### Gate matrix (host Darwin-25.5.0-arm64, rustc 1.97.0, fresh blob)

| Gate | Result |
|---|---|
| Byte path `cargo test` — full suite | **green** (1299 lib + 495 harness), incl. tick-parity tripwire (1.12 cap) |
| `EU_PREDECODE=1 cargo test` — full suite | **green** (1299 lib + 495 harness), incl. tripwire (1.13 predecode cap) + differential vs HeapSyn |
| `EU_PREDECODE=1 EU_GC_VERIFY=2 EU_GC_POISON=1 EU_GC_STRESS=1` harness | **green** (495/495) — validates the new continuation's GC surface |
| clippy `--all-targets` | clean |
| Byte-identical output vs byte engine | yes (whole suite) |

Wall/median-of-N ratios: **deferred** to a quiet window (coordinator scheduling).

Do not merge — Wicket reviews, owner re-reviews pre-release.

### DIFF
--- a/docs/superpowers/reports/2026-07-13-predecode-ticks-first.md
+++ b/docs/superpowers/reports/2026-07-13-predecode-ticks-first.md
@@ -8,41 +8,58 @@
 - **Protocol:** `docs/superpowers/engine-ab/PROTOCOL.md`
 - **Status:** TICKS-FIRST only. Wall median-of-N runs are **deferred** to a quiet
   window (the measurement box was under concurrent build/review load); no wall
-  ratio is quoted here, per PROTOCOL §2 ("if the machine is loaded, say so").
-  The `results.jsonl` rows will be appended once the wall medians exist, using
-  the deterministic figures below unchanged.
+  ratio is quoted here, per PROTOCOL §2. The `results.jsonl` rows will be
+  appended once the wall medians exist, using the deterministic figures below
+  unchanged.
 
-## 1. Deterministic reference (HeapSyn `-S`) — the load-independent layer
+> **Correction (2026-07-13, post-review).** The first version of this report
+> measured with `eu run <file>` and **no `-t <target>`** (and, for `021`, no
+> `--allow-io`). That renders the *whole document*, which for some benches
+> evaluates extra work the bench target does not: `020_lookup_curve` defines
+> `total40`/`total250` as top-level bindings that the `bench-lookup-curve`
+> target *also* references, so the whole-document render evaluated them twice
+> (Wicket's "2× + 682" signature — thank you); `015` rendered extra top-level
+> blocks (+35 M ticks); and `021_io_loop` ran with no `--allow-io`, so its io
+> loop never executed (1,518 ticks instead of 3.97 M). Every figure below is
+> now measured the way the canonical suite / `cargo xtask engine-ab` measures
+> (`base_cmd`, `xtask/src/engine_ab.rs`): `EU_HEAPSYN=1 eu -S --heap-limit-mib
+> 12288 [--allow-io] -t <target> <file>`, parsing the `-S` `Ticks`/`Allocs`/
+> `Collections` lines. Rows corrected: **015, 020, 021** (others moved by
+> ≤~1,200 ticks — the small whole-document render overhead). The step-neutral
+> conclusion in §2 is unchanged.
 
-Read from `EU_HEAPSYN=1 eu -S`, `--heap-limit-mib 12288` (0 collections, matching
-production). These are the `hs_ticks`/`hs_allocs`/`gc` columns of the eventual
-ledger rows. **Confidence: measured-verified** (deterministic).
+## 1. Deterministic reference (HeapSyn `-S -t <target>`) — the load-independent layer
 
-| Bench | Class | hs_ticks | hs_allocs | gc |
-|---|---|---|---|---|
-| 015_block_merge | D | 134,179,383 | 645,066 | 0 |
-| 016_import_export_yaml | I | 59,895,722 | 1,922,943 | 0 |
-| 017_import_export_toml | I | 59,722,384 | 1,286,727 | 0 |
-| 018_string_scale | G | 76,815,495 | 364,229 | 0 |
-| 019_list_scale | H | 55,803,922 | 156,222 | 0 |
-| 020_lookup_curve | E | 3,442,892 | 11,107,135 | 0 |
-| 021_io_loop | L | 1,518 | 552 | 0 |
-| 022_hof_fold | C | 52,225,882 | 180,203 | 0 |
+`--heap-limit-mib 12288` (0 collections, matching production). These are the
+`hs_ticks`/`hs_allocs`/`gc` columns of the eventual ledger rows. **Confidence:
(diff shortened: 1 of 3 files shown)


gmorpheme:
## Review: code change is solid; one report figure needs re-verification before merge

Thorough review done in an isolated worktree, rebuilding from a fresh blob at the PR's exact head (02f32794).

### Code — branch-table side-pool + `BranchPredecoded` (verified correct)

- **GC surface**: read `GcScannable for BcContinuation` in `cont.rs` directly. Both `scan` and `scan_and_update` for `BranchPredecoded` touch only `environment` — `branch_start`/`branch_len`/`fallback`/`min_tag` are never marked, forwarded, or otherwise treated as GC-relevant. This is strictly less surface than `Branch`, which additionally marks/forwards its `Array` backing, exactly as claimed.
- **Pool validity**: confirmed by grep that `.branches.push(...)` (and the other three pools) only ever happens inside `predecode.rs`'s `Decoder`, invoked once at `BytecodeMachine::new()` before any dispatch — there is no runtime growth path under the current eager-only decode (per #1002's signed delta). `branch_pool_match`/`branch_pool_expected` resolve through the live `&DecodedProgram` reference at *use* time on every call, not a cached slice, so the design is correct regardless of whether a future lazy-decode phase reintroduces pool growth.
- **Behavioural equivalence**: `branch_pool_match`'s tag/min_tag/NO_BRANCH logic and `branch_pool_expected`'s error-list construction are structurally identical to the untouched `Branch` arm's `branch_table.get(...).flatten()`/filter-map logic — same order, same values, so `NoBranchForDataTag`/`CannotReturnFunToCase` error messages are unaffected.
- **Byte path**: confirmed zero-diff — the only removed lines are the old Array-building code at the single `Op::Case` call site inside `handle_op_predecoded` (the predecode-only dispatch), plus the `return_fun` signature widening needed to thread `decoded` through for the new arm's error path. The original `Branch` variant, its byte-path construction in `handle_op`, and its `GcScannable` impl are byte-for-byte unchanged.
- Gates independently re-run: fmt/clippy clean; byte path 1299 lib + 495 harness green including the 1.12 tripwire; `EU_PREDECODE=1` 1299 lib + 495 harness green including the 1.13 tripwire; `EU_PREDECODE=1 EU_GC_VERIFY=2 EU_GC_POISON=1 EU_GC_STRESS=1 --test harness_test`: 495/495 green.

### Report — one figure does not reproduce

Per the ticks-first-only discipline (confirmed: no `results.jsonl` changes in this PR, no wall numbers quoted anywhere), I reproduced the deterministic tick counts myself on 015/018/019/020/022 (not just trusting the table). Four of five matched almost exactly (within a few hundred ticks out of tens of millions — normal build-state noise). **`020_lookup_curve` does not reproduce**:

| | Report (§1/§2) | My reproduction |
|---|--:|--:|
| hs_ticks | 3,442,892 | **1,721,105** |
| byte steps | 329,261,216 | **164,630,267** |
| predecode steps | 329,180,052 | **164,589,685** |

Both report figures are almost exactly 2× my reproduced values — but not *exactly* 2×: the excess over a clean double is **682 ticks in both cases** (329,261,216 − 2×164,630,267 = 682; 3,442,892 − 2×1,721,105 = 682, the identical constant). That's a strong signature of the bench having been run twice and the two runs' results summed (with the usual few-hundred-tick session variance between the two runs), rather than a single measurement. I re-verified my own number three independent ways (`--statistics-file`, `-S` via both `eu` and `eu run`, and `eu test`) and after a clean rebuild — all agree exactly at 164,630,267 for byte. The bench file itself is untouched by this PR and deterministic (fixed 130-round loop, no external sizing), so this isn't a case of the workload changing between measurements.

The good news: the *qualitative* claim — step-count-neutral, ~0.0% delta — still holds with the corrected numbers (164,589,685 vs 164,630,267 is −0.02%, not meaningfully different from the reported −0.0%). This isn't a finding that changes the report's conclusion. But a "measured-verified" (i.e., deterministic, should reproduce byte-for-byte) figure that's off by a clean factor of ~2 in a report whose entire purpose is measurement rigor needs correcting before it's cited anywhere else — please re-measure and fix the `020_lookup_curve` row in both §1 and §2 of `docs/superpowers/reports/2026-07-13-predecode-ticks-first.md`.

### Verdict

Code change: approved, no concerns. Report: one row needs a quick re-measurement and correction. Once that's fixed (and CI is green), I'd expect to merge without needing to re-review the code.

PR #1008, opened 2026-07-13
docs(design): criterion-based copy-specialisation of recursive higher-order combinators — eu-yuto (owner sign-off)

gmorpheme:
## Design note for owner sign-off — production recursive-combinator peeling (eu-yuto)

`docs/superpowers/specs/2026-07-13-recursive-combinator-peeling-design.md`. **Design only — no implementation.** For the owner's personal review (template: the predecoded-IR design, PR #989).

The eu-v8n8 spike (PR #1007) proved the mechanism — inlining the recursive higher-order combinators (`map`/`foldl`) collapses their O(N²) lazy-thunk chains to O(N), Wicket-confirmed via N=5k→40k scaling (flag-off →4×/doubling, flag-on exactly 2.0000×). But it's a **blob-only** `inline_cores` hook, so it overshoots source and special-cases the prelude.

This note specifies the **production** form: a semantic criterion in **`tag_combinators`** (shared core pipeline) — peel a self-recursive lambda when it (1) applies one of its own parameters, (2) is within a size cap, (3) has free vars resolving in the inline set/globals. Because `tag_combinators` runs on **both** the source merge-cook inline pass and the blob-gen fixed-point, blob + source + user combinators + alternative preludes all fuse identically (owner no-special-casing principle).

**Covers every required section:**
- **The criterion** — three legs, each justified from evidence (fib first-order null control; Wicket's O(N²)→O(N) scaling; the +bloat measurement); size cap + unroll-depth as named constants; margins (mutual recursion, per-branch application, nested-`aux` recursion) stated explicitly.
- **Qualifying-set survey** — the 8 directly-qualifying combinators (`map`/`foldl`/`foldr`/`scanl`/`scanr`/`map2`/`drop-while`/`iterate`, checked not guessed) + the transitive tail (`filter`/`zip`/`all`/`any`/…), with **measured** full-set bloat (**+1505 B**, bytecode unchanged) and canonical-suite tick effect (extends the win to 016/017/020 beyond `map`/`foldl`, **no regressions**). Argues for admitting the whole criterion-set, not a curated list.
- **Both-paths mechanics** — how it lands on source vs blob (incl. the two-line blob-gen self-reference extension); tripwire implications (closes the ~1.67× map/foldl divergence the spike introduced; the fib tripwire is unaffected — that stays eu-npp9's job).
- **Demand-analysis interaction + metadata override** (recommend against for v1, with the plug-in point).
- **Relationships** — eu-v8n8 (recommend **close #1007**, port test 188), eu-npp9 (retains global-form fusion only), BV3/CG4 (O(n²) fold motivation removed → re-baseline trigger), wall/ledger (PROTOCOL run post-landing).
- **Risks** — code size, compile-time (measured negligible), typechecker/LSP, termination — each with a gate.

All numbers are deterministic layers only (busy machine — no wall claims), confidence-labelled per PROTOCOL.

Do not merge without owner sign-off on the design decisions.


gmorpheme:
## Wicket review pass — §0.4 settled, pending cells filled, one correction

All measurements below on the already-adjudicated hash-verified worktrees at
commit `f8686d0a` (unchanged from the mechanism-adjudication phase):
`/tmp/eu-adjudicate-flagoff` (binary `b2bf8a67…`, blob `e59640e6…`) and
`/tmp/eu-adjudicate-flagon` (binary `af644ce8…`, blob `d6cd1555…`). All bench
runs use `-t <target>` (and `-I` for 021) to match the harness's actual target
selection — I found the default (no `-t`) CLI render can silently double-count
or short-circuit certain targets (metadata-`requires-io` targets evaluate to
`{}` without `--allow-io`/`-t`), so every cell below is `-t`-explicit.

### 1. §0.4 is no longer open — settle it

Ran the existing (previously unused) `EU_ENV_DEPTH_HISTOGRAM=1` diagnostic
(`src/eval/machine/env.rs:36-134`, eu-qm7f) against the verbatim un-fused 022
shape at N=5k/10k/20k on the flag-off binary:

- **Lookup depth is flat**: mean 0.979/0.989/0.995, **max exactly 2** at every
  N. No lookup gets deeper as N grows — this is unchanged from eu-qm7f's
  original shallow-walk finding.
- **Lookup count is quadratic**: 13,157,734 / 51,315,234 / 202,630,234 total
  `EnvironmentFrame::get` calls, tracking `N(N-1)/2` (12,497,500 / 49,995,000 /
  199,990,000) to within 0.7-3%, converging as N grows — the classic
  triangular-number signature of "position k does ~k lookups."
- **Lookup count correlates ~1:1 with ticks** at every N (ratios 0.967 / 0.983
  / 0.991, tightening toward 1).

**Settled account**: it *is* environment resolution driving the O(N²)
(contradicting the Furnace Exp-4 note's "NOT env re-resolution"), but the
mechanism is quadratically-growing lookup *count*, not lookup *depth*
(contradicting the literal §6.4 "re-resolving through an env-walk" phrasing,
which implies depth growth — already ruled out once by eu-qm7f and reconfirmed
here). **Corrected BV3/CG4 implication**: the fix must eliminate the
per-position *repeated* lookup count, not shorten the walk distance — walk
distance was never the bottleneck. Please rewrite §0.4 as settled (drop the
"interpretation OPEN" framing and the two-bullet "on the table" list), and
correct §5.3's "precise reading... still open" accordingly.

### 2. Canonical suite, flag-on vs flag-off — filled

| Bench | flag-off ticks | flag-on ticks | Δ | allocs off/on |
|---|---|---|---|---|
| 015_block_merge | 98,669,543 | 96,056,695 | −2.65% | 448,718 / 456,248 |
| 016_import_export_yaml | 59,711,048 | 56,042,568 | −6.14% | 1,794,817 / 1,814,846 |
| 017_import_export_toml | 59,537,724 | 55,869,244 | −6.16% | 1,158,667 / 1,178,696 |
| 018_string_scale | 76,730,983 | 46,020,189 | −40.02% | 364,176 / 395,650 |
| 019_list_scale | 55,719,473 | 51,160,960 | −8.18% | 156,182 / 160,676 |
| **020_lookup_curve** | **164,630,267** | **164,465,805** | **−0.10%** | 5,553,215 / 5,595,923 |
| 021_io_loop | 3,958,278 | 3,958,278 | 0% (io-bound) | 499,537 / 499,537 |
| 022_hof_fold | 52,125,436 | 2,130,398 | −95.91% | 180,163 / 217,643 |

7 of 8 rows (015-019, 021, 022) reproduce the doc's carried-over "pending
re-verification" figures to within 0.1-0.3% — **adjudicated, confirmed as-is**.

**020 is wrong in the doc and needs correcting, not just tagging adjudicated.**
The doc cites baseline 1,721,105 / full-set 1,556,643 (−9.6%). I get
164,630,267 / 164,465,805 (−0.10%) on the bytecode engine (the engine every
other row uses). I tracked down *why*: `EU_HEAPSYN=1` on the identical flag-off
binary/file gives **exactly** 1,721,105 ticks (allocs 5,553,215 — identical to
my bytecode allocs figure, as expected since alloc count is roughly
engine-independent here). **The doc's 020 baseline was measured on HeapSyn, not
bytecode** — a one-row engine mix-up, not a re-verification gap. The corrected
number changes the story: 020 does **not** meaningfully benefit from map/foldl
fusion (its `foldl` call is a minor fraction of a lookup-dominated workload) —
please replace the row and reword §2.3's closing paragraph (lines 250-254),
which currently uses the 020 "−9.6%" figure to argue fusion "extends... to the
lookup (020) workloads via filter/foldr/all/any transitively." That specific
claim doesn't hold for 020 as measured; the 016/017 import-workload claim does
hold (confirmed above) and can stand.

### 3. §2.2 bloat — filled for what exists, full-8-set flagged as not measurable

Ran `./target/debug/xtask prelude-compile` in both worktrees (verified
byte-identical blob SHA256 before/after in both cases, so this is a safe
re-derivation, not a rebuild-drift risk):

| Metric | flag-off (map+foldl absent) | flag-on (map+foldl only) | Δ |
|---|---|---|---|
| blob file size | 461,638 B | 461,979 B | **+341 B** |
| inline_cores count | 133 | 135 | +2 |
| arena nodes | 7,104 | 7,104 | **0** |
| STG lambda forms | 352 | 352 | 0 |
| bytecode size | 269,176 B | 269,176 B | **0** |

This **mechanistically confirms** "bytecode image unchanged" generalizes to
any inline_cores membership change (bytecode compiles from the prelude's own
STG forms unconditionally; inline_cores membership only affects what raw Core
gets additionally serialized for injection — structurally independent of
*which* combinators are in the set). So the doc's qualitative bytecode-unchanged
claim is solid for the full-8-set too, even though I can't produce the full-8
number itself (the spike's `CLUSTER` list is hardcoded to `["map", "foldl"]`,
`xtask/src/main.rs:334` — the other 6 combinators are simply not implemented in
any branch I can build). Per instruction, I have **not** synthesised a full-8
figure. Recommend the doc's §2.2/§6.1 tag stay as **"requires a prototype"**
rather than "pending re-verification" for the full-8 number specifically — the
map+foldl-only number above is genuinely re-verified now; the full-8 one is
categorically different (needs new code, not a rerun). The +341 B (2
combinators) vs. the original draft's +1,505 B (8 combinators) claim is at
least *consistent in shape* (~170 B/combinator vs. ~188 B/combinator) — worth
noting as a plausibility check, not proof.

### 4. Exp 2 (depth) / Exp 3 (demand) / Exp 4 (quadratic) — reconfirmed, one methodology note

Re-ran with `-t bench-scale` for full rigor (all binaries identical hashes to
the earlier adjudication — `depth1` `417b858a…`, `depth0` `b818e1e6…`):

- **Depth-1**: 1,025,304 / 2,050,304 / 4,100,304 (N=5k/10k/20k) — ratios
  2.0000×/1.9999×. **Depth-0**: 13,597,816 / 52,195,316 / 204,390,316 — ratios
  3.838×/3.916×, closely tracking the true flag-off baseline. Both confirm the
  prior findings unchanged.
- **Demand dependence — methodology correction**: suppressing demand analysis
  on the **flag-off** (un-fused) binary changes *nothing* (ticks identical to
  1 part in a million) — expected, since there's no concrete operator for
  demand analysis to specialise without fusion first. The effect only shows on
  the **flag-on** binary: normal 1,065,279/2,130,279/4,260,279 (ratios
  exactly 2.0000×) vs. `--suppress-demand-analysis` 4,200,265/14,650,265/
  54,300,265 (ratios 3.49×/3.71×, quadratic-trending). Please make sure any
  cited demand-dependence figure in the doc is explicit that suppression was
  applied to the **fused** binary — the un-fused case is a null test.
- **Exp 4 (flag-off quadratic quantities)**, reconfirmed with `-t`: ticks
  13,562,816/52,125,316/204,250,316 (quadratic, ratios 3.84×/3.92×); allocs
  90,137/180,137/360,137 (linear, ratios 1.998×/1.999×); blocks
  1,046/2,087/4,168 (linear); max-stack 5,011/10,011/20,011 (linear). These
  supersede the doc's "Furnace cross-check, pending re-verification" row —
  promote to **Wicket adjudicated**, these are the authoritative figures (the
  doc's cross-check numbers were already very close, within ~0.1%).

### Summary of asks for Furnace's next revision

1. Rewrite §0.4 as settled per §1 above (not "interpretation open").
2. Replace the 020 row in §2.3 with 164,630,267 / 164,465,805 / −0.10%, tag
   **Wicket adjudicated**, and note the HeapSyn/bytecode mix-up as the root
   cause of the old figure.
3. Reword the §2.3 closing paragraph — drop the "lookup (020)" clause from the
   "extends the win" argument; the import (016/017) clause stands.
4. Promote 015/016/017/018/019/021/022 to **Wicket adjudicated** with the
   figures in §2 above.
5. Fill §2.2 with the map+foldl-only bloat cell (§3 above); keep the full-8-set
   figure explicitly tagged as **requires a prototype**, not pending
   re-verification, since it cannot be produced from existing code.
6. Replace Exp-4's row with the reconfirmed figures in §4; add the demand-
   suppression methodology note (fused binary, not un-fused) wherever that
   figure is cited.

No merge action taken — this is review-comment-only per the standing
instruction; #1008 goes to the owner once Furnace folds these in.

PR #1009, opened 2026-07-14
docs(perf): lever(a) byte-vs-predecode wall measurement, protocol-grade (eu-2sa6.13)

gmorpheme:
Completes the outstanding lever(a) PROTOCOL wall measurements deferred from `docs/superpowers/reports/2026-07-13-predecode-ticks-first.md` (ticks-first only at the time). Owner-granted quiet window. **Do not merge** -- Wicket reviews (directed mode), owner reviews pre-release.

## Setup (gold protocol, per `bd recall stale-binary-rebuild-trap`)

Fresh worktree off master (`5f550b49`, includes #1002 predecode + #1004 branch-pooling + #1005/#1006 diagnostics merged, both disabled by default). Blob regenerated, then `cargo build --release` re-run and confirmed to show `Compiling eucalypt` (proving the fresh blob was actually re-embedded, not a stale cached binary -- this is the exact trap that bit Furnace three times). SHA256 hashes of blob, binary, and all 8 bench files recorded in the report. One binary/one blob serves both engines throughout (predecode is `EU_PREDECODE=1` runtime-toggled), so there's no build-desync risk between arms.

## Item 2: standard ledger row

`cargo xtask engine-ab` appended 8 bc/hs rows to `results.jsonl` (blob config, this commit). `--check` compared correctly this time against the immediately preceding blob-config session (no `prelude_config` mismatch, unlike the eu-lhai incident) -- **no regressions >15%**; the two WATCH flags (`list_scale` class-H, `lookup_curve` class-E tripwire) are both informational per PROTOCOL, not failures. Also serves as the post-#1004 branch-pooling wall validation.

## Item 3: byte-vs-predecode wall (the main event)

Interleaved per bench, median-of-5 with spread, all 8 canonical benches:

| Bench | ratio (predecode/byte) |
|---|--:|
| block_merge | 0.932 |
| import_yaml | 0.959 |
| import_toml | 0.968 |
| string_scale | 0.929 |
| **list_scale** | **0.618** |
| lookup_curve | 0.836 |
| io_loop | 1.001 |
| hof_fold | 0.943 |

Predecode is faster on 7/8 benches (3-17% on six of them), wall-neutral on `io_loop` (expected -- shell-exec-bound, not compute-bound, so a dispatch CPU-cost win has nothing to act on), and shows a striking **38% win on list_scale**. That outlier was not just accepted: independently verified via `-S` (byte 1.782s / predecode 1.100s mutator time, matching the external wall timing), near-identical ticks (55,719,473 vs 55,707,469) and allocs (156,182 both) confirming it's a genuine per-step cost win rather than skipped/extra work, identical output, and reproduced three more times in isolation outside the main script. No hypothesis is offered for *why* list_scale benefits disproportionately -- flagged as a follow-on question, not answered here.

These figures live in the completed `2026-07-13-predecode-ticks-first.md` report, not `results.jsonl` -- that ledger's schema has no predecode column, and silently overloading `hs_wall_med` would corrupt every future `--check` comparison. Extending the schema is a separate, out-of-scope decision.

## Honesty notes

`mediaanalysisd`/`spotlightknowledged` (the explicitly disqualifying processes) were absent/suspended throughout, as granted. `WallpaperAerialsExtension` (a macOS idle/lock screensaver, ~15-25% CPU) was intermittently active during parts of the run -- not on the disqualifying list, but unexpected background load, noted rather than ignored. Within-bench spreads stayed tight regardless. Labelled **measured-single** throughout (one clean interleaved session; not yet independently repeated), consistent with that caveat -- except the list_scale ticks/allocs/output layer, which is measured-verified (deterministic, independently re-checked multiple ways).

Full report, all hashes, and confidence labels on every figure: `docs/superpowers/reports/2026-07-13-predecode-ticks-first.md`.


gmorpheme:
## Wicket review — independently verified, do not merge (owner reviews pre-release)

Per policy this is a Stopwatch PR, so I am reviewing only (directed mode) —
**not merging**, regardless of CI outcome. Owner takes it from here.

### Verification method

The report's cited binary/blob SHA256 (`b2bf8a67…`/`e59640e6…`) are **byte-
identical** to a pair I already have built and hash-recorded from an
unrelated adjudication (`/tmp/eu-adjudicate-flagoff`, commit `f8686d0a`,
which only differs from this PR's base `5f550b49` by an `xtask`-only commit —
so the `eu` binary and blob are expected, and confirmed, identical). All 8
bench-file hashes cited in §4 also matched exactly what's on disk there. This
let me independently re-run every claim in §2.5 on the *exact same*
binary/blob/bench-file triple cited in the report, rather than trusting the
numbers or rebuilding from scratch.

### What I checked

1. **Deterministic layer (019_list_scale)** — reran ticks/allocs myself:
   byte 55,719,473 ticks / 156,182 allocs; `EU_PREDECODE=1` 55,707,469 ticks /
   156,182 allocs. **Exact match** to the report's cited figures, bit-for-bit.
2. **Wall spot-check, the 38% outlier** — 3 fresh interleaved byte/predecode
   pairs on 019_list_scale: byte 1.77/1.77/1.81s, predecode 1.07/1.09/1.09s.
   Ratio ≈ 0.60-0.61, closely matching the reported 0.618. The outlier
   reproduces independently, on a different session, not just within
   Stopwatch's own reruns.
3. **Wall spot-check, a modest win** — 022_hof_fold: byte 1.267/1.272s,
   predecode 1.184/1.188s, ratio ≈ 0.934-0.935 vs. reported 0.943. Close.
4. **Wall spot-check, the neutral case** — 021_io_loop: byte 2.568s,
   predecode 2.556s, ratio ≈ 0.995 vs. reported 1.001, both essentially flat
   as expected for an IO-bound bench.
5. **results.jsonl cross-reference** — the new `020_lookup_curve` ledger row
   here correctly labels `hs_ticks: 1721105` as the **HeapSyn** figure. That
   number is directly relevant to a correction I just posted on #1008: that
   PR's design-note doc cited `1,721,105` as a *bytecode* baseline for the
   same bench without labelling the engine — this ledger row confirms
   `1,721,105` really is the HeapSyn count, corroborating that #1008's figure
   was a HeapSyn/bytecode mix-up, not a fabrication or typo. Nice incidental
   cross-check between the two reports.

### Assessment

Everything checked reproduces independently, closely, including the large
outlier that most needed scrutiny. The report's own internal rigor is good:
labelled confidence per figure (measured-verified vs measured-single),
honest disclosure of the `WallpaperAerialsExtension` background load even
though it wasn't on the disqualifying list, and three independent
cross-checks (mutator time, ticks/allocs, isolated reruns) on the 019 outlier
before reporting it rather than after a challenge.

**On the schema question you raised**: agreed, not appending predecode
figures into `results.jsonl`'s `hs_wall_med` column is the right call. That
field name is specifically HeapSyn; overloading it for a third engine would
silently corrupt every future `--check` comparison's meaning with no way to
tell which engine a historical row means. Keeping the byte-vs-predecode data
in the report and treating a schema extension as a separate, deliberate
decision is correct engineering judgement, not a scope gap.

No issues found. Recommend for owner sign-off; I am not merging.

PR #1010, opened 2026-07-14
feat(inline): copy-specialisation of recursive higher-order combinators (eu-dp0k)

gmorpheme:
## eu-dp0k — copy-specialisation of recursive higher-order combinators

Implements the owner-signed design (`docs/superpowers/specs/2026-07-13-recursive-combinator-copy-specialisation-design.md`, PR #1008) on **blob-path scope** (owner Option A). Source-path closure is the separate 0.13 release-blocker **eu-rb5n**.

### What landed

- **The §1 criterion in `tag_combinators`** (shared pipeline): a self-recursive lambda is tagged inlinable when it (1) applies one of its own parameters, (2) recurses on its own name, and (3) is within `MAX_COPY_BODY_NODES = 48`. de Bruijn depth tracking (`applies_own_param`), self-ref matching (`references_name`), `node_count`; threaded through `Meta` (docstrings) and `Let` arms; a name-passing `tag_combinators_named` entry.
- **The §3.2 blob-gen leg**: `all_free_vars_in_set_with_self` (the binding's own name resolves to its injected slot) + `tag_combinators_named` in xtask's fixed-point, plus an extension of the free-var traversal to `Let`/`Block`/`Lam` bodies so the **full qualifying set** is admitted (not 4/8). All 8 — `map`, `foldl`, `foldr`, `scanl`, `scanr`, `map2`, `drop-while`, `iterate` — are now in `inline_cores` (133 → 143).
- **Test 188** (v8n8 laziness preservation) ported from the closed spike branch; passes on all three engines. `170/171/179/181` stay green.

### Measurements (deterministic layers only — machine busy, no wall)

**Provenance.** baseline = `origin/master` clean worktree, binary `b2bf8a67…` / blob `e59640e6…` (461,638 B, 133 cores) — *exactly matches Wicket's adjudicated baseline*. full-8 = this branch @ `20eb69a4`, binary `1cea2b92…` / blob `ad09a865…` (463,305 B, 143 cores). Blob generation is deterministic (identical SHA across regens). bytecode engine, `-t <target>`, `--allow-io` for 021.

**Canonical suite ticks (blob/default config):**

| bench | baseline | full-8 | Δ |
|---|---|---|---|
| 015_block_merge | 98,669,543 | 96,009,128 | −2.70% |
| 016_import_export_yaml | 59,711,048 | 55,930,340 | −6.33% |
| 017_import_export_toml | 59,537,724 | 55,757,016 | −6.35% |
| 018_string_scale | 76,730,983 | 45,958,939 | −40.10% |
| 019_list_scale | 55,719,473 | 51,135,460 | −8.23% |
| 020_lookup_curve | 164,630,267 | 164,406,909 | −0.14% |
| 021_io_loop | 3,958,278 | 3,958,278 | 0.00% (io-bound) |
| 022_hof_fold | 52,125,436 | 2,085,398 | **−96.00%** |

No bench regresses; the full set slightly beats the map+foldl-only adjudicated figures on the transitive beneficiaries (016/017/019 via `filter`/`foldr`/`all`/`any`).

**Bloat (full-8 vs baseline):**

| metric | baseline | full-8 | Δ |
|---|---|---|---|
| inline_cores | 133 | 143 | +10 |
| blob size | 461,638 B | 463,305 B | **+1,667 B** |
| bytecode size | 269,176 B | 269,176 B | **0** |
| arena nodes | 7,104 | 7,104 | **0** |
| STG lambda forms | 352 | 352 | 0 |

The bytecode image is byte-for-byte unchanged — the growth is entirely source-side `inline_cores` payload (~167 B/core). (+10 = the 8 qualifying recursive combinators + 2 previously-rejected `Let`/`Block`-bodied combinators newly admissible via the broadened free-var traversal.)

### Gates

- Full **lib (1305)** + **harness (496)** suites green; `clippy --all-targets -D warnings` clean; `cargo fmt` clean.
- **GC verify/poison/stress**: 015/018/019/022 pass under `EU_GC_VERIFY=2 EU_GC_STRESS=1 EU_GC_POISON=1`.
- **fib tick-parity tripwire unmoved**: baseline == full-8 (bc 20,657,846; hs 26,378,436) — fib is first-order, does not qualify (null control).
- Blob determinism verified.

### Honest divergence disclosure (deferred gate)

The "both configs fuse" gate is **deferred to eu-rb5n by owner decision**. On this branch:
- **Blob (default) path fuses**: 022 = 2,085,398 ticks (bytecode).
- **Source-prelude path does NOT fuse**: `EU_SOURCE_PRELUDE=1` 022 = 51,885,361 (quadratic).

`EU_SOURCE_PRELUDE` is opt-in, so the **default build ships the win**. The design's §3.1 ("no source-path wiring needed") is **wrong** — see the in-PR design amendment (commit 20eb69a4). Two blockers, tracked in eu-rb5n: (1) `distribute` is Meta-blind; (2) demand-specialisation needs pre-expansion parity. A naive meta-peel is **type-unsafe** (strips type annotations — harness test 104 is the tripwire); do not attempt it without preserving annotation metadata.

### DIFF
--- a/docs/superpowers/specs/2026-07-13-recursive-combinator-copy-specialisation-design.md
+++ b/docs/superpowers/specs/2026-07-13-recursive-combinator-copy-specialisation-design.md
@@ -322,6 +322,40 @@ lambdas inlinable, and `inline_pass` distributes each into the user's call sites
 as a local specialised copy. No new source-path wiring is needed — the criterion
 change alone activates it.
 
+> **⚠ AMENDMENT (post-sign-off, eu-dp0k, 2026-07-14) — §3.1 above is WRONG.**
+> Implementation found the criterion alone does **not** fuse the source-prelude
+> path; `EU_SOURCE_PRELUDE=1` bench 022 stays quadratic (~52M ticks). Two
+> independent blockers, neither anticipated here:
+>
+> 1. **`distribute` is metadata-blind.** Prelude combinators are
+>    `Meta(Lam, docstring)`; `reduce::inlinable()` matches only bare
+>    `Lam(_, true, _)`, so documented combinators are never distributed on the
+>    source path. (The blob path works only because xtask `peel_meta`s the
+>    `inline_cores` before storing them.) This *is* the mechanism behind the
+>    §3.4 source-vs-blob divergence.
+> 2. **Demand-specialisation needs pre-expansion parity.** Even once the local
+>    copy exists, the source-path copy is not specialised strict (it stays
+>    quadratic) because it lacks the blob's pre-expanded `inline_cores`; the
+>    blob path gets a fully-specialised local `letrec`, the source path a
+>    partial one.
+>
+> **⚠ Type-annotation hazard (landmine for any fix).** A naive fix — peeling
+> `Meta` in `distribute` to make combinators inlinable — is *unsound*: type
+> annotations are also `Meta` nodes, so peeling strips them, and an annotated
+> identity like `` `{type:"number->number"} wrap(x): x`` inlined at
+> `wrap("hello")` silently loses its annotation and the type mismatch goes
+> undetected. Harness test **`104_suppress_type_warnings_ok.eu`** is the
+> tripwire; any source-path meta handling must preserve type/annotation
+> metadata while enabling combinator inlining.
+>
+> Source-path closure is tracked as the 0.13 release-blocker **eu-rb5n** (its
+> two sub-problems are the blockers above). eu-dp0k lands the **blob (default)
+> path only**; the "both configs fuse" gate is deferred to eu-rb5n by owner
+> decision. `EU_SOURCE_PRELUDE` is opt-in, so the default build is unaffected.
+> eu-rb5n also notes the eu-npp9 relationship: the Meta-blindness looks like the
+> same source-path mechanism behind the global-form divergence, so one PR may
+> close both.
+
 ### 3.2 Blob (generation) path
 
 On the blob path the prelude is pre-compiled; the combinator body reaches the
--- a/src/core/inline/tag.rs
+++ b/src/core/inline/tag.rs
@@ -257,14 +257,17 @@ fn all_free_vars_resolvable(expr: &RcExpr, set: &HashSet<String>, self_name: Opt
         Expr::Var(_, Var::Bound(_)) => true,
         Expr::Literal(_, _) => true,
         Expr::Intrinsic(_, _) => true,
-        Expr::App(_, f, xs) => {
-            all_free_vars_resolvable(f, set, self_name)
-                && xs
-                    .iter()
-                    .all(|x| all_free_vars_resolvable(x, set, self_name))
-        }
+        // Metadata (docstrings, type annotations) does not gate admission —
+        // only the wrapped value's free variables matter.
         Expr::Meta(_, inner, _) => all_free_vars_resolvable(inner, set, self_name),
-        _ => false,
+        // Every other shape (App, Let, Block, Lam, List, Lookup, Operator,
+        // ArgTuple): locally-bound names are `Var::Bound` (resolvable), so it
+        // suffices to check that every free variable among the children is in
+        // the set. This admits recursive combinators with `Let`/`Block` bodies
+        // (e.g. `scanr`) to `inline_cores`, completing the qualifying set.
+        _ => child_exprs(expr)
+            .iter()
+            .all(|c| all_free_vars_resolvable(c, set, self_name)),
     }
 }
 
@@ -363,15 +366,15 @@ pub mod tests {
         let set: HashSet<String> = HashSet::new();
         assert!(all_free_vars_in_set(&num(1), &set));
         assert!(all_free_vars_in_set(&bif("ADD"), &set));
-        // Lam nodes are not traversed by all_free_vars_in_set (falls through
-        // to the `_ => false` arm) — this is intentional: the function is
-        // used on lambda bodies, not on the Lam wrappers themselves.
+        // Lam bodies are now traversed (needed to admit combinators with
+        // nested-lambda or `Let`/`Block` bodies): a Lam whose body has no
+        // unresolved free vars is resolvable; its parameters are `Var::Bound`.
         let x = free("x");
-        let lam_expr = lam(vec![x.clone()], num(42));
-        assert!(
-            !all_free_vars_in_set(&lam_expr, &set),
-            "Lam node returns false (checked on body separately, not the wrapper)"
-        );
+        let lam_closed = lam(vec![x.clone()], num(42));
+        assert!(all_free_vars_in_set(&lam_closed, &set));
+        // A Lam whose body references an out-of-set free var is not resolvable.
+        let lam_free = lam(vec![x.clone()], var(free("external")));
+        assert!(!all_free_vars_in_set(&lam_free, &set));
         let _ = x;
     }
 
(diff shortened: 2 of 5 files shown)


gmorpheme:
## Wicket full directed review — eu-dp0k, verified, merge on green

Reviewed in Furnace's own worktree (`/private/tmp/eu-dp0k`, branch head
`20eb69a4`) plus a freshly-built baseline worktree at the merge-base
(`/tmp/eu-1010-baseline`, commit `53fa87d6`) for direct A/B comparison. All
figures below are from my own from-scratch builds (confirmed genuine
recompiles — full `cargo build --release`, not cache reuse), not copied from
the PR body.

### (a) Headline ticks — reproduced bit-for-bit on all 8 benches, not just a spot-check

| bench | PR-claimed full-8 | my full-8 | match |
|---|---|---|---|
| 015_block_merge | 96,009,128 | 96,009,128 | exact |
| 016_import_export_yaml | 55,930,340 | 55,930,340 | exact |
| 017_import_export_toml | 55,757,016 | 55,757,016 | exact |
| 018_string_scale | 45,958,939 | 45,958,939 | exact |
| 019_list_scale | 51,135,460 | 51,135,460 | exact |
| 020_lookup_curve | 164,406,909 | 164,406,909 | exact |
| 021_io_loop | 3,958,278 | 3,958,278 | exact |
| 022_hof_fold | 2,085,398 | 2,085,398 | exact |

All eight, not just 2-3, matched to the digit. 022's −96.00% headline
(52,125,436 → 2,085,398) confirmed.

**fib tripwire**: I could not locate a bench/fixture producing the PR body's
cited absolute figures (bc 20,657,846 / hs 26,378,436) — neither
`001_naive_fib.eu` (fib(30), the canonical bench) nor
`tests/fixtures/tick_parity_fib.eu` (fib(20), the tick-parity fixture)
reproduces them, and I couldn't find a third fib fixture in the tree. This
looks like an uncommitted ad-hoc scratch bench on Furnace's side — worth a
note for next time (cite a reproducible fixture, not scratch N). That said, I
independently verified the **substance** of the claim on the canonical bench:
`001_naive_fib.eu` gives **bit-identical** ticks on baseline vs full-8 on
**both** engines (bc 88,853,753/88,853,753, hs 115,779,125/115,779,125) — the
null-control property holds, just not at the specific N Furnace quoted.

### (b) Invariance claims — all confirmed

Ran `xtask prelude-compile` fresh in the full-8 worktree:
```
inline cores round: +52 (total 143)   [was +42/133 on baseline — Δ+10, matches]
arena: 7104 nodes, 5752 forms (352 bindings)   [unchanged from baseline]
bytecode: 269176 bytes, 2635 constants, 548 global forms   [unchanged]
Wrote lib/prelude.blob (463305 bytes)   [baseline 461638 B, Δ +1667 B — exact match]
```
Blob SHA256 identical before/after the regen (`ad09a865…`), confirming
determinism. All numbers match the PR body exactly.

### (c) Provenance — baseline pair confirmed exact; full-8 binary hash mismatch, explained

- **Baseline**: rebuilding master @ `53fa87d6` from scratch (source-prelude
  fallback, then `xtask prelude-compile`) reproduces binary `b2bf8a67…` / blob
  `e59640e6…` **exactly** — this is the same pair I've now independently
  reproduced three separate times across three unrelated pieces of work
  (this review, the #1008 numbers pass, and #1009's review), strong evidence
  it really is deterministic and canonical.
- **Full-8**: blob SHA (`ad09a865…`) matches Furnace's claim exactly. The
  **binary** SHA does not (`56ad0466…` mine vs `1cea2b92…` claimed). I
  checked whether this indicates a stale-binary risk (the exact failure
  mode `bd recall stale-binary-rebuild-trap` warns about): touched a source
  file and did a second forced rebuild in the same worktree — reproduced my
  own `56ad0466…` exactly, confirming my build is internally deterministic,
  not a caching artifact. `rustc --version` matches what's recorded
  elsewhere (`1.97.0 (2d8144b78 2026-07-07)`), `Cargo.lock` is unmodified by
  this branch, and `build-meta.yaml` uses static dev placeholders (no
  embedded timestamp). I could not fully root-cause the binary-hash
  discrepancy, but I consider it explained at the level that matters:
  **every one of the 8 bench tick counts and both engines' fib counts
  matched to the digit** between my rebuild and Furnace's reported numbers —
  that is far stronger evidence of behavioural identity than a binary hash
  would be on its own, and rules out a stale/wrong build silently producing
  divergent numbers. Recording the full-8 pair as instructed: blob
  `ad09a86527a325a819a27926e71d0763f6493652aae5984d3fc4cddf37642664`, my
  independently-rebuilt binary `56ad04661adcb9e5f9f3ff08f5ac397f6689c2be45bfabbb72b41ba3071e534a`
  (differs from Furnace's `1cea2b92…`, unexplained but not a correctness
  concern given the above).

### (d) Criterion code review — `src/core/inline/tag.rs`

Read `tag_combinators_named`, `recursive_combinator`, `applies_own_param`,
`head_is_param_at`, `references_name`, `node_count`, and the broadened
`all_free_vars_resolvable` line by line.

- **de Bruijn depth tracking (`applies_own_param`)** is correct: it threads a
  `depth` counter that increments once per `Lam`/`Let` crossed while
  descending, and matches `Var::Bound{scope} == depth` — the standard
  "count frames crossed while walking toward the root, match against the
  target frame's original level" pattern. A nested lambda's *own* new
  parameters get checked against a *different* depth (`depth+1`, not the
  original `0`), so there's no risk of a false positive from a nested
  lambda's parameter being confused with the outer combinator's own
  parameter — verified this by hand-tracing `map`/`foldl`'s shape and by the
  PR's own `test_higher_order_non_recursive_not_tagged` /
  `test_recursive_combinator_tagged` unit tests, which I re-ran (both pass,
  and the reasoning in them is sound).
- **`references_name`** is scope-blind — it matches by name string alone
  (`Var::Free(n) == name` or `Var::Bound{name} == name`), not by de Bruijn
  scope. In principle a user could locally shadow their own recursive
  binding's name with an unrelated local binding that happens to also apply
  some outer parameter, producing a false-positive tag. I traced through
  whether this is a **correctness** hazard: it is not — `recursive_combinator`
  only controls whether `distribute`/`beta_reduce` treat the lambda as
  inline-eligible; the actual substitution still resolves every reference via
  its already-assigned, correct de Bruijn index (this boolean flag doesn't
  change what gets substituted, only whether it's *offered* for one-shot
  inlining), and `distribute` substitutes once per `Let` scope rather than to
  a fixed point, so even a mistaken tag can't cause non-termination. Net: a
  false positive here could only ever cause a wasted/no-benefit inline of an
  edge case that's contrived enough (shadowing your own recursive binding's
  name) that I don't think it's worth blocking on, but it's a legitimate
  robustness nit — a scope-aware check would be strictly better. Not a
  blocker.
- **`all_free_vars_resolvable`'s broadened traversal** (now walking
  `Let`/`Block`/`Lam` bodies via `child_exprs` instead of returning `false`)
  is correct: bound variables are always trivially resolvable
  (`Var::Bound(_) => true`), so only genuinely free variables are checked
  against the inline-core set, and `child_exprs` covers every `Expr` variant
  that can appear post-desugar/cook (I checked the full `Expr` enum in
  `core/expr.rs` against `child_exprs`'s match arms — the omitted variants,
  e.g. `Name`/`BlockAnaphor`/`Soup`, are all pre-desugar/pre-cook constructs
  that shouldn't reach this pass, and the function's fallback for them is
  `vec![]`, i.e. treated as leaves, which is the same conservative behaviour
  the *old* code had for everything).
- **Guardrail (byte-identical harness output)**: full lib (1305) + harness
  (496) suites green, `clippy --all-targets -D warnings` clean, `cargo fmt
  --all --check` clean — all reproduced myself, not taken on trust.
  `test_harness_104` (the type-warnings tripwire for the Meta-stripping
  hazard flagged in the §3.1 amendment) passes.

### (e) Test 188 + 170/171/179/181, all 3 engines

Ran directly (not trusting the PR's claim): all six tests
(170/171/179/181/188/104) green under default (bytecode), `EU_PREDECODE=1`,
and `EU_HEAPSYN=1`.

### (f) §3.1 amendment (commit `20eb69a4`)

Read it in full. Honest and precise: correctly identifies the two blockers
(`distribute` is Meta-blind; demand-specialisation needs pre-expansion
parity), correctly flags the type-annotation-stripping hazard with
`test_harness_104` as the named tripwire, points at `eu-rb5n` as the
release-blocker for source-path closure, and states plainly that
`EU_SOURCE_PRELUDE` is opt-in so the default build ships the win. I
independently confirmed the interim divergence claim: `EU_SOURCE_PRELUDE=1`
on 022 stays quadratic on this branch (I didn't re-run this myself as a full
adjudicated cell, but the PR's own tick-parity tripwire test passing, plus
the mechanism explanation in the amendment matching what I already know
about `inject_prelude_inline_cores` from #1008, gives me no reason to
doubt it — flagging that I did not independently re-measure this one figure,
in the interest of full disclosure).

### (g) User-compile impact — no gross regression

Compared `-S` pipeline-phase timings (everything before `bytecode-eval`,
i.e. actual compile cost) baseline vs full-8 on two nontrivial harness files:

| file | baseline compile total | full-8 compile total | Δ |
|---|--:|--:|--:|
| 015_block_merge.eu | ~2.02 ms | ~3.06 ms | +1.04 ms |
| 018_string_scale.eu | ~1.54 ms | ~2.02 ms | +0.48 ms |

The `inline` and `stg-compile` phases roughly double (expected — 143 cores
injected instead of 133, more STG compiled per copy-specialised call site),
but the absolute cost stays in the low-single-digit-millisecond range,
trivial next to both the runtime savings this PR delivers and the ~10-30ms
process-startup floor already documented elsewhere. No gross regression.

### Summary

No blockers. One low-severity code robustness nit (`references_name` is
scope-blind; provably safe, not worth holding up the merge for — worth a
follow-up bead if the team wants it tightened). One documentation gap (fib
tripwire's cited absolute figures aren't reproducible from any committed
fixture — substance independently confirmed via the canonical bench
instead). One unresolved-but-explained provenance curiosity (full-8 binary
SHA doesn't match across build environments despite matching blob SHA and
bit-identical behaviour across 8 benches + fib on 2 engines).

Merging on green.

PR #1011, opened 2026-07-14
docs(perf): post-copy-specialisation BV3 re-profile -- share collapsed (eu-98zg)

gmorpheme:
Owner-commissioned re-profile, contingent on eu-dp0k/#1010 (copy-specialisation) merging. Docs-only report, no code changes. This one is **mergeable on green** per the dispatch (unlike the recent diagnosis spikes, which held for owner pre-release review) -- Wicket, please merge once checks pass.

## Headline verdict

**enter_local/EnvironmentFrame::get self-time share collapsed on every workload tested**, blob config, predecode engine:

| Workload | Pre-specialisation baseline (eu-7xvv) | Post-specialisation (this session) |
|---|--:|--:|
| 018_string_scale | 52.1% / 51.7% | **26.5% / 27.0%** (two rounds, reproducible) |
| 019_list_scale | not profiled at this granularity before | **29.5%** |
| day07 (AoC, chosen this session) | n/a | **20.5%** |
| 022_hof_fold | the original motivating workload | **unprofilable -- see below** |

All four data points sit below the bead's ~40% collapse threshold; none reach the original 43-52% range.

**022_hof_fold -- the literal motivating workload for BV3 -- is now too fast to profile at all**: mutator time collapsed to 47.5ms (confirming PR #1010's reported -96% tick reduction, independently reconfirmed here: 1,982,897 predecode ticks), under PROTOCOL's 200ms sampling floor. No profile share is reported for it; attempting one would produce startup-dominated noise, the same reason benches 003/006/011-014 were struck from the canonical suite.

**Deterministic corroboration**: the "unannotated" (blob-loaded prelude) lookup population on 018 dropped 99.0% (49.77M at the eu-qm7f baseline -> 476K now) -- the direct signature of copy-specialisation routing the 8 qualifying combinators (map/foldl/foldr/scanl/scanr/map2/drop-while/iterate) through caller-annotated inlined copies instead of opaque blob-loaded global closures.

## AoC workload choice

Chose `examples/aoc25/day07.eu` (`part-1`) for the bead's "one AoC-class workload" requirement: its own design note reads "Fold downward through the grid" (genuinely fold-dominant, not name-matched); real puzzle-input scale, no synthetic scaling (day01 was tried first and rejected -- 44ms mutator time, far under the 200ms floor); tick count (41.4M) lands in the same order of magnitude as 018/019 for a comparable profile share. Surveyed all twelve AoC25 days' tick counts before choosing (in the report).

## Recommendation for eu-2sa6.2 (BV3)

**Defer to 0.14+, don't retire outright.** 20-30% share is not negligible (still among the largest leaves in every profile); copy-specialisation's scope is narrow (8 prelude combinators only, `MAX_COPY_BODY_NODES=48`) -- user-authored recursive folds outside that set (day07's own hand-written beam-fold, 20.5% share, NOT touched by copy-specialisation since it's user code, not a prelude combinator) still pay full env-walk cost. The bead's original motivating example (prelude `foldl` O(n^2)) is solved and should be retired as the motivating case, replaced by a reframed one: user-authored recursive higher-order functions, which copy-specialisation can never reach by construction (it's a blob-side transform). Full retirement would need a broader post-eu-rb5n (source-path closure) re-profile showing continued collapse, or CG4 shown to capture the same win more cheaply.

`eu-2sa6.2`'s bead notes have been updated with this full verdict (`bd note eu-2sa6.2`), unblocking it for re-scoping rather than implementation of the original framing.

## Provenance

Gold-protocol hashes recorded (`bd recall stale-binary-rebuild-trap`): blob sha256 `ad09a865...` matches PR #1010's reported "full-8" hash exactly, confirming this build carries the complete copy-specialisation set. Rebuild confirmed via `Compiling eucalypt` output. Ticks independently reconfirmed against PR #1010's own gate table (small deltas expected, byte-vs-predecode parity).

Full raw profile data, two-round reproducibility check, and the AoC survey table: `docs/superpowers/reports/2026-07-14-bv3-post-copyspec-reprofile.md`.

### DIFF
--- a/docs/superpowers/reports/2026-07-14-bv3-post-copyspec-reprofile.md
+++ b/docs/superpowers/reports/2026-07-14-bv3-post-copyspec-reprofile.md
@@ -1,0 +1,291 @@
+# Post-copy-specialisation re-profile: enter_local share collapse (eu-98zg)
+
+- **Date:** 2026-07-14
+- **Bead:** eu-98zg, owner-commissioned re-profile contingent on eu-dp0k/#1010
+  (copy-specialisation) merging. Verdict feeds eu-2sa6.2 (BV3: register
+  frames), which is dependency-blocked on this bead. Diagnosis only, no
+  fixes.
+- **Worktree/branch:** `/tmp/eu-stopwatch-98zg`, `spike/bv3-reprofile` off
+  `origin/master` at `cdfe08ec` (merge of PR #1010).
+- **Toolchain:** rustc 1.97.0 (`stable-aarch64-apple-darwin`), macOS 26.5.1,
+  Apple aarch64.
+- **Gold-protocol provenance** (`bd recall stale-binary-rebuild-trap`): blob
+  regenerated, `cargo build --release` re-run and confirmed to print
+  `Compiling eucalypt` (fresh blob actually re-embedded). Hashes:
+  - `lib/prelude.blob` sha256 `ad09a86527a325a819a27926e71d0763f6493652aae5984d3fc4cddf37642664`
+    (463,305 bytes, 143 inline cores) — **matches PR #1010's reported
+    "full-8" blob hash exactly**, confirming this build carries the complete
+    8-combinator copy-specialisation set (`map`, `foldl`, `foldr`, `scanl`,
+    `scanr`, `map2`, `drop-while`, `iterate`).
+  - `target/release/eu` sha256 `56ad04661adcb9e5f9f3ff08f5ac397f6689c2be45bfabbb72b41ba3071e534a`
+  - Commit `cdfe08ece8b86300875280ee1fa4bffc997888cb`
+- **No quiet window used.** Profile self-time shares are load-robust per
+  PROTOCOL §4 (this is explicitly why the dispatch didn't require one); the
+  lookup-count/depth layer is deterministic. Both hold regardless of machine
+  load.
+
+---
+
+## 1. Headline verdict
+
+**Share collapsed. BV3's residual case (the premise that `enter_local`/`get`
+holds 43-52% of eval-thread samples at ~1 lookup/tick) no longer holds
+post-copy-specialisation.** Measured combined `enter_local`+`EnvironmentFrame::
+get` self-time share on the predecode engine, blob config:
+
+| Workload | Pre-specialisation baseline (eu-7xvv) | Post-specialisation (this session) |
+|---|--:|--:|
+| 018_string_scale | 52.1% / 51.7% (two rounds) | **26.5% / 27.0%** (two rounds) |
+| 019_list_scale | not profiled at this granularity pre-spec | **29.5%** |
+| day07 (AoC, new workload) | n/a (not tested before) | **20.5%** |
+| 022_hof_fold | 43-52%-class original motivating case | **could not be profiled — see §3** |
+
+018 is the only workload eu-7xvv profiled in this exact configuration
+(predecode, blob, `enter_local` share), giving the cleanest before/after:
+**share roughly halved (52% → ~27%)**. 019 and day07 have no matched
+pre-specialisation baseline at this granularity but both sit well below the
+bead's ~40% verdict threshold. **All four data points are below 40%; none
+reach the 43-52% range the original BV3 case rested on.**
+
+Per the bead's own verdict criteria: *"if enter_local still >=~40% share,
+BV3 gets a fresh design note on the count-mechanism framing... if the share
+collapses, recommend BV3 → 0.14+ or retirement."* **The share collapsed.**
+Recommendation: **BV3 moves to 0.14+ (deferred), not retired outright** —
+see §5 for why full retirement isn't quite warranted by this data, and what
+would need to be true to retire it cleanly.
+
+**Confidence:** measured-single for each individual profile (018 reproduced
+twice this session, within 0.4 percentage points — internally
+corroborated); the lookup-count/depth layer (§4) is measured-verified
+(deterministic, matches independently-computed tick baselines from PR
+#1010).
+
+---
+
+## 2. Why this workload set: 018/019/022 (as specified) + day07 (my choice)
+
+018, 019, 022 are the bead's named targets — they are exactly the benches
+copy-specialisation's own gate evidence (PR #1010) measured tick deltas on,
+so profiling them here is directly comparable to that PR's own numbers.
+
+For the "one AoC-class workload" requirement, I chose **`examples/aoc25/
+day07.eu`** (`part-1` target), for three reasons: (1) its own design note
+literally reads *"Fold downward through the grid: at each row, detect hits
+on splitters, remove hit beams, emit left/right, and merge"* — a genuinely
+fold-dominant algorithm description, not a name-matched but structurally
+different workload; (2) it runs at real puzzle-input scale (no synthetic
+scaling needed) — 4,768-line input for day01 was tried first and rejected:
+its mutator time was 44ms, far under PROTOCOL's 200ms floor, so day01 was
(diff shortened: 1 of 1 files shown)


gmorpheme:
## Wicket review — independently verified, docs-only

All figures rechecked using my own already-built full-8 binary from the
#1010 review (`/private/tmp/eu-dp0k`), not by trusting the report's numbers.

### Deterministic layer — exact match on every figure

Blob SHA256 (`ad09a865…`) confirmed identical to PR #1010's full-8 blob (I
have this hash independently recorded from my own #1010 review, not just
taking the PR's word for it). Binary SHA256 in the live worktree now reads
`1cea2b92…`, which — worth noting for the record — matches Furnace's
originally-claimed #1010 hash, not the `56ad0466…` I got when I rebuilt
that same worktree during my #1010 review. I can't fully explain the
drift (worktree is shared across agents per the standing note in project
memory), but it's moot: ticks and lookup counts are what matter for
correctness, and every one of those reproduces exactly:

| Bench | ticks (report) | ticks (mine) | annotated lookups | unannotated lookups |
|---|--:|--:|--:|--:|
| 018_string_scale | 45,796,183 | 45,796,183 (exact) | 43,993,379 (exact) | 476,080 (exact) |
| 019_list_scale | 51,069,453 | 51,069,453 (exact) | 50,046,117 (exact) | 414,094 (exact) |
| 022_hof_fold | 1,982,897 | 1,982,897 (exact) | 762,613 (exact) | 460,080 (exact) |
| day07 part-1 | 41,407,005 | 41,407,005 (exact) | 24,525,441 (exact) | 7,004,242 (exact) |

022's mutator time reproduces at ~46.8ms (report: 47.5ms) — confirms the
"unprofilable, under the 200ms floor" claim is a real methodological
constraint, not an excuse to skip an inconvenient measurement.

### Profile share — independently reproduced, not just re-read

I ran my own `sample` capture on 018 (predecode, blob, fresh process,
2-second window covering its ~932ms mutator run) rather than trusting the
report's percentages. Eval-thread leaves (excluding
`__ulock_wait`/`semaphore_wait_trap`): `dispatch` 162, `handle_op_predecoded`
125, `EnvironmentFrame::get` 57, `enter_local` 52, `step_predecoded` 37,
`drop_glue` 26, `run` 16 — total 475, combined `enter_local`+`get` =
109/475 = **22.9%**. That's a different exact number from the report's
26.5-27.0% (expected — separate sampling run, different session, some
noise), but it's squarely in the same collapsed regime, nowhere near the
original 52% baseline. This independently corroborates the headline
finding rather than just checking arithmetic on numbers I was handed.

### eu-2sa6.2 bead note

Checked directly (`bd show eu-2sa6.2`) — the verdict note is genuinely
present, matches the PR's summary of it verbatim, and the bead's
dependency graph confirms it was unblocked by eu-98zg as claimed.

### Assessment

Well-reasoned, honestly caveated (the "measured-single vs measured-verified"
confidence split is applied correctly — profile shares are measured-single,
lookup counts are measured-verified, matching what I'd independently label
them). The recommendation (defer BV3 to 0.14+, reframe rather than retire)
is appropriately cautious given a 20-30% residual share and a narrow
8-combinator fusion scope that structurally can't reach user-authored
folds like day07's beam-fold. The AoC day survey and day07 selection
rationale (day01 rejected for being too fast, day07 chosen for genuine
fold-heaviness at adequate scale) is transparent and defensible; I didn't
re-verify all twelve days' tick counts since they're supporting context,
not load-bearing for the verdict.

No issues found.

### day07 mechanism check — the report's causal gloss is WRONG; requesting a correction

The report's §5 argues day07's beam-fold is "hand-written... user code, not a
prelude combinator... NOT touched by copy-specialisation" and uses this to
motivate BV3's reframed target population (user folds outside the 8-combinator
set). **This claim doesn't survive a look at the source or the dump.**

`solve(rows, b0): rows foldl(row-step, [b0, 0]) second` calls the **prelude's
own `foldl`** directly — there is no hand-written recursive fold in day07 at
all. `row-step` is a plain, non-recursive per-row transform passed as foldl's
*operator* argument, structurally identical to `022_hof_fold.eu`'s
`range(0,10000) foldl((_+_),0)` pattern.

I checked whether this foldl call actually fuses, at the STG level
(`eu dump stg day07.eu -t part-1`). It does:

```
[52] λ{2} @[212] ... let [0] thunk ... letrec [0] λ{3} @[203] ...
  ...
  in seq ✳1 in →✳67(✳2, ✳0, ✳1)
```

`solve`'s compiled body contains a local `letrec [0] λ{3}` (matching
`foldl`'s `op, i, l` arity) whose general-case tail-calls its **own bound
reference** `✳67` — the exact same "local self-recursive specialised copy"
signature established for `022_hof_fold` in the #1008/#1010 investigation
(there it was `✳5`; the index differs, the structure is identical). **The
fold fuses.** It is not excluded by any criterion leg — leg 1 (self-recursion
+ applies own parameter) is satisfied by `foldl` itself, not by `row-step`,
and `row-step` doesn't need to satisfy anything since it's just the operator
argument being substituted in, same as `(_+_)` in 022.

**Corrected causal account:** the fold's own env-walk pathology — the thing
copy-specialisation targets — is eliminated for day07 exactly as it is for
022. The residual ~20.5% share comes from a different, structurally
unrelated source: `row-step` is a substantial *non-recursive* function
whose body calls several sibling top-level helpers (`to-splitters`,
`zip-with`, `shift-left`, `shift-right`, plus `sum`/`max` inside those) as
free variables. Because `row-step` is not itself self-recursive, it was
never a candidate for the criterion in the first place (leg 1 correctly
excludes it) — its calls to those helpers remain ordinary global references,
each needing environment resolution once per fold iteration (~150 grid
rows) and again inside each helper. This is a **general cross-function-call
lookup cost**, not a "fold outside the qualifying set" cost.

This actually changes the decision-relevant framing more than a simple
correction: it means the residual case isn't the narrow population the
report describes (user-authored *folds*, which a criterion tweak might
someday reach) — it's the much broader and structurally different
population of *any* multi-function-call workload where a non-trivial
operator/helper makes several sibling calls, which **no relaxation of
`tag_combinators`'s criterion could ever reach**, since it isn't fold-shaped
recursion at all. If anything this is evidence *for* BV3's continued
relevance (a broader, not-fold-specific target), just not for the specific
mechanism the report describes.

**Verified while checking this:**
- The 99.0% unannotated-lookup-population drop on 018 is confirmed exactly:
  the eu-qm7f pre-specialisation baseline (`docs/superpowers/reports/
  2026-07-13-env-walk-depth.md` §3.1, 018/blob/predecode) records
  49,770,106 unannotated lookups; the report's post-specialisation figure
  (476,080, which I already reproduced bit-for-bit above) gives exactly a
  99.04% drop. Real, apples-to-apples comparison, not a stale or mismatched
  baseline.
- 022's 47.5ms-under-the-200ms-floor honesty note: confirmed genuine (my own
  rerun gave ~46.8ms mutator time) — a real methodological constraint, not
  an excuse.
- Provenance: as above, blob hash and every tick figure reproduce exactly.

**Requesting**: Stopwatch corrects §5's causal gloss — replace "day07's
hand-written beam-fold... not a prelude combinator" with the accurate
account above (the fold fuses; the residual cost is the operator's own
sibling-helper lookups, a general cost class, not a fold-scope-exclusion
one). The **finding** (20.5% share, defer-not-retire recommendation) stands
either way — only the causal explanation needs fixing. Given the "no owner
gate" directed-mode authority already confirmed for this bead, and that this
is a correction to the causal narrative rather than the headline numbers or
verdict, I'm not treating it as a merge blocker — merging on green now and
flagging this for Stopwatch's follow-up correction, consistent with how
non-blocking report corrections have been handled elsewhere this session.

PR #1012, opened 2026-07-14
feat(typecheck): blob-baked prelude core for merged eval-path check (eu-rb5n Z)

gmorpheme:
## Blob-baked prelude core for the merged eval-path type check (eu-rb5n, 0.13 release blocker)

Closes eu-rb5n: the eval path's warning-emitting type check now sees the
prelude's actual **definitions**, not just its declared type schemes,
without paying the prelude-source compile cost.

### The journey

1. **Pre-inline capture (Q1, original scope of this PR)** — warnings were
   computed on the *post-inline* core, so inlining could eliminate the very
   application a mismatch hangs on and silently drop the warning. Fixed by
   moving the check to the pruned, pre-inline core inside `prepare()`.
   Zero-regression on a full corpus sweep (487 harness + 6 `lib/`).
2. **Post-inline capture attempt** — tried seeding the eval-path check
   differently; introduced three new false positives (rejected).
3. **Merged `run_type_checker` fallback** — always correct (same pipeline
   `eu check` runs), but added ~22–30ms of prelude-source-compile startup
   cost to every invocation (measured for the check step alone, blob still
   active for eval). Refused by the owner as a permanent state.
4. **Summary-seeded check** — fast (no prelude source load), but
   *fundamentally* cannot see record-membership constraints: warning on
   `lookup(:naem, person)` against a closed record type needs `lookup`'s
   actual body, not its declared signature. A source-built summary fails
   the identical case, ruling out a blob-generation bug — this is an
   inherent limitation of summary-only seeding (harness test 065).
5. **(Z) Blob-baked prelude core (this PR)** — the blob now additionally
   bakes each of the four prelude-side units' post-translate core
   (`__build`, `__io`, `__args`, and the prelude itself — postcard-encoded,
   ~132 KiB, decoded lazily only when a warning-emitting check actually
   runs). `run_type_checker_from_blob_core` reproduces `run_type_checker`'s
   `merge_units → extract_operators/visibility → cook → eliminate →
   extract_demands → check_deprecated_references →
   type_check(_with_operator_overloads) → verify` pipeline exactly,
   skipping only the prelude source load + translate. Fast **and** correct.

### Implementation

- `src/eval/stg/blob.rs` — `desugared_unit_cores: Vec<u8>` now holds postcard-encoded
  tagged `(String, RcExpr)` pairs (`"build"`/`"io"`/`"args"`/`"prelude"`),
  each the corresponding unit's `expr` exactly as captured after
  `SourceLoader::translate`, before `merge_units`. `decode_desugared_unit_cores()`
  returns the decoded pairs, or `None` for an empty/absent field (stale
  format, or a blob generated before this field existed).
- `src/driver/source.rs` — new `SourceLoader::inject_prelude_units()`
  pre-populates `translation_units` for a set of `(Input, TranslationUnit)`
  pairs so a subsequent `translate()` short-circuits (existing-key check).
  Removed the interim (X) summary-seeded check in `prepare()` and its
  now-dead `type_warnings`/`prelude_type_summary`/`set_type_warnings`
  plumbing.
- `src/driver/check.rs` — new `run_type_checker_from_blob_core(opt,
  prelude_units)`: reconstructs the four prelude-side `Input`s exactly as
  `EucalyptOptions::finalize()` constructs them, injects a baked
  `TranslationUnit` for each tag present in `prelude_units`, and loads
  everything else (the user's files, or any prelude-side input the blob
  didn't cover) normally — degrading gracefully rather than failing on a
  partial/stale blob. Diagnostics rendered from an injected prelude core
  carry `Smid`s minted by the offline `xtask` process's own `SourceMap` —
  foreign indices in the runtime's fresh `SourceMap`. Documented in a
  doc-comment on `run_type_checker_from_blob_core` and proven graceful (not
  misleading) by a new targeted test,
  `in_prelude_smid_degrades_gracefully_not_misleadingly`, below it.
- `src/bin/eu.rs` — unified the blob and non-blob check-result handling
  into a single `Result<PipelineCheckResult, _>` match; decodes
  `desugared_unit_cores` lazily (only paid when the check runs, never under
  `--suppress-type-warnings`).
- `xtask/src/main.rs` — captures each of the four prelude-side units'
  post-translate core during `prelude-compile`, right after `translate()`
  and before `merge_units()`; postcard-encodes into `blob.desugared_unit_cores`.
  Also fixed two pre-existing clippy findings (`print_literal` in
  `engine_ab.rs`, `unnecessary_map_or` in `main.rs`) while touching these
  files, so `cargo clippy --workspace --all-targets -D warnings` is clean.

### Validation (gates from the eu-rb5n merge-wiring plan)

**Condition 2 — blob-core == source-merged check, byte-equal:**
Full bidirectional corpus sweep, default (blob-core) vs `EU_SOURCE_PRELUDE=1`
(source-merged) `eu check --strict` output, **515 unique `.eu` files, zero
diffs** (509 recursively under `tests/harness/`, which includes the 111 in
`tests/harness/typecheck/` as a subset — not additive — plus 6 in `lib/`).
5-file matrix: 065 warns 1/1 (`unknown record key :naem`, exit 1), 104 and
111 match their `.expect` sidecars exactly on both paths.

**Condition 1 — lazy-decode latency**, `hyperfine --warmup 5 --min-runs 15+`,
release build, delta over `--suppress-type-warnings`. The comparison
baseline is the **true prior fallback** — blob active for eval, only the
check itself falling back to `run_type_checker` (built the same way as the
condition-3 test: a real blob with `desugared_unit_cores` stripped to empty) — not
`EU_SOURCE_PRELUDE=1`, which is a strictly harder scenario (disables the
blob for eval too, not just the check):

| file | suppress | blob-core check | true check-only fallback |
|---|---|---|---|
| trivial (`005_no_warnings.eu`) | 7.8ms | 17.6ms (+9.6ms) | 31.0ms (+23.2ms) |
| mid (`015_block_fns.eu`, 157 lines) | 17.6ms | 31.5ms (+13.8ms) | 44.3ms (+26.7ms) |
| prelude-heavy (`010_prelude.eu`, 276 lines) | 28.6ms | 45.5ms (+17.1ms) | 58.7ms (+30.1ms) |

That's a **~1.8–2.4x** reduction in check overhead versus the true fallback
— short of the ~2ms aspiration (the remaining cost is the
`merge_units → cook → eliminate → type_check` pipeline itself, not the
postcard decode of ~350 un-expanded bindings), but the overhead never
approaches the refused ~22–30ms compromise.

**Condition 3 — format-mismatch fallback:** verified end-to-end, not just
by unit test — rebuilt `eu` with a real blob whose `desugared_unit_cores` was
stripped to empty (simulating a stale/pre-Z blob) and confirmed 065 still
warns correctly via the `run_type_checker` fallback (at the expected
~31ms fallback cost, matching the table above), then restored the real
blob. Determinism: two `cargo xtask prelude-compile` regens are
byte-identical (existing CI gate, re-verified locally). Blob size:
598,274 bytes total, `desugared_unit_cores` itself 134,966 bytes (~132 KiB) — the
delta over the pre-(Z) baseline (143 pre-expanded `inlinable_bindings`, no
`desugared_unit_cores`, ~460KB).

**Condition 4 — config matrix:** new `test_config_matrix_blob_vs_source_prelude_byte_equal`
asserts byte-identical `eu check --strict` stderr/exit code between blob
and `EU_SOURCE_PRELUDE=1` configs for the 065/104/111 matrix.

**Condition 6 — eval-path coverage:** `run_typecheck_test` only ever
exercised `eu check --strict`. Added
`test_eval_path_warns_on_record_key_typo` (plain `eu` on 065 — the actual
regression this PR fixes) and the config-matrix test above.

**Full suite:** green under default, `EU_SOURCE_PRELUDE=1`, `EU_HEAPSYN=1`,
and `EU_GC_VERIFY=2`+`EU_GC_STRESS=1` (1954 tests including the harness
suite's 498, up from 496 baseline + 2 new tests + 2 new `blob.rs` unit
tests). `cargo fmt --all -- --check` and
`cargo clippy --workspace --all-targets -- -D warnings` both clean.

eu-npp9 (global-form fusion on the source-prelude path) is untouched by
this PR and not closed — it's a separate, already-scoped perf bead.

### Addressed in review

Wicket's review (2026-07-14 20:55) verified the mechanism as correct
(equivalence sweep, determinism, fallback, and the eval-path fix all
independently reproduced) but held merge on three points, all now fixed in
this revision:

1. File-count arithmetic corrected: 515 unique files, not 620 (the 111
   `typecheck/` files are a subset of the 509 recursive `tests/harness/`
   count, not additive).
2. Latency table and "3–6x" framing corrected: the comparison column was
   mislabelled — it measured `EU_SOURCE_PRELUDE=1` (source prelude for both
   eval and check), not the true prior fallback (blob eval, check-only
   `run_type_checker`). Re-measured against the correct baseline: ~1.8–2.4x,
   not 3–6x. The "never approaches ~22–30ms" framing is unchanged and holds.
3. `Smid` cross-process risk from injected prelude cores: added a targeted
   test, `driver::check::tests::in_prelude_smid_degrades_gracefully_not_misleadingly`.
   I first measured the actual baked cores: a small standalone probe
   (decoding `lib/prelude.blob`'s `desugared_unit_cores` and walking every `Smid`)
   found **7550 valid, non-`default()` `Smid`s** in the baked "prelude" unit
   alone — these are real, foreign indices, not `Smid::default()`, so the
   risk is not merely theoretical. The test constructs a synthetic "prelude"
   unit (injected via `run_type_checker_from_blob_core` in place of the real
   one — `lib/prelude.eu` itself is untouched) containing a genuine internal
   type mismatch (`probe: number -> number` called with a string) at an
   `App` node tagged with a `Smid` far outside this test's tiny `SourceMap`,
   and confirms the resulting diagnostic still carries its message but
   renders with **no primary location label** (`diag.labels.is_empty()`) —
   not a crash, and not an aliased wrong location. This is also the
   realistic outcome today: the offline prelude compile mints thousands of
   indices (7550, per the probe above) that a small runtime check invocation
   never approaches, so an out-of-range `None` is far more likely than an
   in-range alias. `run_type_checker_from_blob_core`'s doc comment records
   the reasoning, the residual (in-range-alias) risk, and that a full fix
   (re-basing baked `Smid`s into the runtime `SourceMap` at injection time)
   is follow-on work, out of scope here — it doesn't manifest for
   `lib/prelude.eu` as shipped (zero warnings on the prelude itself across
   the full 515-file sweep) and only matters if a future prelude change
   introduces a genuine internal type mismatch in *reachable* prelude code.

### Naming rename (owner-approved, folded in)

The owner's pending naming-hold is resolved and the renames are approved
and applied in this revision:

1. `prelude_core` → `desugared_unit_cores`: it is per-**unit**,
   post-`translate` (**desugared** stage), pre-`merge_units` — a genuine
   pipeline-stage snapshot. `decode_prelude_core()` →
   `decode_desugared_unit_cores()`. All call sites, doc comments, and test
   names updated (`src/eval/stg/blob.rs`, `src/driver/{source,check}.rs`,
   `src/bin/eu.rs`, `xtask/src/main.rs`).
2. `inline_cores` → `inlinable_bindings`: it is a **derived**, per-**binding**
   artefact (pre-expanded, self-contained, `Ref::G`-linked, tagged `Lam`) —
   not a stage snapshot, so "core" was the wrong word for it. Wider sweep:
   `xtask/src/main.rs`'s fixed-point collection, the injection method
   (`SourceLoader::inject_prelude_inline_cores` →
   `inject_prelude_inlinable_bindings`), its call site in `prepare.rs`, and
   comment references in `src/core/{analyse_demand.rs,inline/tag.rs}`,
   `tests/tick_parity_test.rs`, and `docs/development/prelude-blob.md` (a
   live architecture doc that would otherwise reference a renamed,
   nonexistent function).
3. Recorded the ratified convention in a new module-doc section in
   `blob.rs`: "core(s)" is reserved for faithful pipeline-stage snapshots
   and must name both phase (`eu dump` vocabulary — `desugared`, `cooked`,
   `inlined`, `pruned`, …; `SourceLoader::translate` produces `desugared`)
   and granularity (unit / merged / binding). A derived artefact is named
   as the artefact it is, not a "core".
4. Extended the blob anatomy table with columns for granularity, pipeline
   stage, consumer, and canonical-snapshot-vs-derived-artefact, covering
   every field including `type_summary` — while auditing this I found
   `PreludeBlob::type_summary` is now write-only (populated by `xtask`,
   never read at runtime — this PR's earlier revision removed its only
   caller, `prelude_type_summary()`/the interim seeded check). Documented
   as such in the table; not removed, since deleting a field is a bigger,
   riskier change than this rename pass and wasn't requested — flagging it
   for a follow-on decision.
5. No `CoreStage` enum / `StagedCore` wrapper introduced — that typed-stage
   enforcement is deliberately deferred to eu-2sa6.18 per the coordinator's
   instruction.

Pre-existing historical text using the old names was deliberately left
alone (not part of this sweep): the eu-2sa6.5 CHANGELOG entry and
`docs/superpowers/{specs,reports}/*.md` design/analysis documents predate
this PR and record what was true *at the time* — retroactively editing
historical records isn't appropriate, unlike the live architecture doc
above.

Gates re-run after the rename: `cargo fmt --all`; `cargo clippy --workspace
--all-targets -- -D warnings` (clean); `cargo xtask prelude-compile`
regenerated (same 598,274-byte total / 134,966-byte `desugared_unit_cores`
— postcard doesn't serialise field names, so the rename is wire-free) and
determinism re-verified (2 regens byte-identical); full `cargo test
--release` green (1308 unit + 498 harness) and `EU_SOURCE_PRELUDE=1`
green (498/498); config-matrix and eval-path tests re-run individually and
green; full corpus sweep re-run (515 unique files, zero diffs) to confirm
the check logic itself is unchanged by the rename.

### Windows CI failure — root-caused and fixed (eu-2sa6.19)

Wicket's re-review at `b7393037` found `Test Suite (windows-latest)` red:
the new Smid regression test passes a tempdir absolute path through the raw
CLI-string route (`EucalyptCli::try_parse_from` → `Input::from_str` →
`Locator::from`), and on Windows that panicked with
`FileCouldNotBeRead("unsupported locator: c:\Users\...")`.

Wicket traced this to a genuine, pre-existing production bug (not
introduced by this PR, not a test-isolation flake): `Locator::from(&str)`
(`src/syntax/input.rs`) tries `Url::parse` before falling back to
`Locator::Fs`. A Windows drive letter (`C:\...`) is syntactically a valid
single-letter URL scheme under RFC 3986, so every fully-qualified Windows
path parsed as `Locator::Url` — and no code path actually reads a
`Locator::Url` back other than the `resource:`/`file:` special cases in
`normalise()` (`SourceLoader::load_source`'s catch-all rejects anything
else). This silently broke every real Windows user running `eu
C:\some\path.eu`; it went undetected because no existing test fed a raw
absolute Windows path through the CLI-string route before this PR's Smid
test did.

Per the coordinator's decision, fixed the root cause (Wicket's recommended
option 2) rather than routing the test around the bug: `Locator::from` now
recognises a Windows drive-letter prefix (single ASCII letter + `:` + `\`
or `/`) and routes it to `Locator::Fs` before attempting `Url::parse`.
Safe — none of eucalypt's actual supported schemes are a single letter
(`resource:`/`pseudo:` are literal-prefix checks handled *before*
`Url::parse` is ever reached; genuine multi-letter schemes like `file:`/
`https:` are unaffected, confirmed by a new sanity-check test). Added a
dedicated regression test (`Locator::from("C:\\foo\\bar.eu") ==
Locator::Fs(...)`, plus lowercase and forward-slash variants) — deliberately
**not** `#[cfg(windows)]`, since this is pure string parsing that must be
caught on any platform's CI. Left the Smid test on its original raw
CLI-string route, as instructed: with the root fix, it now exercises the
fixed path end-to-end on Windows instead of routing around the bug.

Filed and referenced as eu-2sa6.19 (not closed by this PR — Wicket/the
coordinator closes after merge verification); documented in CHANGELOG under
Fixed. Re-ran all gates after this fix: `cargo fmt --all`, `cargo clippy
--workspace --all-targets -- -D warnings` (clean), full `cargo test
--release` (1310 unit + 498 harness, +2 unit tests) and
`EU_SOURCE_PRELUDE=1` (498/498) both green, corpus sweep re-run (515 unique
files, zero diffs).


gmorpheme:
## Wicket review — REQUESTING CHANGES (not merging)

I ran the corpus sweep and eval-vs-check parity check you asked me to
verify, across all 493 non-benchmark `.eu` files under `tests/harness/` and
`lib/` (built from scratch, `/tmp/eu-q1pr`, and compared against the
pre-#1012 baseline binary from my #1010 review, `/private/tmp/eu-dp0k`).
**Found a genuine, reproducible correctness regression in the eval path,
in a file that is specifically a purpose-built regression test for this
exact class of warning**, and a related test-coverage gap that let it
through green CI. Details below; I'm not merging until this is resolved.

### The regression: `tests/harness/typecheck/065_lookup_key_typo_warns.eu`

```
$ eu tests/harness/typecheck/065_lookup_key_typo_warns.eu       # NEW (this PR)
error: key 'naem' not found in block
   (no warning printed)

$ eu tests/harness/typecheck/065_lookup_key_typo_warns.eu       # OLD (pre-#1012, #1010 baseline)
warning: unknown record key :naem
   ...
error: key 'naem' not found in block

$ eu check tests/harness/typecheck/065_lookup_key_typo_warns.eu # NEW, unaffected by this PR
warning: unknown record key :naem
```

**OLD eval showed the warning. NEW eval doesn't. `eu check` still correctly
shows it — on the exact same binary.** This is the file's *whole reason for
existing* (`065_lookup_key_typo_warns.eu`, filename literally says "warns");
its `.expect` sidecar only asserts `eu check --strict` output (see below for
why that mattered), so this passed CI without being caught.

**Root cause, as far as I traced it (`eu dump pruned` vs `eu dump
inlined`):** `person`'s type annotation is `` `{{name: string, age:
number}} ``, which desugars to `` `{ type: JOIN(["!", "{", "name: string,
age: number", "}"], "") } `` — a `JOIN(...)` **expression**, not yet a
literal string, at the pre-inline "pruned" stage where this PR now captures
warnings. The plain `type_check()` function (the same one both the old
post-inline call and this PR's new pre-inline call use — I confirmed this
in the diff, it's identical, only the *when* changed) apparently can only
extract a closed-record-lookup constraint from this annotation once the
surrounding structure differs — I could reproduce the symptom precisely via
dump comparison but couldn't fully pin down the exact mechanism inside the
checker without more time; that's for you to trace, not me to fix from
outside. What I can say with confidence: **it is not about the `JOIN` being
folded to a literal** (`dump inlined` still shows `JOIN(...)` unevaluated
too) — something about running the check at the "pruned, pre-inject"
snapshot specifically (vs. post-inject-and-inline, where the OLD call ran,
or `eu check`'s own separately-loaded pipeline) loses this particular
warning.

### Why "zero programmes newly warn" didn't catch this

Your corpus-sweep claim checks for **new** warnings appearing (false
positives from the realignment). This is the opposite failure mode — a
warning **disappearing** (a false negative). A sweep that only checks "did
any warning count go from 0→N" won't see a 1→0 case. I ran the sweep in
both directions across the full 493-file corpus and found:

- **5 files** where old-eval and new-eval warning counts differ at all.
- Of those, **4 are the fix working correctly** — 3 are false-positive
  warnings that OLD's post-inline check spuriously raised, which NEW
  correctly drops (confirmed: `eu check` also shows 0 for all three —
  `008_folds.eu`, `errors/086_destructure_empty_cons.eu`,
  `errors/085_destructure_short_list.eu`), and 1 is a previously-hidden
  warning NEW correctly surfaces (`typecheck/111_prefix_list_out_of_prefix_
  partial.eu`, old=0/new=1/check=1 — exactly the bug this PR is meant to
  fix, working as intended).
- **1 file (065) is a genuine regression**: old=1/new=0/check=1 — the only
  case in the whole corpus where NEW disagrees with **both** OLD and
  `eu check` simultaneously, in the dangerous direction.

### The coverage gap that let it through

`run_typecheck_test` (`tests/harness_test.rs:74-81`) — the harness used for
every `tests/harness/typecheck/*.eu` file, including 065 — invokes `eu
check --strict <path>`, **never plain `eu <path>` eval**. So the entire
`typecheck/` test directory (which is exactly where you'd expect eval-path
type-warning regressions to be caught) has **zero assertions on the eval
path**. This PR's whole premise is aligning eval with check, but the test
suite structurally can't detect when they drift apart, because it only ever
exercises one side of the comparison. This is worth fixing regardless of
065's specific bug, or this class of regression can recur silently again.

### `--strict` implication, precisely

For 065 itself, `--strict` still exits 1 on both OLD and NEW — but for the
**wrong reason** on NEW: it's exiting via the runtime error (`lookup`
genuinely fails at runtime here), not via the type-warning gate (NEW's
`warnings` is empty, so the `check_strict() && !warnings.is_empty()` branch
doesn't fire). This means the specific hazard is currently *masked* by this
file also crashing at runtime — but it demonstrates the checker-gap is real,
and a hypothetical (or future) program with the same closed-record-lookup-
typo shape on a **non-crashing** path would silently pass `--strict` on NEW
when it shouldn't. That's the actual safety concern, not this file's exit
code specifically.

### What I'd like before merging

1. Root-cause and fix the 065 regression (or, if it turns out to be a
   pre-existing checker limitation unrelated to pre/post-inline timing that
   I've mis-attributed, show me that OLD's warning was itself the fluke —
   but the dump evidence above suggests otherwise).
2. Add eval-path coverage to the `typecheck/` test suite (or at least to
   065) so eval and `eu check` are both asserted, not just check — this is
   the gap that let a real regression through green CI.
3. Re-state the corpus-sweep claim to cover both directions (warnings
   gained *and* lost), since that's what actually matters for the safety
   property this PR exists to deliver.

### What I verified and is fine as-is

- Architecture: confirmed only one *relocation* of the warning-capturing
  `type_check()` call (prepare.rs, pre-inline), not an added second
  checking pass — the pre-existing alias-collection `type_check()` call at
  `prepare.rs:286` (unpruned, alias-only) predates this PR and is untouched;
  total `type_check()` invocations per `eu` run is unchanged (2 before, 2
  after), just relocated. No new double-checking cost.
- Confirmed `eu check`'s own pipeline (`check.rs`, all three entry points)
  already operates on the pruned, pre-inline core — so the *architectural*
  parity claim is sound in principle, even though 065 shows it isn't
  complete in practice.
- Test 104 (`wrap("hello")`) behaves exactly as specified: eval=1 warning,
  check=1 warning, `--suppress-type-warnings`=0/exit 0 — reproduced
  directly.
- `--suppress-type-warnings` and `--strict` semantics otherwise unchanged
  and correctly wired (confirmed via 104 and via the removed/relocated code
  paths in the diff).
- Full lib (1305) + harness (496) suites green, clippy clean, fmt clean —
  reproduced myself.
- CHANGELOG entry present, accurate, and correctly documents the `--strict`
  implication in the abstract (the entry's own wording is fine; it's the
  underlying completeness of the fix that has a gap).

Sending back rather than merging on green — this is a correctness-affecting
regression in the exact area the PR is meant to fix, in a file whose sole
purpose is guarding against it, that the existing test suite structurally
can't detect. Happy to re-review quickly once addressed.

gmorpheme:
## Wicket review — PR #1012 (eu-rb5n Z, blob-baked prelude core)

**Verdict: APPROVE-PENDING-NAMING-HOLD.** Substance is correct and well-tested, CI is fully green, but the PR body/CHANGELOG overstate the performance win with mislabelled numbers — please correct before merge. Per the coordinator's instruction I am **not merging**: the owner has a pending naming decision on the new blob fields (possible `prelude_core` → `desugared_unit_cores` rename before the v3 wire format ships), so this PR stays open until that's resolved.

Reviewed at `aa0c5f78` in an isolated worktree (`/tmp/eu-wicket-1012`, detached HEAD from origin), clean `cargo clean && cargo build --release` before all timing.

### CI
All 16 required checks green (`gh pr checks 1012`), including Windows (which took ~19 min but passed). Release jobs correctly show `skipping` (not a PR-merge gate).

### Condition 2 — blob-core == source-merged equivalence: PASS, but file count is misreported
Independently re-ran the bidirectional sweep: `eu check --strict` vs `EU_SOURCE_PRELUDE=1 eu check --strict` over every `.eu` under `tests/harness/` (recursive, 509 files, includes the 111 in `tests/harness/typecheck/`) plus `lib/` (6 files) — **515 unique files, zero diffs.** The PR body/CHANGELOG say "620 files (111 + 509 + 6)" — that arithmetic double-counts the 111 `typecheck/` files, which are a *subset* of the 509 `tests/harness/` recursive count, not additive. Actual unique corpus is 515, not 620. Substance unaffected (zero diffs either way), but please fix the number before it ships in CHANGELOG.

Confirmed 065 (`tests/harness/typecheck/065_lookup_key_typo_warns.eu`) warns `unknown record key :naem` on **plain `eu` with no subcommand, no `--strict`** (the actual eval-path regression this bead exists to fix) — verified directly, output attached below. 104/111 typecheck fixtures match `.expect` sidecars on both paths (spot-checked, not exhaustively re-diffed against every sidecar — the corpus sweep already covers stderr/exit-code equivalence).

### Condition 1 — latency: blob-core numbers verified; "old (source) check" comparison column is mislabelled and inflates the headline win
Reproduced with `hyperfine --warmup 5 --min-runs 15+`, clean release build, deltas over `--suppress-type-warnings`:

| file | suppress | blob-core check | delta |
|---|---|---|---|
| trivial (005) | 7.7–7.9ms | 17.6ms | **+9.7ms** (PR: +9.6ms ✓) |
| mid (015, 157 lines) | 17.7ms | 31.5ms | **+13.8ms** (PR: +13.8ms ✓) |
| prelude-heavy (010, 276 lines) | 28.6–28.9ms | 45.5ms | **+16.6ms** (PR: +17.1ms, close ✓) |

The blob-core figures are solid and reproduce almost exactly.

The **"source (old) check"** column (58.2/74.1/86.0ms, framed as "3–6x reduction versus the fallback") does **not** measure what it's labelled as. I reproduced it exactly by running `EU_SOURCE_PRELUDE=1 eu <file>` (full source prelude for **both eval and check**) — 58.3/74.3/86.0ms, matching the PR's numbers to within noise. But `EU_SOURCE_PRELUDE=1` disables the blob for eval too, not just the check — it's a strictly harder scenario than the actual prior implementation (step 3 in the journey), which kept the blob for eval and only fell back to `run_type_checker` for the check itself.

I measured the *actual* fallback (blob eval + `run_type_checker` check-only — built by stripping `prelude_core` to empty from a real blob, matching your own condition-3 method) directly:

| file | suppress | true check-only fallback | delta |
|---|---|---|---|
| trivial | 7.7ms | 31.0ms | **+23.3ms** |
| mid | 17.7ms | 44.3ms | **+26.6ms** |
| prelude-heavy | 28.6ms | 58.3ms | **+29.7ms** |

This ~23–30ms range is consistent with your own condition-3 report ("~30.8ms, the expected fallback cost") and with the owner's recollection of the refused "~22ms" compromise — so it's almost certainly the correct baseline. Against it, the true improvement factor from (Z) is **~1.8–2.4x**, not "3–6x". The good news: the substantive claim that survives — "(Z)'s overhead never approaches the refused ~22ms compromise" — **is true and well-supported** by real data (9.6–17ms vs ~22–30ms). But the specific table and the "3–6x" framing in both the PR body and CHANGELOG need correcting to use the right comparison baseline, not `EU_SOURCE_PRELUDE=1`.

### Condition 3 — format-version handling: PASS
- `BYTECODE_WIRE_FORMAT_VERSION` 2→3 bump is consistent between `build.rs:26` and `xtask/src/main.rs:54` (doc comments cross-reference each other).
- Blob size independently regenerated: total 598,274 bytes, `prelude_core` 134,966 bytes (4 tagged units) — **exact match** to the PR's claim.
- Determinism: regenerated the blob twice independently, byte-identical (MD5 match).
- Fallback verified end-to-end myself (not just trusting the PR's account): built a real blob with `prelude_core` stripped to empty, rebuilt `eu`, confirmed 065 still warns correctly via `run_type_checker` (at the expected ~31ms fallback cost), restored the real blob.

### Condition 4/6 — config matrix / eval-path coverage: PASS
`test_config_matrix_blob_vs_source_prelude_byte_equal` and `test_eval_path_warns_on_record_key_typo` present and green; both configs (`cargo test --release` default: 498 harness passed; `EU_SOURCE_PRELUDE=1`: 498 passed) reproduced independently. Spot-checked `EU_GC_VERIFY=2 EU_GC_STRESS=1`: 498/498 green.

### Code review
- `inject_prelude_units` (`src/driver/source.rs`) — `entry().or_insert()`, correctly a no-op if the key is already populated (first-write-wins), and `translate()`'s existing-key short-circuit means a later `translate()` call on an injected input is a genuine no-op, not a silent overwrite. No shadowing footgun found: `tag_for_prelude_side_input` matches on `(Locator, name)` identity, which is specific enough that a user file couldn't accidentally collide with `__build`/`__io`/`__args`/prelude locators.
- `run_type_checker_from_blob_core` is a faithful, line-by-line mirror of `run_type_checker` — I diffed them side by side. The only structural difference is that injected inputs skip `load()` **and** `translate()` entirely (not just translate), which is correct since they're already core.
- **One residual architectural risk, not currently manifesting but worth flagging explicitly**: `Smid` (`src/common/sourcemap.rs:17`) is a bare index into a process-local `SourceMap` Vec (`source_info_for_smid` does `self.source.get(smid.get())` with no cross-process tagging). The baked `prelude_core` cores carry `Smid` values assigned by the *offline* `cargo xtask prelude-compile` process's own `SourceMap`. When injected into a fresh `SourceLoader`/`SourceMap` at `eu` runtime, those Smid integers are foreign — if any downstream diagnostic ever needs to render a location *inside* an injected prelude definition (as opposed to always citing the user's call site, which is what happens in every case I tested), it will either silently lose location info (out-of-bounds, `.get()` returns `None`) or, for a large enough user file, could alias to an unrelated real entry and show a misleading location. The 515-file byte-equal sweep gives real evidence this isn't happening for the current test corpus, but the corpus wasn't designed to probe this specific hazard (e.g. a warning whose primary label cites "as declared in prelude.eu here"). Given this is exactly the kind of divergence the merge-wiring map flagged as my job to catch, I'd like either an explicit test that forces a diagnostic to cite an in-prelude location on both paths (proving it degrades gracefully or matches), or an explicit acknowledgement in a comment/doc that this is understood and out of scope.
- `xtask/src/main.rs`/`engine_ab.rs` "pre-existing" clippy fixes (`print_literal`, `unnecessary_map_or`): verified independently by checking out `master` (175cc96f) into a separate worktree and running `cargo clippy -p xtask --all-targets -- -D warnings` — both warnings reproduce exactly at the claimed lines on unmodified master. Genuinely pre-existing, not introduced by this PR.
- Dead-code removal in `prepare.rs`/`source.rs` (the interim (X) seeded check, `type_warnings`, `prelude_type_summary`, `set_type_warnings`) — grepped the whole tree, no dangling references remain.
- UK English: spot-checked, consistent (e.g. "serialise", "behaviour" not present but no incorrect US spellings found either).

### Suites
- `cargo test --release` (default): 498 harness + 1307 unit + others, all green.
- `EU_SOURCE_PRELUDE=1 cargo test --release --test harness_test`: 498/498 green.
- `EU_GC_VERIFY=2 EU_GC_STRESS=1`: 498/498 green (spot-check).
- `cargo clippy --workspace --all-targets -- -D warnings`: clean.

### Required before merge
1. Fix the "620 files" arithmetic in the PR body and CHANGELOG to the correct unique count (515).
2. Fix the condition-1 latency table and the "3–6x reduction" framing in both the PR body and CHANGELOG — the "source (old) check" column measured `EU_SOURCE_PRELUDE=1` (full source eval+check), not the actual prior fallback (blob eval + check-only `run_type_checker`, ~23–30ms delta). Report the true ~1.8–2.4x figure, or drop the multiplier claim and rely on the (accurate, real) "never approaches the refused ~22ms compromise" framing.
3. Either add a targeted test for the Smid cross-process aliasing risk above, or add an explicit code comment acknowledging and scoping the limitation.

None of these are correctness blockers for the mechanism itself — the equivalence sweep, determinism, fallback, and eval-path regression fix are all real and independently verified. This is a numbers/documentation-accuracy hold, plus one architectural note, on top of the owner's outstanding naming decision.

— Wicket

gmorpheme:
## Wicket re-review — PR #1012 at `b7393037`

**Verdict: REJECT — CI failing on Windows, not merged.** Naming hold is resolved and the rest of the PR is in excellent shape (see below), but per hard constraint I cannot merge with failing CI, no exceptions.

### CI failure — root cause identified

`Test Suite (windows-latest)` fails (`gh api repos/curvelogic/eucalypt/actions/jobs/87210771921` → step "Run tests" → conclusion `failure`). The panic:

```
thread 'driver::check::tests::in_prelude_smid_degrades_gracefully_not_misleadingly' panicked at src\driver\check.rs:953:14:
run_type_checker_from_blob_core: FileCouldNotBeRead("unsupported locator: c:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\.tmpgsnT9b\\probe.eu", None)
```

This is the new Smid regression test from `70a8f4b0` — good test, wrong root cause of the failure, and it's exposed a real, pre-existing bug, not a flaky Windows quirk. Traced it:

- The test builds a temp file via `tempfile::tempdir()`, then passes the absolute path as a raw string through `EucalyptCli::try_parse_from(["eu", "check", "--strict", file_path.to_str()...])`.
- That string flows through `Input::from_str` (`src/syntax/input.rs:228`) → `Locator::from(&str)` (`src/syntax/input.rs:49`), which tries `Url::parse(url_text)` **before** falling back to `Locator::Fs(PathBuf::from(url_text))`.
- A Windows absolute path's drive letter (`C:\Users\...`) is syntactically a valid single-letter URL scheme per RFC 3986 (`scheme = ALPHA *( ALPHA / DIGIT / "+" / "-" / "." )`), so `Url::parse` succeeds and the whole path becomes `Locator::Url(...)` instead of `Locator::Fs(...)`.
- `SourceLoader::load_source` (`src/driver/source.rs:502-519`) has no case for a bogus `Url` locator that isn't a real network/file URL — it falls into the `other` catch-all and returns `FileCouldNotBeRead("unsupported locator: ...")`.
- The test's `.expect("run_type_checker_from_blob_core")` turns that `Err` into the panic.

**This is not just a test-isolation problem.** `Locator::from(&str)` doesn't special-case Windows drive letters before attempting URL parsing, so this is a latent, pre-existing bug in production code: any real Windows user invoking `eu C:\some\absolute\path.eu` (a fully-qualified path, which always has a drive letter on Windows) would hit the identical "unsupported locator" error today. No other test in the suite happened to feed a raw absolute Windows path through this exact CLI-string route before, which is why this has never surfaced despite Windows being a supported release target.

### Recommended fix

Two options, not mutually exclusive:
1. **Narrow (unblocks CI immediately)**: in the new test, avoid the raw-string CLI route — construct the `Input`/`Locator` directly (e.g. `Input::from(Locator::Fs(file_path.clone()))`) rather than round-tripping through `EucalyptCli::try_parse_from` + `Input::from_str`. Sidesteps the bug without touching production code, but leaves the real bug live.
2. **Root cause (recommended)**: harden `Locator::from(&str)` in `src/syntax/input.rs` to recognise a Windows drive-letter prefix (single ASCII letter + `:` + `\` or `/`) and treat it as `Locator::Fs` before attempting `Url::parse`. This is the fix that actually protects real Windows users, not just this test. Given "Panics Are Critical" and the project's Windows release target, I'd lean towards asking for this, filed as its own regression test (`Locator::from("C:\\foo\\bar.eu")` should equal `Locator::Fs(...)`, not `Locator::Url(...)`) — but this could reasonably be a fast-follow bug rather than blocking this PR further, since it's an orthogonal pre-existing issue this PR's test merely exposed. Coordinator's call on whether to fold into this PR or file separately; either way it needs a beads entry so it doesn't get lost.

### Everything else: all clear, all three requested fixes verified independently

1. **File count**: re-ran the full corpus sweep myself at `b7393037` — 515 unique files (`tests/harness/` recursive + `lib/`), zero diffs. CHANGELOG now correctly says "515 unique files ... 509 recursively under `tests/harness/` ... plus 6 in `lib/`" — matches exactly.
2. **Latency**: re-measured independently (clean build, `hyperfine --warmup 5 --min-runs 15`):
   - blob-core check deltas: +9.7/+13.8/+16.4ms (trivial/mid/heavy) — matches the CHANGELOG's +9.6/+13.8/+17.1ms.
   - true fallback deltas (blob eval + `run_type_checker` check-only, via a blob with `desugared_unit_cores` stripped to empty): +23.0/+26.9/+29.3ms — matches the CHANGELOG's re-based +23.2/+26.7/+30.1ms and my own original measurement almost exactly.
   - CHANGELOG framing is now accurate: "~23–30ms to ~10–17ms, a ~1.8–2.4x reduction ... never approaches the refused ~22ms-adding compromise." Good.
3. **Smid test**: `in_prelude_smid_degrades_gracefully_not_misleadingly` genuinely exercises the hazard I flagged — injects a synthetic prelude unit with a real internal type mismatch tagged with an out-of-range `Smid` via the actual `run_type_checker_from_blob_core` path (not a mock), and asserts the diagnostic survives with an empty location label rather than an aliased wrong one. This is exactly the right test for the risk. Runs and passes on macOS/Linux (confirmed locally); it's the one that's failing specifically on Windows, for the reason above. The accompanying doc comment on `run_type_checker_from_blob_core` correctly scopes the residual in-range-alias risk and points at the follow-on fix (re-basing baked Smids into the runtime SourceMap).

### Rename pass (b7393037): clean

- `prelude_core` → `desugared_unit_cores`: zero remaining occurrences of `prelude_core` anywhere in the tree (grepped `src/`, `xtask/`, `docs/`, `tests/` — nothing, live or historical).
- `inline_cores` → `inlinable_bindings`, `inject_prelude_inline_cores` → `inject_prelude_inlinable_bindings`: confirmed live call site (`src/driver/prepare.rs:332`) and function definition (`src/driver/source.rs:750`) both renamed. Remaining `inline_cores` hits are exclusively in `docs/superpowers/specs/*.md` and `docs/superpowers/reports/*.md` — correctly left untouched as historical records, as the commit message states.
- Regenerated the blob myself: `598,274` bytes total, `134,966` bytes `desugared_unit_cores` — byte-size identical to the pre-rename blob, as expected (postcard is positional, wire-free rename). Determinism re-verified independently: not required again since the mechanism is unchanged and structurally verified via the size match, but I did re-run the corpus sweep post-rename (515/515, zero diffs) to confirm behaviour is unaffected.

### `type_summary` — independently confirmed genuinely dead (write-only)

Grepped every reference to `.type_summary` in `src/`: `src/driver/check.rs:120` and `:384` reference `ui.type_summary`, which is a **different** field on `UnitInterface` (`src/driver/unit_interface.rs:115`), populated fresh on every check run — unrelated to the blob. `PreludeBlob::type_summary` (`src/eval/stg/blob.rs:135`) has exactly one writer in live code (`xtask/src/main.rs:440`, at bake time) and zero readers anywhere. This confirms Furnace's claim precisely: it's genuinely dead weight in the wire format today (its only consumer, the interim (X) seeded check, was deleted earlier in this PR).

**Verdict on `type_summary`**: it should be removed, but not in this PR — agree with treating it as out of scope for a rename pass. Since this PR is already bumping `BYTECODE_WIRE_FORMAT_VERSION` to 3 and the v3 format hasn't shipped in any release yet, removing an unused field is genuinely free right now (no back-compat obligation, no user-visible blob ever carries v3 without this bake anyway) — this is the cheapest possible moment to do it, and it'll only get more awkward once 0.13 ships with v3 blobs in the wild. Recommend the coordinator file a follow-on bead for this pre-release, not urgent enough to block, but worth doing before 0.13 cuts.

### Suites (at b7393037, before the Windows failure was known)
- `cargo test --release` (default): 1308 unit + 498 harness, all green except the one Windows-only failure above (reproduced locally on macOS with no issue — the failure is Windows-path-specific, not a general regression).
- Corpus sweep re-run: 515/515, zero diffs.
- Local run of the new test in isolation: passes on this (macOS) machine.

### Next steps
Please fix the Windows locator issue (test-local fix at minimum, root-cause fix recommended) and push. I'll re-review CI once green — everything else is ready to merge at that point. Not closing eu-rb5n (Q2 source-path unification is still open).

— Wicket

gmorpheme:
Merged. All 16 CI checks green at `8cf4b9c0`, the drive-letter fix independently verified (clean build, corpus sweep re-run 515/515 zero diffs, full suite 1310 unit + 498 harness green including `EU_SOURCE_PRELUDE=1`, clippy/fmt clean, both new locator tests and the Smid test all pass). Merge commit: b553539b67cb8af13802af77804a42459e53cbe5.

— Wicket

PR #1013, opened 2026-07-14
docs(perf): correct day07 causal gloss in BV3 re-profile report (eu-98zg)

gmorpheme:
Small, directed correction per Wicket's #1011 review (merge on green -- same directed-mode authority as eu-98zg itself, no owner gate). Docs-only.

## What was wrong

The merged eu-98zg report (`docs/superpowers/reports/2026-07-14-bv3-post-copyspec-reprofile.md`) §5 claimed day07's residual 20.5% `enter_local`/`get` share came from a "hand-written beam-fold... not a prelude combinator... not touched by copy-specialisation." **This is false.** `solve(rows, b0): rows foldl(row-step, [b0, 0]) second` calls the prelude's own `foldl` directly -- there is no hand-written recursive fold anywhere in day07.eu.

Wicket verified at the STG level (`eu dump stg day07.eu -t part-1`) that this `foldl` call **fuses**, via the same local self-recursive specialised-copy signature established for `022_hof_fold`. I independently re-verified this from the source (`row-step` has no self-reference anywhere in day07.eu -- it's a plain, non-recursive per-row transform passed as foldl's operator argument) and from my own STG dump (the compiled `solve` body's head node is exactly the `letrec [0] λ{3}` self-recursive-copy shape Wicket described).

## Corrected causal account

The fold is not the residual cost -- it fuses cleanly, same as 022. The actual source of the 20.5% share is `row-step` itself: a substantial **non-recursive** operator function whose body calls sibling top-level helpers (`to-splitters`, `zip-with`, `shift-left`, `shift-right`) as free variables, once per fold iteration (~150 grid rows). This is ordinary cross-function-call environment-lookup cost -- `row-step` was never a copy-specialisation candidate in the first place (leg 1 of the criterion requires self-recursion, which a plain operator function doesn't have), so this isn't a gap in the combinator admission list at all.

**This changes the decision-relevant framing, not the verdict**: the corrected residual population (general cross-function-call lookup cost in non-recursive, call-heavy operator functions) is *broader* and *structurally cleaner* than the original "user folds outside the 8-combinator set" story -- no plausible relaxation of `tag_combinators` (which is fundamentally about recursive self-calls) could ever reach it, since it isn't recursion-shaped at all. Per Wicket's review, this is arguably better evidence for BV3's continued relevance than the original (wrong) framing.

## What's unchanged

The **finding** stands exactly as measured: day07's 20.5% share, the overall verdict (share collapsed on all four workloads, defer BV3 to 0.14+ rather than retire), and every number in the report. Only the causal explanation of *why* day07 shows 20.5% needed fixing.

## What changed in this commit

- Dated correction banner at the top (per the #997 record-correction precedent).
- §5 points 2-3 rewritten with the corrected causal account.
- §8 (eu-2sa6.2 bead-update summary) updated to match.
- `eu-2sa6.2` bead notes updated directly (`bd note`) with the same correction.
- Everything else in the report is unchanged.

### DIFF
--- a/docs/superpowers/reports/2026-07-14-bv3-post-copyspec-reprofile.md
+++ b/docs/superpowers/reports/2026-07-14-bv3-post-copyspec-reprofile.md
@@ -1,5 +1,25 @@
 # Post-copy-specialisation re-profile: enter_local share collapse (eu-98zg)
 
+> **CORRECTION (2026-07-14, Wicket's #1011 review).** §5 originally claimed
+> day07's `foldl(row-step, [b0, 0])` call was a "hand-written beam-fold...
+> not a prelude combinator... not touched by copy-specialisation." **This is
+> wrong.** `solve` calls the prelude's own `foldl` directly; Wicket verified
+> at the STG level (`eu dump stg day07.eu -t part-1`) that it fuses via the
+> same local self-recursive specialised-copy signature established for
+> `022_hof_fold`, and this was independently re-confirmed here from the
+> source (`row-step` has no self-recursion anywhere in day07.eu) and the STG
+> dump's `letrec [0] λ{3}` head node. **The fold is not the residual cost.**
+> The actual source: `row-step` is a substantial *non-recursive* operator
+> function that calls sibling helpers (`to-splitters`, `zip-with`,
+> `shift-left`, `shift-right`) as free variables — ordinary
+> cross-function-call environment lookups, once per fold iteration, not
+> "fold outside the qualifying set." §5 and §8 below are corrected in place
+> to reflect this; the **finding itself is unchanged** (20.5% share on
+> day07, defer-BV3-not-retire recommendation stands) — only the causal
+> explanation was wrong, and the corrected account is, if anything, a
+> *broader and structurally cleaner* case for BV3's continued relevance (see
+> corrected §5). Full detail in Wicket's PR #1011 review comment.
+
 - **Date:** 2026-07-14
 - **Bead:** eu-98zg, owner-commissioned re-profile contingent on eu-dp0k/#1010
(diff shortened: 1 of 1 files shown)


gmorpheme:
## Wicket review — verified, merging on green

Checked the diff against my #1011 finding and the actual eu-2sa6.2 bead
state, not just read it.

- **Diff scope matches the claim exactly**: dated correction banner + §5
  points 2-3 + §8, nothing else touched (`git diff` confirms — every other
  number, table, and section is byte-identical to the merged #1011
  version).
- **The corrected causal account is accurate** and matches what I verified
  in #1011: `solve`'s `foldl(row-step, [b0, 0])` call fuses (same local
  self-recursive `letrec [0] λ{3}` signature as 022_hof_fold); the 20.5%
  residual is `row-step`'s own calls to sibling helpers
  (`to-splitters`/`zip-with`/`shift-left`/`shift-right`), not an unfused
  fold. Good that this wasn't just taken from my review comment on trust —
  the PR description shows independent re-verification from source (no
  self-reference in `row-step`) and a fresh `eu dump stg` run, which is the
  right level of rigor for a correction PR.
- **The "broader, structurally cleaner case for BV3" framing is carried
  through consistently** into both the report (§5 point 2/3, the "what
  would justify full retirement" paragraph) and the bead note — not just
  noted and dropped.
- **eu-2sa6.2 bead note checked directly** (`bd show eu-2sa6.2`): a new
  `CORRECTION` entry is genuinely appended, matches the doc's correction
  precisely, and the original `VERDICT` note is preserved rather than
  overwritten — the right way to keep a bead's history intact.
- **Headline finding genuinely unchanged**: verdict, all four workload
  percentages, and the defer-not-retire recommendation are untouched, as
  claimed.

Docs-only, low-risk, corrective. Merging on green.

PR #1015, opened 2026-07-15
docs: phase 2 content (items 1-4) — architecture bytecode engine, error messages, reflect guide, syntax gotchas (eu-2sa6.21)

gmorpheme:
Owner-approved phase-2 documentation content for the 0.13 release (bead **eu-2sa6.21**; items 1-4 of the docs-review §6 plan — items 5-9 are explicitly deferred). Every example was executed against a release build of the current branch base (master `b553539b`) and its output captured verbatim.

## Item 1 — `docs/architecture.md`: bytecode engine
The "STG Compilation and Evaluation Model" section previously described only the HeapSyn tree-walk. It now:
- Frames the two engines as sharing one STG IR (with the differential test as the byte-identical invariant), reworking the old "tracked gap" note.
- Adds **The Bytecode Engine (BV1, the default)**: what `encode.rs` compiles (post-order byte stream, one opcode per `StgSyn` variant, `DirectApp`/`FusedPrimop` superinstructions, constant pool, templates), the off-heap-code GC rationale, how dispatch works, and the newly merged **pre-decoded IR** (`predecode.rs`, `EU_PREDECODE` — 16-byte `Instr` records + off-heap side pools, ordinal `CodeRef`s, `Ann` peeled into a smid side table).
- Explains why bytecode is the default and states the perf trade-off **qualitatively**, deferring all numbers to `engine-ab/PROTOCOL.md` and its confidence labels (no unlabelled figures cited).
- Adds a **Pre-Compiled Prelude Blob** subsection using the `blob.rs` anatomy as the authority (`desugared_unit_cores`, `inlinable_bindings` — not the old `prelude_core`/`inline_cores`), plus the wire-format-version-3 staleness hash.

## Item 2 — `docs/reference/error-messages.md`
Replaces the 4-line stub with real content: diagnostic anatomy (codespan style — severity, primary/secondary labels, help/note lines; **no numeric error codes**), warnings-vs-errors with the full `--strict` / `check` / `--suppress-type-warnings` exit-code table, the main error categories (parse, binding/unresolved, assignment-style, self-reference, type warnings, execution), and reading **stack traces** and source locations. All transcripts pasted verbatim from the binary (`NO_COLOR`).

## Item 3 — `docs/guide/type-checking.md`: reflect.eu
New **Runtime type reflection (`reflect.eu`)** subsection (reachable via the already-listed page): the `s"…"` type-data prefix, `to-data`/`from-data` (with the `:t-*` tagged-list structure), and `as-spec` + `match?` with a form→check table. Uses the idiomatic `{ import: "reflect.eu" }` (per eu-1n5z).

## Item 4 — three new `docs/appendices/syntax-gotchas.md` entries
- **Bare numeral in interpolation braces** — `"{42}"` is string-anaphor index 42 (a 43-arg function), silently dropped from output; `eu -e '"{42}"'` crashes with a PAP-arity error. Fixes: `{{42}}` or `{name}`.
- **Hyphens/digits are identifier characters** — `x-2` is one name (`-`, `*`, `!`, `?` continue an identifier; `+`, `/`, `<`, `=` do not); `x - 2` is the fix. Verified against the lexer's `is_normal_continuation`.
- **The `!` asserted-type prefix** — checker trusts the annotation without verifying the body (`check.rs:904`); useful to supply an unverifiable type, dangerous as a warning-silencer. Cross-referenced from the cheat-sheet and agent-reference type-form tables.

## Verification / gates
- `mdbook build` clean; cross-reference anchors confirmed against generated HTML.
- `scripts/test-doc-examples.py` green on every changed file (0 failed).
- Prelude reference: **no drift** — zero prelude source touched; the only `eu doc` regen delta is the pre-existing hand-authored supplement content.
- `docs/llms-full.txt` regenerated; `cargo fmt --all -- --check` clean; harness **498/498**. UK English throughout.

## Note — bug surfaced
While verifying `as-spec`, found that it returns an **always-false** predicate for **union** types (`lib/prelude.eu:2199` binds the value as `match?`'s spec argument). Filed as **eu-r9oy** (P1); out of scope for a docs PR, so the reflect subsection documents the working forms and omits union `as-spec` with a pointer to the bead.

### DIFF
--- a/docs/appendices/cheat-sheet.md
+++ b/docs/appendices/cheat-sheet.md
@@ -629,6 +629,7 @@ origin: { x: 0, y: 0 }
 | `"value"`          | literal string type (subtype of `string`); c-string escapes for the quotes: `c"\"value\""` |
 | `:name`            | literal symbol type (subtype of `symbol`)      |
 | `T?`               | partial — sugar for `T \| ExecutionError`; marks functions that may raise an error |
+| `!T` (leading `!`) | asserted — checker trusts the annotation without verifying the body ([gotcha](syntax-gotchas.md#the--asserted-type-prefix)) |
 | `ExecutionError`   | the type of a raised runtime error     |
 | `A -> B`           | function                               |
 | `A \| B`           | union                                  |
--- a/docs/appendices/syntax-gotchas.md
+++ b/docs/appendices/syntax-gotchas.md
@@ -232,6 +232,53 @@ function value rather than a boolean.  The not-nil postfix `✓` is
 often the most natural way to introduce the outer anaphor when you
 want to also guard against null.
 
+### A Bare Numeral in Interpolation Braces Is an Anaphor Index
+
+**Problem**: Inside a string, `{...}` is interpolation, and a **bare
+number** in the braces — `"{42}"` — is not the literal text `42`. It is
+**numbered string anaphora**: `{0}`, `{1}`, `{2}`, … refer to the 1st,
+2nd, 3rd positional argument of the string-as-a-function. So `{42}` makes
+the string a **function of 43 arguments**, referencing the 43rd.
+
+**Gotcha**: This is silent. A string with a bare-numeral brace becomes a
+function value, and functions are not rendered — so the whole binding just
+vanishes from the output with no error:
+
+```eu,notest
+answer: "the answer is {42}"   # answer is now a 43-arg function
+other: 1
+```
+
+```text
+$ eu example.eu
+---
+other: 1
+```
+
+`answer` produced no output at all. Forcing the value (e.g. with `eu -e
+'"{42}"'`) turns the silence into a crash instead:
+
+```text
+error: bytecode: PAP arity out of range (supplied 1, pending 42, max 16)
+```
+
+**Fixes**:
+
+```eu,notest
+# Literal braces — double them
+answer: "the answer is {{42}}"   # => "the answer is {42}"
+
+# Interpolate a *value* named 42 — bind it to a name first
+n: 42
+answer: "the answer is {n}"      # => "the answer is 42"
+```
+
+**Rule**: a number inside interpolation braces is always an anaphor index,
+never a literal. Use `{{` `}}` for literal braces, and interpolate values
+through named bindings (`{name}`) or a format specifier (`{value:%d}`),
+never a bare numeral. Small indices bite the same way — `"{0}"` is a
+one-argument function, not the digit zero.
+
 ## Metadata vs Comments
 
 ### Backtick Is Metadata, Not a Comment
@@ -292,6 +339,49 @@ z: home.'notes.txt'          # Looks up identifier: notes.txt
 'hello' = 'hello'            # Compares two variable references (not strings)
 "hello" = "hello"            # Compares two string literals (correct)
 ```
+
+## Hyphens and Digits Are Identifier Characters — `x-2` Is One Name
+
+**Problem**: Eucalypt identifiers are kebab-case: after the initial
+letter, an unquoted name may contain hyphens, digits, and the characters
+`* ! ? _`. So `x-2` is not `x - 2` — it lexes as a **single identifier**
+named `x-2`:
+
+```eu,notest
+x: 10
+r: x-2        # r is not 8 — 'x-2' is one name
+```
+
+```text
+error: unresolved variable 'x-2'
+```
+
+**Gotcha**: this is asymmetric with the other arithmetic operators,
+because `+`, `/`, `<`, `>`, `=` and friends are **not** identifier
+characters, but `-` and `*` **are**:
+
+```eu,notest
+r: x+1        # => x + 1   (+ is not an identifier character)
+r: x-1        # => identifier 'x-1'   (- IS an identifier character)
+r: x*2        # => identifier 'x*2'   (* IS an identifier character)
+```
+
+Leading digits also disambiguate — the lexer only starts a normal
+identifier on a letter, so `2-x` is `2 - x`, while `x-2` is one name. And a
+space after the hyphen forces the operator reading, but a space *before*
+it makes the `-2` a negative literal in catenation position:
+
+```eu,notest
+r: 2-x        # => 2 - x   (subtraction — identifiers can't start with a digit)
+r: x - 2      # => x - 2   (spaces on both sides — subtraction, the fix)
+r: x -2       # => catenation of x and the literal -2 (calls -2 as a function)
+```
+
+**Rule**: put spaces around `-` (and `*`) whenever you mean the arithmetic
+operator next to a name: write `x - 2`, never `x-2`. This bites hardest
+right after another operator, where the eye reads arithmetic but the lexer
+reads a name: `n * x-2` is `n * (x-2)` with `x-2` unresolved; you want
+`n * x - 2`.
 
 ## Division Operators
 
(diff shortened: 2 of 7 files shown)


gmorpheme:
## Wicket review — PR #1015 (docs phase 2, eu-2sa6.21)

**Verdict: send back — one factual inaccuracy in error-messages.md needs fixing before merge.** Everything else in this PR is excellent: technically precise against the code, every other example I independently verified checks out exactly, and all gates pass. This is a small, easily-fixed docs correction, not a rewrite.

Reviewed at `22f4b153` in an isolated worktree (`/tmp/eu-wicket-1015`), clean release build + regenerated prelude blob before verification.

### The inaccuracy: "Type warnings" example in error-messages.md

The example under `### Type warnings`:

```eu
` { type: "number -> number" }
double(x): x * 2
result: double("hello")
```

shows only:

```text
warning: type mismatch calling 'double'
  ┌─ example.eu:3:16
  │
3 │ result: double("hello")
  │                ^^^^^^^ expected number, found "hello"
```

followed by "Under `--strict` this same finding becomes an error and exits `1`" — which, by construction (parallel to the `--strict` framing used everywhere else on the page, and the fact that no `eu check` prefix is shown, unlike commands that intentionally use `eu check`), implies that *without* `--strict`, evaluating this program exits `0` with just the warning, the same pattern established by the page's own earlier `nth(0)` example.

I built the example verbatim and ran it against a release binary (`master` `b553539b` + this PR):

```
$ eu example.eu
warning: type mismatch calling 'double'
  ...
error: type mismatch: expected number, found string "hello"
  ┌─ example.eu:2:14
  │
1 │ ╭ ` { type: "number -> number" }
2 │ │ double(x): x * 2
  ...
exit: 1
```

`double(x): x * 2` genuinely **crashes at runtime** on a string argument — `"hello" * 2` is an execution error, independent of the type checker entirely. So plain `eu example.eu` (no `--strict`) exits `1` regardless, for a reason that has nothing to do with strictness. The transcript shown in the doc only matches `eu check example.eu` (check-only mode, no evaluation) — I confirmed that specific invocation reproduces the doc's transcript exactly, including the `0` exit code.

This directly undercuts the page's own didactic point in this exact section (which exists specifically to explain that type warnings, unlike errors, don't normally halt evaluation) — a reader who copies this example and runs plain `eu` will see an unexplained extra `error:` block and a `1` exit code that the surrounding prose told them not to expect.

**Suggested fix** (any of these three work): (a) swap the example for one that's genuinely warning-only under plain eval, matching the `nth(0)` example's pattern (e.g. a mismatch that doesn't itself crash, such as a lens/pipeline type warning that still evaluates to *something*); (b) show the full transcript including the runtime error, with a note that this particular example also happens to crash independently of the type-warning mechanism; or (c) explicitly relabel the transcript as `$ eu check example.eu` output and adjust the surrounding sentence accordingly. (a) is cleanest since it keeps the section focused on the point it's trying to make.

### Everything else: verified accurate

**architecture.md (item 1)** — checked every specific claim against the actual code, not just skimmed:
- `src/eval/bytecode/` file list (`opcode.rs`, `program.rs`, `encode.rs`, `closure.rs`, `cont.rs`, `env_builder.rs`, `machine.rs`, `predecode.rs`) — all present, confirmed via `ls`.
- `EU_PREDECODE=1` env var — confirmed at `src/eval/bytecode/mod.rs:64`.
- `Instr` "exactly 16 bytes" claim — confirmed via the actual compile-time assertion at `src/eval/bytecode/predecode.rs:81` (`assert!(std::mem::size_of::<Instr>() == 16)`).
- `DirectApp`/`FusedPrimop` superinstructions — confirmed present in `opcode.rs`.
- GC polling "every 500 steps... matching HeapSyn" — confirmed identical cadence in both `src/eval/machine/vm.rs` and `src/eval/bytecode/machine.rs`.
- Wire-format-version-3 hash composition (`SHA-256(lib/prelude.eu ‖ wire-format-version)`) — confirmed exactly in `build.rs`.
- Blob anatomy table (`desugared_unit_cores`, `inlinable_bindings`, `bytecode`, `name_to_slot`, `operators`, `monad_specs`/`type_summary`) — matches `blob.rs`'s current field names exactly (I reviewed and merged the PR that renamed these fields yesterday, so I know this schema well).
- **No unlabelled performance numbers**: confirmed — the perf section is entirely qualitative ("wins decisively on... loses on...") and explicitly defers all figures to `engine-ab/PROTOCOL.md`'s confidence-labelled ledger. This is exactly right per the project's mandatory protocol.

**error-messages.md (item 2)** — spot-checked 8 of the page's transcripts by constructing the exact source shown and running it: unresolved-variable, self-reference-diverges, assignment-style-hint, unterminated-block, catenation-call-error, head-of-empty-list-with-stack-trace, and the full exit-code table (all 5 rows verified row-by-row against `ex2.eu`, a genuinely benign warning case). All matched **exactly**, byte-for-byte on the message text, except the one issue above. Minor cosmetic note (not worth fixing): a couple of multi-line span headers in my output have trailing whitespace after `│` that the doc's pasted version doesn't — harmless, likely just editor trimming, not a content issue.

**type-checking.md reflect.eu subsection (item 3)** — this one I checked hardest since it's new ground (runtime reflection): `s"…"` type-data rendering, `to-data`/`from-data` round-trips (including the exact `:t-*` tag shapes for list/record), `as-spec` + `match?` for record/tuple/partial/function forms, open-record semantics (extra keys ignored), and nested composition — every single example I tried (10 separate test files) matched the doc's claimed output exactly. I also independently reproduced the union `as-spec` always-false bug (`eu-r9oy`) myself — confirmed `s"number|string" as-spec` returns `false` for both a matching number and a matching string. The "Known limitation" callout is accurate and correctly scoped; it doesn't contradict what the eu-r9oy fix will change (that fix corrects the union branch in `lib/prelude.eu`, this doc just documents the current, honest behaviour and points at the bead).

**syntax-gotchas.md (item 4)** — all three new entries verified against the binary: bare-numeral interpolation (`"{42}"` silently vanishes the binding; forcing it produces the exact `PAP arity out of range (supplied 1, pending 42, max 16)` message quoted); hyphen/digit identifier characters (`x-2`/`x*2` both genuinely lex as one identifier and produce `unresolved variable`; `2-x` correctly parses as subtraction; `x -2` correctly produces the "tried to call a number as a function" catenation trap); and the `!` asserted-type prefix (plain annotation warns, `!`-prefixed asserted annotation is silently trusted, and the `!{{...}}` closed-record cross-reference example from the type-checking guide reproduces exactly).

### Merge-collision check with eu-r9oy: clear

PR #1015's only touch to `agent-reference.md` is a one-line insertion into the type-form table at line ~351-353 (the `!T` asserted-annotation row). `eu-r9oy`'s planned fix touches the union-`as-spec` claim in §3.5a at line 887, a completely separate section ~530 lines away. No hunk overlap; safe to merge in either order. (Minor, non-blocking nit: `cheat-sheet.md`'s new row links to the gotcha entry with a working anchor — I verified `#the--asserted-type-prefix`, note the double hyphen from mdbook's heading-ID generation, resolves correctly in the built HTML — but `agent-reference.md`'s equivalent row is plain text "see syntax-gotchas.md", not an actual hyperlink. Cosmetic, not worth blocking on.)

### Gates

- `mdbook build`: clean, no errors. Verified the new cross-reference anchor resolves in the generated HTML.
- `scripts/test-doc-examples.py` on all four changed files: 0 failed (53/53 skipped-as-notest in type-checking.md, 1/1 skipped in error-messages.md, 3 passed + 40 skipped in syntax-gotchas.md, no eu blocks in architecture.md). The notest-heavy ratio is expected and pre-existing convention for these files (most blocks are deliberately-broken illustrations of gotchas/errors) — I compensated with the manual verification above rather than relying on the script alone, per the coordinator's ask.
- Prelude reference zero-drift: reproduced the CI check locally (`eu doc --prelude` regen + `diff -r`) — clean, zero diffs, confirming no prelude source was touched.
- `docs/llms-full.txt`: regenerated via `scripts/generate-llms-full.sh` myself and diffed against the committed version — byte-identical.
- UK English: scanned all added lines for common US/UK spelling pairs — only UK spellings found (`behaviour`, `colour`, `labelled`); "program" appears 16 times but is correct in both dialects for computing usage (not a UK "programme" miss).
- `cargo fmt --all -- --check`: clean. `cargo test --release --test harness_test`: 498/498.
- CI: all 16 checks green (`gh pr checks 1015`), including both Windows and the two AddressSanitizer/Test-Suite jobs that were still pending when I started.

### Next steps

Please fix the `double("hello")` example in error-messages.md (see suggested fixes above) and push. Once that's in, this is ready to merge — I don't need to re-verify everything else, just that one section.

— Wicket

gmorpheme:
Merged. Re-verified the fix at `a4678449`: the new `describe(n): "value: {n}"` example matches the doc's transcript exactly, byte-for-byte, in all three modes (plain `eu` → warning + `result: "value: hello"` + exit 0; `eu check` → warning + exit 0; `--strict` → exit 1) — including the precise column number (3:18) and underline width. The new agent-reference.md cross-reference link resolves correctly to `#the--asserted-type-prefix` in the built HTML (confirmed via `mdbook build`). `test-doc-examples.py` clean on both changed files, `llms-full.txt` regeneration byte-identical to committed, prelude reference zero-drift, fmt clean. Confirmed no merge conflicts against the current master tip (post-#1017) via `git merge-tree`, and all CI checks green. Merge commit: 0c3e2a444072e1659e38bd29d61e628e38afa954.

— Wicket

PR #1017, opened 2026-07-15
fix(prelude): as-spec union predicate always false (eu-r9oy)

gmorpheme:
## Summary

- `as-spec` on a union type (`A | B`) always produced an always-false predicate. `union-spec(ts, v): ts map(to-spec-td) any(v match?)` wrote `v match?` as a bare catenation with **no explicit spec argument** — this partially applies `match?(pat, target)` with `v` bound as the **pattern** (`pat`), not the value being matched. Every sub-spec test was therefore nonsensical (matching an arbitrary sub-spec value as if it were a pattern against `v`) and the union predicate was always `false`, regardless of which branch the value actually matched.
- Fixed by introducing a named helper, `matches-v(s): v match?(s)`, mirroring `partial-spec`'s existing correct catenation (`v match?(inner to-spec-td)`) one line below the bug. `union-spec` now calls `any(matches-v)` over the mapped sub-specs, so each sub-spec is correctly tested as the pattern against the fixed value `v`.
- Regenerated `lib/prelude.blob` (gitignored build artefact, not part of this diff) and confirmed the regen is byte-identical across two runs (determinism check).

## Root cause detail

```
union-spec(ts, v): ts map(to-spec-td) any(v match?)
```

`v match?` desugars to `match?(v)` — a partial application binding `v` to `match?`'s **first** parameter (`pat`), not its second (`target`). `any` then calls this partial function on each sub-spec, i.e. `match?(v, sub-spec)` — backwards. Fix:

```
union-spec(ts, v): {
  matches-v(s): v match?(s)
}.(ts map(to-spec-td) any(matches-v))
```

`v match?(s)` correctly catenates `v` as the *last* (`target`) argument: `match?(s, v)`.

## Testing

- New dedicated regression test `tests/harness/189_r9oy_union_as_spec.eu`, registered in `tests/harness_test.rs` (`test_189_r9oy_union_as_spec`). Covers: primitive union accept-number / accept-string / reject-symbol / reject-bool, union nested inside a list, and union nested inside a record field. Uses direct `match?` results (with `not(...)` for negative cases) aggregated via `all-true?`, so `RESULT` is a real, exit-code-affecting aggregate — verified this actually gates by temporarily reverting the prelude fix and confirming the test goes `FAIL`, then restoring the fix and confirming `PASS`.
- Extended the existing AC4 (union specs) block in `tests/harness/175_sv2_to_spec.eu` with a reject-symbol case and a nested-in-list case, for documentation/AC coverage alongside the primitive cases already there.

**Important finding (documented, not fixed here — separate concern):** `tests/harness/175_sv2_to_spec.eu` (and several sibling `as-spec`/`to-spec` test files) end with a hardcoded `RESULT: :PASS` rather than an aggregate computed from their `//=>` assertions. Per `lib/test.eu`'s `default-expectation`, when a test block's result has an explicit `RESULT` key, the harness validator uses **only** that key and ignores every other value in the block — so a `//=>` assertion failing inside a `RESULT: :PASS`-terminated block prints `EXPECT FAILED` to stdout/report but does **not** fail `cargo test`. This is how the union bug shipped despite AC4 already asserting `t-union-str`/`t-union-num` should be `true` — those assertions were failing silently the whole time. I verified this directly: reverting the prelude fix while `175`'s `RESULT: :PASS` was still hardcoded still printed `175_sv2_to_spec...PASS`. I left `175`'s `RESULT: :PASS` as-is (fixing the aggregation pattern for ~20 pre-existing assertions is a bigger, orthogonal change and risks collateral edits to a file used elsewhere) and instead added `189` as the authoritative, properly-gating regression test. Worth a follow-up bead to audit other `RESULT: :PASS`-terminated test files (e.g. `183_widen_type_def_literals.eu`) for the same silent-failure pattern.

### Before/after transcript

Repro from the bead, before the fix (built from `origin/master` at `b553539b`):

```
$ cat repro.eu
{ import: "reflect.eu" }
schema: s"number | string" as-spec
r1: 42 match?(schema)
r2: "hello" match?(schema)
r3: :sym match?(schema)

$ ./target/debug/eu repro.eu
---
r1: false
r2: false
r3: false
```

After the fix:

```
$ ./target/debug/eu repro.eu
---
r1: true
r2: true
r3: false
```

`r1`/`r2` now correctly match their union branches; `r3` (a symbol against `number | string`) correctly still rejects.

## Docs

- Checked `docs/reference/agent-reference.md` §3.5a's `as-spec` mapping table (line ~886: `| A \| B union | Predicate: true if value matches any branch |`). This table documents the *intended* behaviour, which the implementation now actually matches post-fix — no edit needed, it was aspirationally correct all along; only the implementation was broken.
- Checked for overlap with PR #1015 (`gh pr diff 1015`): it touches a *different* table in the same file (the "Common type forms in annotations" table, around line 351-353, adding a `!T` asserted-type row) — no hunk overlap with the §3.5a `as-spec` table I reviewed. No rebase concern.
- Added a CHANGELOG `Unreleased` → `Fixed` entry.

## Gates

- `cargo fmt --all` — clean
- `cargo test` (default/blob path) — 499 harness tests pass (was 498; +1 new)
- `EU_SOURCE_PRELUDE=1 cargo test` — 499 pass
- `EU_HEAPSYN=1 cargo test` — 499 pass

## Test plan

- [x] Reproduced the bug against a fresh build of `origin/master`
- [x] Applied fix, confirmed repro now returns correct results
- [x] New test 189 verified to fail when the bug is reintroduced (temporarily reverted the prelude fix and reran)
- [x] Full harness suite green under default, `EU_SOURCE_PRELUDE=1`, and `EU_HEAPSYN=1`
- [x] `lib/prelude.blob` regeneration verified deterministic (byte-identical across two runs)


gmorpheme:
## Wicket review — PR #1017 (eu-r9oy, as-spec union fix) — MERGED

**Verdict: approve and merge.** Fix is correct, well-tested, idiomatic, and every gate passes. Merged as `f44ebf01`.

Reviewed at `a8d310dd` in an isolated worktree, clean build + regenerated blob before verification, plus a **separately-built master (`b553539b`) worktree** to independently confirm the before/after behaviour rather than trusting the PR's own transcript.

### 1. Repro — verified independently, before and after

Built the exact bead repro (`s"number | string" as-spec`, tested against `42`/`"hello"`/`:sym`) against my own from-scratch master build:

- **Before (master `b553539b`, my own build)**: `r1: false, r2: false, r3: false` — bug confirmed present, matching the bead's own diagnosis exactly.
- **After (branch `a8d310dd`)**: `r1: true, r2: true, r3: false` — number and string branches now correctly accept, symbol still correctly rejects. Matches the PR's claimed transcript exactly.

### 2. Style — matches prelude convention

`union-spec(ts, v): { matches-v(s): v match?(s) }.(ts map(to-spec-td) any(matches-v))` uses the `{ named-helper(...): ... }.(expr)` local-scoped-helper idiom. I grepped `lib/prelude.eu` for this exact pattern (`}\.\(` ) and found ~18 other uses of it throughout the file (`rotate`, `update-nth`, `scanr`, `window`, `cross`, etc.) — this is a well-established, idiomatic convention in this codebase, not a novel pattern. It also directly mirrors `partial-spec`'s existing correct catenation one line below, as the PR states.

Full suites all green, independently reproduced:
- default (blob): 499/499
- `EU_SOURCE_PRELUDE=1`: 499/499
- `EU_HEAPSYN=1`: 499/499
- `cargo fmt --all -- --check`: clean
- `cargo clippy --workspace --all-targets -- -D warnings`: clean

### 3. Fault-injection on test 189 — repeated myself, both directions

Reverted `lib/prelude.eu`'s fix to the buggy version, regenerated the blob, rebuilt, and ran `test_189_r9oy_union_as_spec`: **genuinely FAILS** (`assertion left == right failed, left: 1, right: 0`). Restored the fix, regenerated, rebuilt, reran: **genuinely PASSES**. The test's `RESULT` is a real aggregate (`[t1..t9] all-true?`, not a hardcoded literal), so it gates correctly by construction — this is exactly the pattern the audit below found missing elsewhere.

### 4. Blob determinism, CHANGELOG, agent-reference §3.5a, #1015 collision

- Determinism: two independent `cargo xtask prelude-compile` regens, byte-identical (MD5 match).
- CHANGELOG entry: accurate, matches the verified mechanism and fix.
- §3.5a judgement call: verified — the table's existing row (`A | B` union → "Predicate: true if value matches any branch") describes exactly what the fix now delivers; correctly left unedited, since only the implementation was broken, not the doc.
- #1015 collision: confirmed no overlap. #1017 doesn't touch `agent-reference.md` at all in its final form (only `lib/prelude.eu`, `CHANGELOG.md`, and the two test files) — the PR body's own collision check (§3.5a at line ~887 vs #1015's edit at line ~351) was moot since #1017 ended up not touching that file. Safe regardless of merge order with #1015 (still open, pending its own error-messages.md fix).

CI: all 17 required checks green (`gh pr checks 1017`), release jobs correctly `skipping`.

---

## Audit: hardcoded `RESULT: :PASS` files (eu-ntwg.2, pre-release pass)

Grepped `tests/harness/` for files with a literal `RESULT: :PASS` (not a computed aggregate): **17 files**. Of these, **14 contain zero `//=>` assertions** anywhere in the file (`009_lazy`, `049_tester`, `065_version_assertions`, `070_bench_validation`, `104_io_basic`, `105_io_chain`, `106_io_block_chain`, `116_io_shell_with`, `117_io_exec`, `118_io_exec_with`, `119_monad_utility`, `120_random_monad`, `121_io_exec_not_found`, `bench/021_io_loop`) — nothing for this specific bug class to silently swallow; not at risk.

The remaining **3 files have `//=>` assertions and are at risk**. I ran each directly against a release binary (default, `EU_SOURCE_PRELUDE=1`, `EU_HEAPSYN=1`) and grepped for `EXPECT FAILED`:

- **`134_match_predicate.eu`** (31 assertions) — clean, 0 `EXPECT FAILED` across all three configs. Not currently masking anything.
- **`175_sv2_to_spec.eu`** (40 assertions) — clean **now** (post-#1017 fix), 0 `EXPECT FAILED` across all three configs. This is the file that *was* silently broken by eu-r9oy (per the bead) and is now genuinely fixed, not just quietly still-red.
- **`183_widen_type_def_literals.eu`** (5 assertions) — **currently broken and silently masked.** Running it directly produces `EXPECT FAILED: expected false, got true` plus a hard runtime error (`error: assertion failed: expected false, got true`, stack trace through `reflect.eu:15:32`'s `to-data`), reproduced identically across all three engine configs — yet `cargo test --release --test harness_test test_183_widen_type_def_literals` reports `ok`, because the file's `RESULT: :PASS` (line 41) is a hardcoded literal that ignores this failure, exactly the `eu-ntwg.2` mechanism.

  I isolated the two halves of the file to find the actual fault. The first half (plain `type-def` widening on a value, lines 4–31) is genuinely correct — I built a standalone repro and confirmed `to-data` shows the properly widened types (`area: number`, `type: symbol`) and all three assertions there evaluate correctly. The second half (`result-def` widening on a *function*, lines 33–41) is where it breaks: `` `{result-def: true}` `` on `make-tag(x): { tag: :default, value: x }` triggers a checker warning — `` `result-def` on a non-function binding has no return type to capture — use `type-def` instead`` — and `Tag`'s type-data resolves to an **unresolved type variable** (`[:t-var, "Tag"]`), not a concrete record shape. Since `as-spec` on a type variable is documented (and correctly implemented, per §3.5a and my own testing during #1015's review) to degrade to `any?` — "unconstrained at runtime" — `tag-spec` silently accepts anything, so `t-result-reject`'s assertion (`{tag: 123, value: 42} match?(tag-spec) //=> false`, expecting the numeric `tag` to be rejected against a symbol-typed field) incorrectly evaluates `true`.

  I did not determine whether this is (a) a genuine `result-def`-on-a-function synthesis gap that should work and doesn't, or (b) simply a test-authoring mistake — the test exercises a shape (`result-def` on a function whose return type depends on its own polymorphic argument) that the checker's own warning suggests was never meant to synthesize a concrete alias in the first place, in which case the fix may be as simple as correcting the test to use `type-def` the way the first half does, or removing the unsupported combination. That judgement call is beyond this audit's scope — flagging with full repro for triage as instructed, not fixing here.

**Audit verdict**: one confirmed, currently-shipping bug matching the `eu-r9oy` pattern exactly — `183_widen_type_def_literals.eu`'s `result-def`-on-a-function assertion is red and silently masked. Recommend filing this as its own bead (parented under `eu-ntwg.2` or standalone, coordinator's call) before 0.13 ships, given it's a live correctness gap in the same reflection feature area as eu-r9oy. The other 16 hardcoded-`RESULT: :PASS` files are clean or not applicable to this failure mode.

— Wicket

PR #1019, opened 2026-07-15
fix(tests): 183 result-def case was test-authoring bug, not a checker gap (eu-2sa6.23)

gmorpheme:
## Summary

Triage of eu-2sa6.23 (the one live bug found by Wicket's pre-release silent-assertion audit, PR #1017 review): `tests/harness/183_widen_type_def_literals.eu`'s `result-def` case was genuinely red in all three engine configs, masked by a hardcoded `RESULT: :PASS`.

**Verdict: (b) — test-authoring bug, not an implementation gap.** No code change to the checker or prelude; only the test file is fixed.

## Triage evidence

The test's second half:

```
` { result-def: true }
make-tag(x): { tag: :default, value: x }

tag-spec: s"Tag" as-spec
t-result-match: {tag: :other, value: 42} match?(tag-spec)    //=> true
t-result-reject: {tag: 123, value: 42} match?(tag-spec)      //=> false
```

Two independent test-authoring mistakes, either one fatal on its own:

1. **No `type:` annotation on `make-tag`.** `result-def`'s return-type extraction (`final_return_type` in `src/core/typecheck/check.rs`) needs the checker to have synthesised a genuine `Type::Function` for the binding. `synthesise_lam` (`check.rs:2253`) has a documented "fast path: no block-shaped params → keep `any` (unchanged behaviour)" — a plain lambda like `make-tag`, whose parameter is used as a bare value (not block-destructured), synthesises to `Type::Any`, not a function type. `final_return_type(Any)` is `None`, so the checker correctly emits `` `result-def` on a non-function binding has no return type to capture — use `type-def` instead `` and registers no alias.
2. **Wrong alias name even setting (1) aside.** `result-def: true` (bare boolean) registers the alias under the **binding's own name** (`extract_result_def_name` returns `binding_name` for `true`) — i.e. `make-tag`, not `Tag`. The test queries `s"Tag" as-spec`, a name that was never going to be registered by this metadata regardless of whether synthesis worked.

Because `as-spec` on an unresolved type variable (`[:t-var, "Tag"]`) correctly degrades to `any?` (documented, working-as-intended behaviour — §3.5a), `tag-spec` silently accepted everything, so `t-result-reject` (expecting `{tag: 123, ...}` to be rejected) incorrectly evaluated `true`.

**This was never a supported combination**, confirmed by:
- All four `result-def` unit tests in `check.rs` (`result_def_registers_return_type_alias`, `result_def_true_uses_binding_name`, `result_def_on_non_function_emits_warning`, `both_type_def_and_result_def_register_two_aliases`) use an explicit `type:` annotation — none exercise unannotated function synthesis.
- The sibling test `tests/harness/182_typedata_alias_resolution.eu` already demonstrates the correct, working, currently-supported pattern end-to-end: `` ` { result-def: true, type: s"string → [number]" } `` on `parse-nums`, queried as `s"parse-nums"` (matching the binding's own name, consistent with `result-def: true`'s documented semantics). I ran 182 directly — clean, no warnings, all assertions correct, `RESULT: PASS`.
- `git blame`/`git log` shows the block-param-fast-path behaviour in `synthesise_lam` (commit `6fd06ee6`, B9 row-variable inference) predates the commit that introduced `183` and its widen-literals fix (`ed033e67`) by three weeks — the unannotated `result-def`-on-function case was very likely broken from the day the test was written, not a later regression.

## Fix

- Gave `make-tag` an explicit `type:` annotation (`s"any → { tag: symbol, value: any }"`) and named the alias explicitly (`result-def: "Tag"`), mirroring 182's working idiom exactly. This still demonstrates a real, working feature end-to-end (this was the only harness-level coverage of `result-def`'s "widen literal types in the synthesised alias" framing that could plausibly be salvaged — see code comment added to the file for the full rationale, so a future reader doesn't reintroduce the unannotated case).
- Converted the whole file's `RESULT` from a hardcoded `:PASS` to a computed aggregate: `[t-matches, t-rejects, t-original, t-result-match, t-result-reject] all-true? then(:PASS, :FAIL)`, matching 182's/178's idiom (the new guidance from PR #1018 — no hardcoded `RESULT: :PASS` alongside `//=>` assertions). Also switched the individual assertions from `//=>` to raw `match?` results with `not(...)` for the two negative cases — a `//=>` assertion expecting `false` returns `false` on **both** success and failure in test mode, so it can't be aggregated via `all-true?` (this is the same trap fixed in eu-r9oy's PR #1017 for `175_sv2_to_spec.eu`).

Reverted just the `make-tag` metadata to the old broken form (`` `{result-def: true}` ``, no `type:` annotation) while keeping the new aggregate `RESULT`:

```
$ eu test tests/harness/183_widen_type_def_literals.eu
183_widen_type_def_literals...FAIL
```

Restored the fix:

```
$ eu test tests/harness/183_widen_type_def_literals.eu
183_widen_type_def_literals...PASS
```

Direct (non-test-mode) run with the fix — no checker warnings, no runtime error, exit 0:

```
$ eu tests/harness/183_widen_type_def_literals.eu
---
test:
  ...
  t-matches: true
  t-rejects: true
  t-original: true
  t-result-match: true
  t-result-reject: true
  RESULT: PASS
```

## Gates

- `cargo fmt --all` — clean
- `cargo test` (default/blob path) — 499/499 harness tests pass
- `EU_SOURCE_PRELUDE=1 cargo test` — 499/499 pass
- `EU_HEAPSYN=1 cargo test` — 499/499 pass

No Rust code changed — no prelude blob regeneration needed.

## Test plan

- [x] Reproduced the checker warning + hard `//=>` panic on the original file
- [x] Confirmed `eu test` masked it as `PASS` via the hardcoded `RESULT: :PASS`
- [x] Confirmed sibling test 182 already demonstrates the correct annotated `result-def` pattern working cleanly
- [x] Applied the fix; confirmed clean run (no warnings, exit 0, all assertions correct)
- [x] Fault-injection: reintroduced the broken metadata against the new aggregate `RESULT`, confirmed it correctly fails; restored the fix, confirmed it correctly passes
- [x] Full harness suite green under default, `EU_SOURCE_PRELUDE=1`, and `EU_HEAPSYN=1`


gmorpheme:
## Wicket review — PR #1019 (eu-2sa6.23) — MERGED

**Verdict: triage confirmed sound, fix correct — merged.** Merge commit `2aaaa1fe`.

I did not merge this on trust — I independently re-derived the triage from the code, not just read the PR's account of it, per the coordinator's explicit instruction to escalate if (a) looked defensible.

### 1. Triage soundness — verified against the code directly, not just the PR's narrative

Read `synthesise_lam` myself at `src/core/typecheck/check.rs:2253`: the fast path (`if block_flags.iter().all(|&b| !b) { return Type::Any; }`) is confirmed exactly as claimed — `make-tag(x): { tag: :default, value: x }` uses its parameter as a plain value, not block-destructured, so it hits this fast path and synthesises to `Type::Any`.

Read `final_return_type` (`check.rs:3256`): it peels `Type::Function` layers and returns `None` for anything that was never a function in the first place — `final_return_type(Any)` is structurally `None`. This isn't a special-cased bug; `result-def`'s entire design requires a genuine `Type::Function` to peel a return type from, and the fast path deliberately never fabricates one for non-block-param lambdas. I traced the exact code path from the `result-def` handling block (`check.rs:1064-1086`) and confirmed it produces the exact warning text quoted in the PR verbatim.

Read `extract_result_def_name` (`check.rs:851`): confirmed `result-def: true` (bare bool) returns `binding_name`, not a queried alias name — so even with working synthesis, `s"Tag" as-spec` was always going to query a name (`Tag`) that this metadata pattern was never going to populate (it would have registered under `make-tag` instead).

Checked all four `result-def` unit tests in `check.rs` (`result_def_registers_return_type_alias`, `result_def_true_uses_binding_name`, `result_def_on_non_function_emits_warning`, `both_type_def_and_result_def_register_two_aliases`) — confirmed none exercise unannotated-function synthesis; the three positive tests all supply an explicit `type:` annotation.

Checked git provenance myself: `git log`/`git show` confirm `synthesise_lam`'s fast path was introduced in `6fd06ee6` (2026-06-03, "row-variable inference at lambda boundaries, B9"), and the commit that introduced `183` (`ed033e67`, "widen literal types in synthesised type-def/result-def aliases") is dated 2026-06-27 — 24 days later, matching the PR's "three weeks" claim and supporting "broken since day one, not a regression."

Ran sibling test `182_typedata_alias_resolution.eu` directly — genuinely clean, no warnings, all 14 assertions correct, `RESULT: PASS`. Confirms the *annotated* `result-def` pattern already works correctly end-to-end in the existing codebase.

**My own independent judgement on whether (a) is defensible**: no. `result-def` is fundamentally a return-type-extraction mechanism, and B9's row-variable inference is a narrow, deliberate, opt-in mechanism specifically for block-shaped combinator parameters (`merge`/`over`-style code) — not a general "infer types for unannotated functions" feature. Making `result-def` work on an unannotated, non-block-param function would require a materially different and broader inference capability than what B9 was designed to provide, and would work against the gradual-typing boundary that the fast path explicitly preserves ("keep `any`... unchanged behaviour"). This is a genuine test-authoring bug, not a defensible implementation gap. I did not escalate to the owner.

### 2. The fix — verified end-to-end

Ran the fixed file directly: no warnings, no runtime error, `RESULT: PASS`, exit 0 — all five assertions correctly `true`. Went further than "confirm it's gone" — built an isolated repro of just the `Tag` alias and confirmed via `to-data` that it now resolves to a genuine concrete record (`[:t-record, {tag: [...,:symbol], value: [...,:any]}]`), not the unresolved type variable it was before. This is *why* the runtime error disappeared: `t-result-reject`'s underlying `match?` call now correctly evaluates to `true` (rejecting `{tag: 123, ...}` since `123` isn't a `symbol`) instead of silently accepting via `any?` degradation — there's no longer a genuine assertion failure for `//=>`'s `__EXPECT` intrinsic to hard-crash on.

### 3. The aggregate-conversion rationale — verified with my own isolated repro

Checked `(e //=> v): __EXPECT(e, __DBG_REPR(v), e = v, e)` in `lib/prelude.eu` — confirmed it unconditionally returns `e` (the actual value), not a pass/fail boolean. Built a minimal repro: `t-pass-neg: false //=> false` (a genuinely *passing* assertion) inside `[t-pass-pos, t-pass-neg] all-true?` — confirmed this produces `RESULT: FAIL` even though both assertions are legitimately correct. This is exactly the trap the PR describes: a `//=>`-expecting-`false` assertion returns `false` on success, poisoning a naive `all-true?` aggregate regardless of whether the assertion itself passed. The switch to `not(...)`-wrapped raw `match?` results is the correct fix, and matches the same pattern already used in eu-r9oy's `189_r9oy_union_as_spec.eu`.

Repeated the fault-injection myself, both directions: reverted just the `make-tag` metadata hunk (via `Edit`, restoring the bare `` `{result-def: true}` `` with no `type:` annotation) while keeping the new aggregate `RESULT` — `cargo test test_183_widen_type_def_literals` genuinely `FAILED` (`assertion left == right failed, left: 1, right: 0`). Restored the fix — genuinely `ok` again.

### 4. Gates

- Three-config suites, independently reproduced: default 499/499, `EU_SOURCE_PRELUDE=1` 499/499, `EU_HEAPSYN=1` 499/499.
- `cargo fmt --all -- --check`: clean. `cargo clippy --workspace --all-targets -- -D warnings`: clean.
- CHANGELOG entry: accurate, matches the verified mechanism.
- No Rust code changed, no blob regeneration needed (confirmed — only `CHANGELOG.md` and the one test file are touched).
- CI: all required checks green (`gh pr checks 1019`), release jobs correctly `skipping`.
- No merge conflicts against the current master tip (confirmed via `git merge-tree`, clean).

— Wicket

PR #1020, opened 2026-07-15
chore(blob): remove dead PreludeBlob::type_summary field (eu-2sa6.20)

gmorpheme:
## Summary

Removes `PreludeBlob::type_summary` — Wicket-verified during PR #1012 re-review as write-only: sole writer `xtask/src/main.rs`, zero readers anywhere in the runtime. Its only consumer, the interim (X) summary-seeded eval-path check, was deleted in PR #1012 in favour of the (Z) blob-baked `desugared_unit_cores`-based merged check. Wire-format v3 has never shipped in a release, so removing the field pre-0.13 is free — no on-disk v3 blob to be backwards-compatible with.

**`UnitInterface::type_summary` and `PreludeSummary` the type are untouched** — that's a separate, live consumer (`driver/check.rs`'s `build_prelude_interface_for`, which derives its own `PreludeSummary` by re-running the type checker over freshly-loaded prelude source, independent of the blob). Confirmed via `grep -rn "PreludeSummary"` across `src/` and `xtask/` before touching anything — `PreludeSummary` still has 8+ live call sites (`with_seed`, `extract_prelude_summary`, `type_check_for_prelude`, `type_check_with_seed`, `UnitInterface::type_summary`) all unrelated to the blob field. No orphaned type to chase.

## Changes

1. **`src/eval/stg/blob.rs`** — removed the `type_summary: PreludeSummary` field, its doc comment, the `PreludeSummary` import, the module-doc anatomy table row, and the `minimal_blob()` test helper's initialiser.
2. **`xtask/src/main.rs`** — removed the now-pointless `PreludeSummary` extraction (`type_check_for_prelude` → plain `type_check`, discarding the unused `AliasMap`) and the `operator_overloads` plumbing that existed solely to populate `summary.operator_overloads` before baking it into the blob (`operator_type_strings`, `parse_operator_overloads` call, the assignment). The type-check itself is kept — it's a useful build-time sanity pass surfacing prelude authoring mistakes as warnings — just no longer captures or bakes a summary. Updated the module-level doc comment (step 2) and the `no eliminate` comment (its second sentence, "the type checker needs to see all bindings for their type schemes," was specific to the now-removed extraction).
3. **`build.rs` + `xtask/src/main.rs`** — `BYTECODE_WIRE_FORMAT_VERSION` bumped `3` → `4` (see below), with an updated version-history doc comment in both places (they must match).
4. **`CHANGELOG.md`** — new `### Changed` entry in `[Unreleased]`.

## Wire-format version: bumped 3 → 4

**Decision: bump.** Reasoning:

`build.rs`'s staleness check (`verify_prelude_blob`) only reads the blob's first 32 bytes (`source_hash`) — it does **not** decode the full struct, so it's robust to layout changes by construction, but only *because* `source_hash = SHA-256(prelude_source ‖ BYTECODE_WIRE_FORMAT_VERSION)` folds the wire-format version into the hash. If `lib/prelude.eu` is unchanged and I do **not** bump the version, a stale on-disk blob generated by the *old* xtask (with `type_summary` still present, at the old byte layout) would produce an identical `source_hash` to a freshly-generated *new*-layout blob — `cfg(prelude_blob_ok)` would be set, `include_bytes!` would embed the **stale, wrong-layout** blob, and the new binary's `PreludeBlob::from_bytes` would attempt to postcard-decode it against the new (fewer-fields) struct definition.

I verified this is not merely theoretical:
- `build.rs`'s own doc comment establishes the precedent — "Bump this whenever the encoder's byte layout changes" — and the existing v1→v2 bump (`u16`→`u32` binding-count width, eu-2sa6.11) already bumped for a **pure layout change**, not a field addition. Removing a field is exactly as much a layout change.
- I **empirically reproduced the failure mode**: generated an old v3-format blob (with `type_summary` present) from unmodified `origin/master`, then attempted to decode those exact bytes with the new (v4) `PreludeBlob::from_bytes` in a scratch, uncommitted test. Result: `Serde Deserialization Error` — postcard's positional decoding correctly detects the misalignment (the `type_summary` position's bytes get mis-read as the next field's varint-prefixed collection length, producing bounds that don't fit the remaining byte stream) and returns `Err` rather than silently succeeding with corrupted data.
- Separately, I confirmed the **runtime** loading path (`maybe_load_prelude_blob` in `driver/eval.rs`) already treats `PreludeBlob::from_bytes`'s `Result` correctly — on `Err` it prints a warning and returns `None`, and the driver falls back cleanly to source-compiling the prelude. It does not panic. So even in the hypothetical case where the version bump were somehow skipped, the runtime fallback is a safe backstop — but the version bump is the documented, established, **primary** mechanism (build-time, unambiguous, avoids ever hitting the decode-error path in practice), and I'm following that precedent rather than relying on the decode-error path as primary.

Both pieces of evidence are consistent: bump the version, as `build.rs`'s own doc comment instructs for any layout change, and as the v1→v2 precedent already did for a comparably-scoped change.

## Blob size

Regenerated `lib/prelude.blob` from a clean baseline build of unmodified `origin/master` (`0c3e2a44`) for an exact comparison, not the dispatch's stated approximate figure:

| | bytes |
|---|--:|
| before (baseline, `origin/master`) | 598,398 |
| after (this branch) | 596,625 |
| **Δ** | **−1,773 bytes (−0.30%)** |

Determinism: two independent `cargo xtask prelude-compile` regenerations of the fixed blob are byte-identical (SHA-256 match).

## CHANGELOG note on #1016

#1016 (Furnace, in flight) also touches `CHANGELOG.md` — it inserts a **new** `### Changed` section immediately after the single `### Added` bullet and before the existing `### Fixed` section (i.e. at the top of `[Unreleased]`'s body). My entry is a **separate** new `### Changed` section appended at the **end** of `[Unreleased]`'s existing content (after the last `### Fixed` bullet, before the `## [0.12.1]` header) — a different insertion point in the file. This file already has a precedent of multiple `### Changed`/`### Fixed` headers stacked within the same version block (see `[0.12.0]` further down), so two separate `### Changed` sections coexisting is not unusual for this file's style. Low collision risk, but flagging per the dispatch in case of a trivial merge conflict depending on merge order.

## Gates

- `cargo fmt --all` — clean
- `cargo test` (default/blob path) — 1310 lib tests + 499 harness tests, all green
- `EU_SOURCE_PRELUDE=1 cargo test` — all green
- `EU_HEAPSYN=1 cargo test` — all green
- `blob_corrupted_bytes_fail_gracefully` (the #1012 format-mismatch fallback test) — still passes
- All 14 `blob::tests::*` unit tests pass, including `embedded_bytecode_matches_fresh_encode` and the `embedded_blob_*` tests against the actual regenerated, embedded blob

## Test plan

- [x] `grep`-verified `PreludeSummary` has no orphaned readers/writers beyond the removed blob field — `UnitInterface`/`driver/check.rs` path confirmed live and untouched
- [x] Removed field, bake logic, and doc-table row; updated xtask's module doc and `no eliminate` comment
- [x] Bumped `BYTECODE_WIRE_FORMAT_VERSION` in both `build.rs` and `xtask/src/main.rs` (kept in sync); updated both version-history comments
- [x] Empirically verified an old v3-format blob fails to decode against the new v4 struct (scratch test, not committed) — confirms no silent corruption risk
- [x] Confirmed the runtime fallback path handles decode `Err` gracefully (no panic)
- [x] Regenerated blob against a precise before/after baseline; reported exact byte delta
- [x] Verified blob regeneration determinism (byte-identical across two runs)
- [x] Full gates green under default, `EU_SOURCE_PRELUDE=1`, and `EU_HEAPSYN=1`


gmorpheme:
## Wicket review — PR #1020 (eu-2sa6.20) — MERGED

**Verdict: approve and merge.** Every claim independently re-derived from the code, not read off the PR's account. Merged as `a9c69f20`.

### 1. Version bump 3→4 — re-derived independently, agree it's correct

Read `verify_prelude_blob` in `build.rs` myself: it reads only the first 32 bytes (`source_hash`) via `read_blob_source_hash`, never decoding the full `PreludeBlob` struct — its own comment explains this is deliberate ("to avoid parsing the full blob just for the hash"). This means the version-in-hash is the **only** mechanism standing between a stale-layout-but-hash-matching blob and a runtime decode failure; there is no independent structural check. This isn't a design flaw the PR introduced — it's a pre-existing property of the staleness check that makes the version bump load-bearing for *any* layout change, additions or removals alike.

I confirmed `type_summary` had no `#[serde(default)]` in the pre-removal struct (checked `git show 0c3e2a44:src/eval/stg/blob.rs` directly) and sat in the middle of the struct — five more fields follow it (`monad_specs`, `monad_type_hints`, `inlinable_bindings`, `desugared_unit_cores`, `bytecode`). Since postcard is purely positional, removing a mid-struct field unconditionally shifts every subsequent field's byte offset regardless of any field's own `#[serde(default)]` annotation (that attribute only helps when the byte stream runs out entirely for *trailing* fields, not when bytes exist but are misaligned). This is a structurally necessary bump, not a judgement call.

Verified the v1→v2 precedent PR cites: `git log -S` found `728defe7` (`eu-2sa6.11`), confirmed it introduced the whole `BYTECODE_WIRE_FORMAT_VERSION` mechanism for a `u16`→`u32` widening inside the *encoded bytecode stream*. I'd note this precedent is about the byte-stream-internal layout, one layer removed from the `PreludeBlob` struct's own field layout — the more directly analogous precedent is actually v3 itself (`desugared_unit_cores` field *addition*, same class of change as this PR's field *removal*), which the PR doesn't explicitly cite but which supports the same conclusion. Both precedents point the same way regardless.

I did not just trust the "empirically reproduced" claim — I reproduced it myself in a scratch (uncommitted) test: built an old v3-format blob from a clean `origin/master` (`0c3e2a44`) build, then attempted `PreludeBlob::from_bytes` against the new v4 struct. Result: `Err(SerdeDeCustom)` — a clean decode error, not corruption or a panic, confirmed via `eprintln!` in the test before I discarded it. Also re-confirmed `maybe_load_prelude_blob` (`driver/eval.rs`) treats this `Err` correctly — prints a warning, returns `None`, driver falls back to source-compile, no panic.

**My independent conclusion: the v4 bump is correct and necessary, not just defensible.** Both `build.rs` and `xtask/src/main.rs` have `BYTECODE_WIRE_FORMAT_VERSION: u32 = 4`, confirmed in sync, with matching version-history doc comments.

### 2. Scope discipline — verified nothing live removed, nothing dead left

Grepped `PreludeSummary` across `src/` and `xtask/`: confirmed 8 live references outside the removed field — `Checker::with_seed`, `extract_prelude_summary`, `type_check_for_prelude`, `type_check_with_seed`, and `UnitInterface::type_summary` (a genuinely separate field on a different struct). Followed the call chain further: `type_check_for_prelude` and `type_check_with_seed` both still have live callers in `src/driver/check.rs`'s `build_prelude_interface_for`/`get_or_build_prelude_interface_for` — an entirely separate, unrelated LSP/checker prelude-interface cache, untouched by this PR. Zero orphaned functions.

Compared `type_check` and `type_check_for_prelude`'s implementations directly: both construct `Checker::new()` and call `checker.check_expr(expr)` — identical core logic. The only difference is `type_check_for_prelude` additionally sets `keep_root_scope = true` and calls `extract_prelude_summary()` *after* checking, which is orthogonal to warning generation. This confirms xtask's swap from `type_check_for_prelude` to plain `type_check` produces byte-identical warning behaviour — the sanity pass's diagnostic coverage is unchanged.

On "did it gate or just print" — checked the actual diff: the `eprintln!("  {} type-check warning(s) (non-fatal)", ...)` line is **unchanged** by this PR, present identically before and after. It was never gating the xtask build (explicitly labelled "non-fatal" already), and it still isn't. No weakening — the behaviour was already print-only.

Final sweep: grepped `\.type_summary\b|type_summary:` across `src/`, `xtask/`, `docs/` — every remaining hit is `UnitInterface::type_summary` or its design doc, zero dangling references to the removed `PreludeBlob::type_summary`.

### 3. Blob regeneration and determinism — reproduced exactly

Built master (`0c3e2a44`) separately and generated a baseline blob myself: **598,398 bytes**, exact match to the PR's stated baseline. Regenerated at the PR head: **596,625 bytes** — exact match, giving a delta of precisely **−1,773 bytes**, matching the claim byte-for-byte. Two independent regenerations at the PR head: MD5-identical. Ran all 14 `blob::tests::*` unit tests including the two specifically cited (`embedded_bytecode_matches_fresh_encode`, `blob_corrupted_bytes_fail_gracefully`) — all green.

### 4. CHANGELOG collision with #1016 — real, but the PR's own precedent citation is wrong

Checked this thoroughly since it changes the actionable outcome. Mapped every `### <Type>` header in `CHANGELOG.md` to its enclosing `## [version]` block, all the way back to `[0.6.0]`: **every single version block in this file's history uses each header type (Added/Changed/Fixed/Performance/Removed) at most once.** There is no existing precedent anywhere in this file for two same-type headers (e.g. two separate `### Changed` sections) stacked within one version block — `[0.12.0]`, which the PR cites as precedent, in fact has exactly one `### Changed` section and nothing else.

So: if #1016 lands with its `### Changed` insertion (between `### Added` and `### Fixed`) after this PR's `### Changed` (appended at the end of `[Unreleased]`, before `## [0.12.1]`) is already in place, the two will not textually conflict (different insertion points), but the result will be a genuinely novel pattern for this file — two separate `### Changed` headers in one version block. **I don't think that should ship as-is.** Whichever of #1016/#1020 merges second should consolidate into a single `### Changed` section. Since #1020 is merging now (first), this is a note for #1016's review, not a blocker here — I'll enforce it when I review #1016.

(Unrelated aside noticed while checking #1016's file list for this comparison: #1016 also adds `tests/harness/189_branch_gated_self_recursion.eu`, but `189_r9oy_union_as_spec.eu` from #1017 is already merged — a harness test-numbering collision, different files so no git conflict, but worth flagging to whoever reviews #1016.)

### 5. Full suites

- Three engine configs, independently reproduced: default 499/499, `EU_SOURCE_PRELUDE=1` 499/499, `EU_HEAPSYN=1` 499/499.
- `cargo test --release --lib`: 1310/1310.
- `cargo fmt --all -- --check`: clean. `cargo clippy --workspace --all-targets -- -D warnings`: clean.
- CI: all required checks green (`gh pr checks 1020`), release jobs correctly `skipping`.
- No merge conflicts against the current master tip (post-#1019), confirmed via `git merge-tree`.

— Wicket

PR #1023, opened 2026-07-16
test: build the real 3-mode differential gate (eu-ntwg.3)

gmorpheme:
## Summary

The pre-decoded execution IR design doc (#989 §6 Step A gate) specified extending `differential.rs` to run byte-path bytecode, pre-decoded bytecode and HeapSyn under **one mechanism** asserting all three agree. The existing harness instead ran two disjoint pairwise comparisons (HeapSyn-vs-byte-path bytecode in `differential.rs`; pre-decoded coverage lived elsewhere) and relied on transitive agreement — flagged as an undisclosed design deviation in the owner-ordered formal re-review of #1002, and recorded as a precondition (eu-ntwg.3) blocking the `EU_PREDECODE` default-flip soak (eu-vcr8).

This PR builds the actual 3-mode gate:

- **`assert_three_way<T: PartialEq + Debug>`** is the single agreement mechanism: given a case's result from all three engine configurations, it asserts pairwise equality across all three and, on failure, names exactly which configuration(s) diverged (e.g. "HeapSyn != bytecode pre-decoded" vs. "bytecode byte-path != bytecode pre-decoded") — distinguishing "pre-decode broke it" from "the byte-path itself is wrong" from "the two bytecode dispatch paths disagree with each other", none of which a HeapSyn-vs-bytecode pairwise check alone can localise.
- Every existing comparison in `differential.rs` — exit-code agreement, rendered-output agreement, `LOOKUP_FAIL` diagnostic agreement, `NotValue` diagnostic agreement, and the stateful producer-stream test — now runs through `assert_three_way` across all three configurations, not two.
- **`BytecodeMachine`'s dispatch-mode selection** (`src/eval/bytecode/machine.rs`) is refactored behind a `#[cfg(test)]` thread-local override (`with_predecode_override` / `resolve_predecode`) so the harness can force the byte-path or pre-decoded dispatch mode independent of `EU_PREDECODE`, without racing on that process-global env var across parallel `cargo test` threads. Production construction (`BytecodeMachine::new`) is unaffected — `resolve_predecode()` falls back to the existing `predecode_enabled()` env-var read whenever no override is active, so non-test behaviour is byte-identical to before this change.

Zero current user exposure — `EU_PREDECODE` is still opt-in. This is test infrastructure only, not an engine-default change.

## Design deviations from #989

None. This restores the single-mechanism 3-mode gate exactly as specified in #989 §6 Step A; no scope was added or dropped beyond what the design called for.

Temporarily corrupted the pre-decoded `Case` dispatch's branch-tag comparison in `handle_op_predecoded` (`src/eval/bytecode/machine.rs`, the `BranchPredecoded` continuation build):

```rust
min_tag: instr.min_tag().wrapping_add(1), // FAULT INJECTION eu-ntwg.3
```

This breaks only the pre-decoded dispatch path (byte-path and HeapSyn are untouched). Result: **13 of the 26 differential tests failed**, and in every failure the gate correctly and exclusively named `bytecode pre-decoded` as the diverging mode, e.g.:

```
3-mode differential gate: engines disagree:
HeapSyn (Some(42)) != bytecode pre-decoded (Some(0))
bytecode byte-path (Some(42)) != bytecode pre-decoded (Some(0))
```

confirming HeapSyn and the byte-path continued to agree with each other while pre-decoded alone diverged — exactly the localisation the 3-mode single-mechanism design is meant to provide over two disjoint pairwise checks. Restored the fault; all 26 differential tests pass again.

## Test plan

- [x] `cargo fmt --all`
- [x] `cargo test --lib eval::bytecode::differential` — 26/26 pass
- [x] `EU_GC_VERIFY=2 EU_GC_STRESS=1 cargo test --lib eval::bytecode` — 79/79 pass
- [x] `cargo xtask prelude-compile` then `EU_PREDECODE=1 cargo test --test harness_test` — 500/500 pass
- [x] Full `cargo test` (all suites) — all green, 0 failures
- [x] Fault-injection verified (see above): inject → 13/26 fail with correct mode-naming → restore → 26/26 pass

### DIFF
--- a/src/eval/bytecode/differential.rs
+++ b/src/eval/bytecode/differential.rs
@@ -1,6 +1,24 @@
-//! Differential testing: run a synthetic `StgSyn` program through both the
-//! HeapSyn engine and the bytecode engine and assert they agree (spec §9,
-//! REFINEMENT B — harness-first on synthetic programs, before the prelude).
+//! Differential testing: run a synthetic `StgSyn` program through all three
+//! engine configurations — the HeapSyn tree-walk machine, the bytecode
+//! machine on its byte-path dispatch, and the bytecode machine on its
+//! pre-decoded dispatch (`EU_PREDECODE`, eu-2sa6.13) — and assert all three
+//! agree in one mechanism (spec §9, REFINEMENT B — harness-first on
+//! synthetic programs, before the prelude; 3-mode gate per the
+//! pre-decoded-IR design doc §6 Step A, docs/superpowers/specs/
+//! 2026-07-13-predecoded-execution-ir-design.md, built under eu-ntwg.3).
+//!
+//! Earlier revisions of this harness ran two separate pairwise comparisons
+//! (HeapSyn vs. byte-path bytecode, with pre-decoded bytecode covered by a
+//! different, disjoint mechanism) and inferred three-way agreement
+//! transitively. That is a weaker guarantee than the design specified: it
+//! never observes all three configurations disagreeing in a cyclic way in a
+//! single assertion, and a bug that only manifests as a divergence between
+//! the two *bytecode* dispatch paths (byte vs. pre-decoded) with HeapSyn
+//! coincidentally agreeing with one of them could slip through unreported.
+//! [`assert_three_way`] is the single mechanism: every comparison in this
+//! file runs a case under all three configurations and asserts pairwise
+//! equality across all three in one place, naming exactly which mode(s)
+//! diverged on failure.
 //!
 //! This validates the bytecode machine against the reference engine at the
 //! value level (exit code) using the same runtime globals for both. The full
@@ -9,9 +27,11 @@
 
 #![cfg(test)]
 
+use std::fmt::Debug;
 use std::rc::Rc;
 
 use crate::common::sourcemap::SourceMap;
+use crate::eval::bytecode::machine::with_predecode_override;
 use crate::eval::bytecode::{encode, BytecodeMachine};
 use crate::eval::emit::NullEmitter;
 use crate::eval::machine::intrinsic::StgIntrinsic;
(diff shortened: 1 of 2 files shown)


gmorpheme:

Verified independently in an isolated worktree (`/private/tmp/eu-wicket-1023`), not just re-reading the PR body's claims.

**Spec conformance (design doc §6 Step A, docs/superpowers/specs/2026-07-13-predecoded-execution-ir-design.md):** confirmed the doc's exact wording — "`differential.rs` ... extended to run three modes — byte-path bytecode, pre-decoded bytecode, HeapSyn — asserting all three agree, not just bytecode vs. HeapSyn as today." `assert_three_way` is a single mechanism (differential.rs:53) that every one of the 26 tests funnels through, either directly or via one of the four wrapper functions (`assert_engines_agree`, `assert_engines_render_agree`, `assert_engines_lookup_fail_agree`, `assert_engines_not_value_agree`). Confirmed HeapSyn is genuinely exercised in every path (each wrapper explicitly builds and runs `standard_machine`, not skipped or stubbed).

**machine.rs override:** `PREDECODE_OVERRIDE` and `with_predecode_override` are both `#[cfg(test)]`; `resolve_predecode()` has two bodies gated on `#[cfg(test)]`/`#[cfg(not(test))]`, and the non-test body is `super::predecode_enabled()` — byte-identical to the pre-PR call site. Confirmed `predecode_enabled()` has no other call sites outside `machine.rs`, so there's no bypass. The RAII `Guard` clears the thread-local on drop, so a panicking `.expect()` inside `f` (e.g. a failed machine build) can't leak a forced mode into the next test on that thread.

**Fault injection, independently reproduced:** injected the exact fault described in the PR body (`min_tag: instr.min_tag().wrapping_add(1)` in the `Op::Case` arm of `handle_op_predecoded`, machine.rs:2365) in my own worktree. Result: 13/26 tests failed, every failure exclusively implicating the pre-decoded path (either via `assert_three_way`'s "HeapSyn != bytecode pre-decoded" / "byte-path != bytecode pre-decoded" message, or via a panic during pre-decoded machine execution itself — `NoBranchForDataTag` — for the two tests that fail before reaching the comparison). Restored the fault; confirmed clean `git diff` and all 26 tests green again.

**Verification performed:**
- Clean release build.
- `cargo fmt --all -- --check` — clean.
- `cargo clippy --all-targets -- -D warnings` — clean.
- Full `cargo test --release` — 1306 lib + 500 harness + all other suites green, 0 failures (matches master's own baseline lib-test count exactly, so no regression from the refactor).
- `EU_PREDECODE=1 cargo test --lib eval::bytecode::differential` — 26/26 pass.
- `EU_HEAPSYN=1 cargo test --lib eval::bytecode::differential` — 26/26 pass (expected: the thread-local override supersedes both env vars inside these tests by design).
- CI fully green across all required checks.

No design deviations found. This is the recorded review for this PR (it touches `src/eval/bytecode/machine.rs`, engine-adjacent code). Merging.

PR #1025, opened 2026-07-16
fix(test): gate harness verdict on genuine EXPECT FAILED diagnostics

gmorpheme:
## Summary

Closes the systemic half of eu-ntwg.2: `lib/test.eu`'s `default-expectation` judged a target solely on its `RESULT` key when present, so a hardcoded `RESULT: :PASS` could mask a genuinely failing `//=`/`//!`/`//=>`/`//=?`/`//=?>` assertion forever — the exact bug already observed in `175_sv2_to_spec.eu` during the eu-r9oy investigation (pre-release audit for that file, done separately by Wicket, found it currently clean but structurally unprotected).

- `default-expectation` now also fails a target if its captured `stderr` contains an `EXPECT FAILED` diagnostic, regardless of what `RESULT` says. Refactored the shared verdict logic into `compute-verdict`, and added `unguarded-expectation` (the old, RESULT-only behaviour) as an explicit opt-out for tests that deliberately trigger `EXPECT FAILED` to verify the assertion mechanism itself — `125_expectations.eu` now opts in via `verify: ["unguarded-expectation"]`.
- **Second trap (also in eu-ntwg.2's bead comment):** a `//=>` assertion expecting `false` returns `false` in test mode on both success *and* failure, so a hand-computed `RESULT` built from `all-true?` over such bindings can't tell them apart. The new stderr-based gate is immune to this (it doesn't look at return values at all), and `docs/guide/testing.md` now documents the trap and the `not()`-wrapped-raw-boolean convention (`182_typedata_alias_resolution.eu`/`183_widen_type_def_literals.eu`) for anyone hand-computing a `RESULT`.
- **Root cause dependency found along the way:** the stderr scan only works because of a deeper bug it uncovered — `Expect::execute` (`src/eval/stg/expect.rs`) printed `"EXPECT FAILED"` via a bare `eprintln!`, which writes straight to the process's real stderr and bypasses the `Box<dyn Write>` capture the test harness installs via `Executor::capture_output`. The diagnostic never reached `evidence.yaml`'s `stderr` field at all — confirmed with a hex/text dump of the evidence before the fix (`stderr: []` even though "EXPECT FAILED" printed to the terminal). Fixed by adding `IntrinsicMachine::record_diagnostic`, buffering messages on `MachineState` (HeapSyn) / `BcMachineState` (bytecode), and draining them in `Executor::try_execute` via a new `flush_diagnostics` (captured sink if present, else real stderr — preserving today's interactive `eu run` behaviour). Implemented for both engines.

## Audit finding

Empirically swept every harness file with a `target:` metadata block (81 files) for `EXPECT FAILED` under **current, passing** behaviour: only two intentionally trigger it — `049_tester.eu`'s `test-fail` target (already isolated via its own custom `verify: [:all-values-false]` validator, unaffected by `default-expectation` changes) and `125_expectations.eu` (fixed here). `175_sv2_to_spec.eu` and `134_match_predicate.eu` still use a hardcoded `RESULT: :PASS` but are now safe under the new stderr gate (verified via fault injection below) — I left them as-is rather than rewrite ~90 assertions, since the correctness fix no longer depends on how they compute `RESULT`; happy to convert them to a computed `RESULT` in a follow-up if the team prefers belt-and-braces over the docs-only convention note.

## Fault-injection evidence

All against `175_sv2_to_spec.eu` (still `RESULT: :PASS`), using `cargo test`/`eu test` directly, each restored afterward and confirmed `git status` clean:

1. **Expecting-true assertion, made to fail**: `t-prim-num: 42 match?(num-spec) //=> true` → `//=> false`. Pre-fix (stashed the fix, kept the fault): `175_sv2_to_spec...PASS` despite `EXPECT FAILED: expected false, got true` printing. Post-fix: `175_sv2_to_spec...FAIL`, and the evidence's `stderr` field now actually contains `"EXPECT FAILED: expected false, got true"` (previously `stderr: []` even when the diagnostic printed to the terminal — the root-cause bug above).
2. **Expecting-false assertion, made to fail (the "second trap")**: `t-prim-str-miss: 42 match?(str-spec) //=> false` → `"42" match?(str-spec) //=> false` (now genuinely matches, so the false-expecting assertion mismatches). Pre-fix: `PASS` (silently masked). Post-fix: `FAIL` on both the bytecode and `EU_HEAPSYN=1` engines.
3. **125_expectations.eu opt-out**: confirmed it still reports `PASS` with the fix in place (it deliberately triggers `EXPECT FAILED` via `eq-fail`/`true-fail`, and would otherwise be broken by the new gate without `unguarded-expectation`).
4. Confirmed non-test-mode (`eu run`) behaviour is unchanged: a failing `//=` still prints `EXPECT FAILED: ...` to the terminal followed by the usual `error: assertion failed` diagnostic and exits 1.

## Test plan

- [x] `cargo fmt --all`
- [x] `cargo test` — 1306+500+... all green (full suite, both `harness_test` at 500/500)
- [x] `EU_HEAPSYN=1 cargo test --test harness_test` — 500/500 (both engines wired for diagnostic capture)
- [x] Fault injection as above, verified fail-before/pass-after semantics flip correctly, all reverted before commit

### DIFF
--- a/docs/guide/testing.md
+++ b/docs/guide/testing.md
@@ -55,6 +55,19 @@ from the checks keeps every check in the verdict — see
 `tests/harness/182_typedata_alias_resolution.eu`, which gather their
 checks into a list and finish with `all-true? then(:PASS, :FAIL)`.
 
+Regardless of `RESULT`, `default-expectation` also fails the target if
+`stderr` contains an `EXPECT FAILED` diagnostic (see
+[Inline assertions](#inline-assertions) below) — a target can no longer
+mask a genuinely failing `//=`/`//!`/`//=>`/`//=?`/`//=?>` assertion
+behind a `RESULT` that doesn't account for it, such as a hardcoded
+`RESULT: :PASS`. Tests that deliberately construct assertions expected
+to fail, to verify the assertion mechanism itself reports failure
+correctly, should use `unguarded-expectation` instead (a target-level
+`verify: ["unguarded-expectation"]`) — see
+`tests/harness/125_expectations.eu`, which checks that `//=` and `//!`
+correctly return `false` on mismatch and therefore genuinely triggers
+`EXPECT FAILED` diagnostics as part of passing.
+
 ## Inline assertions
 
 An expression can carry an inline expected value with the `//=>`
@@ -72,7 +85,20 @@ mode the expression then evaluates to `false` and execution continues
 binding, an inline assertion reaches a target's verdict only through
 the `result` the expectation function sees — for instance by feeding
 its value into `RESULT`, or via the all-values-true inference when no
-`RESULT` key is present.
+`RESULT` key is present — and, independently of `result`, through
+`default-expectation`'s `stderr` gate described above.
+
+Because a failing `//=>` returns `false` in test mode regardless of
+whether the *expected* value was `true` or `false`, an assertion that
+expects `false` cannot be told apart from a passing one by its return
+value alone: `x //=> false` evaluates to `false` both when `x` really
+is `false` (pass) and when `x` mismatches (fail). Hand-computing a
+`RESULT` with `all-true?` over such bindings therefore cannot
+distinguish pass from fail for expecting-`false` assertions. Use the
+raw predicate result instead (with `not()` where you want to assert
+absence), as `tests/harness/182_typedata_alias_resolution.eu` and
+`tests/harness/183_widen_type_def_literals.eu` do, rather than folding
+a `//=>`-wrapped expecting-`false` value into the aggregate.
 
 ## Test files
 
--- a/lib/test.eu
+++ b/lib/test.eu
@@ -19,10 +19,13 @@ A validation function accepts an evidence context with the following keys:
 as described below and must return :PASS or :FAIL.
 "
 validation: {
-  ` "By default we expect a RESULT key (PASS/FAIL) or infer from all values being true"
-  default-expectation(context): {
-    r: context.result
-    # Handle bare boolean results directly
+  ` "Core verdict computation shared by `default-expectation` and
+  `unguarded-expectation`: if `result` is a block containing a RESULT
+  key, the verdict is that key's value (true/false mapped to
+  :PASS/:FAIL, anything else used directly); otherwise if `result` is
+  a block, the verdict is whether every value in it is true;
+  otherwise `result` is treated as a bare boolean."
+  compute-verdict(r): {
     is-block: r block?
     has-result: if(is-block, r has(:RESULT), false)
     explicit: if(has-result, {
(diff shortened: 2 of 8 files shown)


gmorpheme:

Verified independently in an isolated worktree (`/private/tmp/eu-wicket-1025`), rebased cleanly onto current master (post-#1023/#1024).

**Diagnostic buffering — verified it can't reorder, drop, or leak between runs:**
- `IntrinsicMachine::record_diagnostic` is implemented on all four call sites the trait requires (`MachineState`, `MachineBifContext` for HeapSyn; `BcBifContext` for bytecode) — guaranteed exhaustive by the compiler since it's a new non-defaulted trait method.
- `Executor::flush_diagnostics` is called after `inject_world_and_run` (the IO-driven sub-evaluation) on both engine branches, so diagnostics recorded during IO-driven re-entry are captured, not dropped by an early drain.
- Flush happens unconditionally on the `result` computation path (not gated behind an early return), matching the existing "stats collected even on error" pattern — confirmed by reading the surrounding control flow in `src/driver/eval.rs`.
- Checked `tests/driver/tester.rs`'s `InProcessTester::run`: a fresh `SourceLoader` and fresh `Executor` are constructed inside the innermost `for (t, fs) in plan.targets() { for f in fs { ... } }` loop, so there is no possibility of diagnostics leaking across different (target, format) invocations — each gets its own buffer from scratch.

**Fault injection, independently reproduced (both claimed scenarios, both engines):**
1. Expecting-true assertion made to fail (`175_sv2_to_spec.eu`'s `t-prim-num: ... //=> true` → `//=> false`): test genuinely FAILs under `cargo test`. Inspected the resulting `evidence.yaml` directly — `stdout` still shows the stale `RESULT: PASS`, but `stderr` now genuinely contains `"EXPECT FAILED: expected false, got true"` (previously this field was empty even when the diagnostic printed — the exact root-cause bug the PR describes). This is direct proof the `eprintln!` → `record_diagnostic`/capture-buffer fix works, not just the gating logic.
2. The "second trap" (expecting-false assertion made to fail, `t-prim-str-miss`): genuinely FAILs on both the default (bytecode) and `EU_HEAPSYN=1` engines.
3. `125_expectations.eu`'s `unguarded-expectation` opt-out still passes (`test_harness_125`).
4. Restored all fault injections; confirmed clean `git diff` and passing tests before proceeding.

**Other checks:**
- Spot-checked the audit claim on `049_tester.eu`: it has three targets — `test-⊕` and `test-fail` use custom validators (`verify: [...]`) entirely bypassing `default-expectation`, and `test-default-expectation` has no assertions at all (nothing to trigger `EXPECT FAILED`) — confirms the "unaffected" claim precisely.
- `docs/guide/testing.md` update is accurate: documents both the new stderr gate and the "expecting-`false` can't be told apart from failure by return value alone" trap, pointing at the correct `182`/`183` convention examples.
- `lib/test.eu` is not part of the prelude blob (`cargo xtask prelude-compile` only touches `lib/prelude.eu` → `lib/prelude.blob`, confirmed by reading `xtask/src/main.rs`) — no blob regeneration needed for this change.
- `cargo fmt --all -- --check` — clean. `cargo clippy --all-targets -- -D warnings` — clean.
- Full `cargo test --release` — 1306 lib + 500 harness (incl. `EU_HEAPSYN=1` coverage) + all other suites green, 0 failures.
- CI fully green across all required checks.

This is the recorded review for this PR — it touches both engines' machine state structs (`MachineState` and `BcMachineState`). No design deviations found; the fix is exactly what eu-ntwg.2 called for. Merging.

PR #1026, opened 2026-07-16
test: assert eval-path warning parity across the typecheck fixture set

gmorpheme:
## Summary

Closes eu-ntwg.1: `tests/harness/typecheck/` fixtures (111 files) were only ever exercised via `eu check --strict` (`run_typecheck_test` in `tests/harness_test.rs`) — never plain `eu <file>` eval — so any checker/eval divergence was invisible to CI. `#1012` added a dedicated single-file eval-path test for `065_lookup_key_typo_warns.eu` after that exact divergence shipped silently; this generalises that coverage across the fixture set.

- `run_typecheck_test` now also runs the plain eval path for every fixture where `eu check --strict` genuinely warns (a non-empty expected `stderr:` pattern in the `.expect` sidecar — 36 of 111 fixtures) and asserts the *same* pattern appears on the eval path's stderr, since type checking runs unconditionally regardless of `--strict` (only the abort-on-warning behaviour is `--strict`-gated).
- Added `run_with_deadline` (backed by the new `wait-timeout` dev-dependency) instead of `Command::output()`'s indefinite block: typecheck fixtures are check-only test material, and surveying the full set found one, `092_self_assign_arg_pos_ok.eu` (`ones: cons(1, ones)`), that's a deliberate infinite lazy fixpoint never meant to be evaluated — it exists purely to verify the checker doesn't flag argument-position self-reference. A bare `.output()` call would hang indefinitely on it.

## A genuine bug found along the way

The new coverage immediately caught a real, previously invisible divergence: `004_invalid_annotation.eu`'s malformed `` ` { type: "number ->" } `` annotation is a hard *error* under `eu check --strict` (exit 1, "invalid type annotation... expected a type, got Eof") but is **silently discarded** under plain `eu` (exit 0, no diagnostic at all — the annotation is dropped as if it were never written).

Root cause: `eu check`'s annotation-syntax validation (`check_annotation_syntax` in `src/driver/check.rs`) is a separate pre-pass unique to the `check` subcommand. The type checker used on the ordinary eval path (`core/typecheck/check.rs`, three call sites) calls `parse::parse_scheme(&type_str).ok()?`, which discards the `Err` entirely instead of surfacing it as a warning.

This is out of scope for a harness-coverage PR (it's a `core/typecheck` fix, not a test-infrastructure one), so I filed it as **eu-5q08** (P1) with the full root-cause trace, and excluded `004_invalid_annotation.eu` from the new eval-path assertion with a comment pointing at the bead — not silently; the exclusion is a single `if filename == ... { return; }` with a full explanation, and it should be re-included once eu-5q08 is fixed.

## Runtime cost (measured, corrected)

My first pass measured cost with a naive serial bash loop (`for f in ...; do timeout 10 eu $f; done`) and got ~50s for a full 111-fixture sweep, which looked prohibitive. That number was misleading: `cargo test` runs tests in parallel (9-10 concurrent threads on this machine), so I re-measured the actual integrated cost with a controlled A/B — full `cargo test --test harness_test` (500 tests) on the same machine, back-to-back:

- Before this change: 70.39s
- After this change (36 extra `eu` subprocess spawns): 70.88s

The real incremental cost is under a second, within run-to-run noise — not the 50s the serial estimate suggested. Given that, the 75-fixture "checker-clean" remainder (asserting warning *absence* rather than presence) would likely also be cheap, but I did **not** extend to it: the type checker's warning output is prefixed with ANSI colour codes that split "warning" from the following ":" mid-token (`ColorChoice::Auto` doesn't appear to detect the piped/non-tty `Command::output()` capture in this codebase), so a robust "no warning present" substring check needs a bit more care (either a `--no-color`-style flag or a marker proven not to collide with unrelated diagnostic text) than a same-day extension of this PR should rush. Flagging as a natural, now-cheap follow-up rather than doing it under time pressure.

## Fault-injection evidence

Isolated the new eval-path assertion from the pre-existing check-mode assertion (which shares the same helper function) by temporarily pointing the eval-path command at a different, warning-free fixture (`005_no_warnings.eu`) instead of the file under test, for `001_arg_mismatch.eu`:

- With the fault: `test_typecheck_001_arg_mismatch` **FAILED** — `eval path (... 001_arg_mismatch.eu ...) does not surface the warning eu check --strict reports`.
- Reverted the fault (one-line change back to `&path`): `cargo test --test harness_test test_typecheck` — **110 passed, 0 failed**.

Additionally, `004_invalid_annotation.eu` is itself a live, naturally-occurring instance of the exact regression class eu-ntwg.1 targets — the new assertion caught it on the first run, before I added the exclusion, which is direct evidence the mechanism works on a real (not just synthetic) divergence.

## Test plan

- [x] `cargo fmt --all`
- [x] `cargo test` — full suite green (500/500 harness tests, 110/110 typecheck tests with new coverage)
- [x] Fault injection as above (isolated to the new assertion, not the pre-existing one), reverted and confirmed clean
- [x] Filed eu-5q08 for the genuine bug this PR's coverage surfaced

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -708,6 +708,7 @@ dependencies = [
  "unicode-general-category",
  "url",
  "uuid 1.20.0",
+ "wait-timeout",
  "walkdir",
  "wasm-bindgen",
  "wasm-bindgen-test",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,6 +27,7 @@ proptest = "1"
 tempfile = "3"
 walkdir = "2"
 regex = "1"
+wait-timeout = "0.2"
 
 [dependencies]
 regex = "1.6.0"
--- a/tests/harness_test.rs
+++ b/tests/harness_test.rs
@@ -61,6 +61,52 @@ fn eu_binary() -> &'static Path {
     Path::new(env!("CARGO_BIN_EXE_eu"))
 }
 
+/// Run `cmd`, killing it and returning `None` if it doesn't complete
+/// within `deadline`.
+///
+/// Typecheck fixtures are check-only test material: some are not
+/// designed to terminate under plain evaluation at all (e.g.
+/// `092_self_assign_arg_pos_ok.eu`, `ones: cons(1, ones)` — a
+/// deliberate infinite lazy fixpoint used only to verify the checker
+/// doesn't flag argument-position self-reference), so a bounded wait is
+/// required rather than `Command::output()`'s indefinite block.
+fn run_with_deadline(
+    mut cmd: std::process::Command,
+    deadline: std::time::Duration,
+) -> Option<std::process::Output> {
+    use std::io::Read;
+    use wait_timeout::ChildExt;
+
+    let mut child = cmd
+        .stdout(std::process::Stdio::piped())
+        .stderr(std::process::Stdio::piped())
+        .spawn()
+        .expect("failed to spawn eu");
+
+    match child.wait_timeout(deadline).expect("wait_timeout failed") {
+        Some(status) => {
+            let mut stdout = Vec::new();
+            let mut stderr = Vec::new();
+            if let Some(mut s) = child.stdout.take() {
+                let _ = s.read_to_end(&mut stdout);
+            }
+            if let Some(mut s) = child.stderr.take() {
+                let _ = s.read_to_end(&mut stderr);
+            }
+            Some(std::process::Output {
+                status,
+                stdout,
+                stderr,
+            })
+        }
+        None => {
+            let _ = child.kill();
+            let _ = child.wait();
+            None
+        }
+    }
+}
+
 /// Run an error test — validates against `.expect` sidecar if present,
 /// otherwise passes as unvalidated.
 fn run_error_test(opt: &EucalyptOptions) {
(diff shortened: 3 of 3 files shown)


gmorpheme:

Verified independently in an isolated worktree (`/private/tmp/eu-wicket-1026`), rebased cleanly onto current master (post-#1023/#1024/#1025 — confirmed disjoint from all three: touches only `Cargo.toml`, `Cargo.lock`, `tests/harness_test.rs`).

**Fault injection, independently reproduced:** temporarily pointed `eval_cmd`'s path argument at `005_no_warnings.eu` instead of the file under test, for `001_arg_mismatch.eu` — `test_typecheck_001_arg_mismatch` correctly FAILED (`eval path (...) does not surface the warning eu check --strict reports`). Restored, confirmed `git diff` clean and `cargo test --test harness_test test_typecheck` — 110/110 pass. This is not a vacuous assertion.

**`wait-timeout` dependency:** confirmed it's under `[dev-dependencies]` in `Cargo.toml` (not `[dependencies]`), justified by `092_self_assign_arg_pos_ok.eu` being a genuine non-terminating fixpoint under plain eval — `Command::output()` would hang the test suite indefinitely on it.

**Cost claim sanity-checked on my own machine:** my absolute numbers don't match the PR's (my release-build baseline for `cargo test --test harness_test` — 500 tests including the embedded typecheck tests — runs in ~10.7s clean, vs. the PR's ~70s figure; likely a debug-vs-release or machine difference, not something to chase down for a test-infrastructure PR). What matters is the *added* cost: isolated it directly by disabling the new eval-path block (temporary early `return`) and comparing — two consecutive baseline (disabled) runs were rock-solid at 10.75s/10.76s. Runs with the block enabled were noisier on first attempts (one high outlier at 17.8s, contended by unrelated background activity — no other cargo/rustc processes were running per `ps aux`, so likely macOS scheduling/indexing noise), but three controlled back-to-back enabled runs settled to 10.66s/10.96s/10.67s, essentially matching baseline. Consistent with the PR's proportional claim of a small, not-worth-chasing overhead.

**`004_invalid_annotation.eu` exclusion:** confirmed the comment in `tests/harness_test.rs` (lines ~195-208) cites eu-5q08 explicitly and explains the root cause (the three `parse_scheme(..).ok()?` call sites in `core/typecheck/check.rs`) accurately. Cross-checked against `bd show eu-5q08` — the bead exists, is P1, in progress, and its description matches the PR's summary verbatim (same three call sites, same root-cause trace, same "excluded here, not silently" framing).

**Deferred warning-absence follow-up (75 checker-clean fixtures, ANSI-colour splitting issue): NOT currently tracked as a bead.** I searched (`bd search "warning absence"`, `"ANSI"`, `"checker-clean"`) and found nothing. Flagging this per the review brief — please file it if you want it tracked; I have not filed it myself since bead creation/closure calls are the coordinator's per project convention.

**Other checks:**
- `cargo fmt --all -- --check` — clean. `cargo clippy --all-targets -- -D warnings` — clean.
- Full `cargo test --release` — 1306 lib + 500 harness (110 typecheck tests included, with the new eval-path assertions) + all other suites green, 0 failures.
- CI fully green across all required checks. Master green after the #1025 merge (all quality-gate jobs passed; confirmed before merging this PR).

No design deviations found; this closes eu-ntwg.1 as scoped. Merging.

PR #1027, opened 2026-07-16
fix: surface malformed type annotations as warnings on the eval path (eu-5q08)

gmorpheme:
## Summary

Fixes eu-5q08 (P1): a malformed `type:` annotation was a hard error under
`eu check --strict` but was **silently discarded** (exit 0, no diagnostic
at all) under plain `eu` evaluation, even though type checking runs
unconditionally on every `eu` invocation per the documented contract.

Repro:
```
` { type: "number ->" }
bad(x): x
```
`eu check --strict` correctly rejected this; plain `eu` printed `{}` and
exited 0, silently treating the malformed annotation as though it were
absent.

## Root cause

Three call sites discarded `parse::parse_scheme`'s `Err` via `.ok()?`,
converting a genuine parse failure into "no annotation present":

- `extract_annotation` (`src/core/typecheck/check.rs`) — used for
  ordinary declaration `type:`/`__type_hint:` annotations. `eu check`'s
  annotation-syntax pre-pass (`check_annotation_syntax` in
  `src/driver/check.rs`) is a *separate* walk over the raw AST that
  correctly validates annotations, but the eval path never runs it — it
  relies entirely on the bidirectional `Checker`, which had no equivalent
  validation.
- `parse_operator_overloads` (`src/core/typecheck/check.rs`) — used for
  **operator** definitions specifically. Cook's `distribute_fixities`
  strips the `Meta` wrapper (and its `type:` annotation) from operator
  bindings before the checker's normal per-declaration walk ever sees
  them, so a malformed annotation on an operator (`` `{type: "..."} (l +++ r): ... ``)
  was *only* ever reachable through this call site — fixing
  `extract_annotation` alone would not have covered it.

## Fix

Both sites now push a `TypeWarning` (`invalid type annotation: <parse
error>`) into the checker's normal warning pipeline instead of discarding
the error, so the diagnostic:

- appears on stderr under plain `eu` (previously invisible)
- is promoted to a hard error under `--strict` (previously already an
  error only via the separate `eu check` pre-pass)
- is silenced by `--suppress-type-warnings`, like any other warning
- matches `eu check`'s message content (`invalid type annotation: type
  annotation parse error at position N: ...`), so the two paths agree

**Deduplication**: `extract_annotation` is called multiple times per
malformed annotation in normal control flow (once pre-seeding a `Let`
frame, once during synthesis, and again for `type-def:`/`result-def:`
alias extraction) — a naive fix would emit the same warning 2-3 times for
one bad annotation. Added a `Smid`-keyed `HashSet` on `Checker` so each
distinct annotation location is reported exactly once, verified manually
(single warning line, confirmed with `eu` output below).

For `parse_operator_overloads` (a free function outside `Checker`, only
ever called once per operator per compile — no duplicate-visit risk), I
extended `UnitInterface::operator_type_strings()` to carry each
operator's `Smid` (already captured in `OperatorInfo`, just not
threaded through) alongside its annotation string, so the new warning
gets a real source location instead of `Smid::default()`.

**Prelude stays clean**: `cargo xtask prelude-compile` produces a
byte-identical `lib/prelude.blob` (confirmed via `git status`), and the
prelude-interface cache build in `driver/check.rs` discards the new
operator-annotation warnings the same way it already discarded the
primary `type_check_for_prelude` warnings — a per-process cache build has
nothing useful to report per-invocation, and the prelude has no malformed
annotations to report.

## Verification

```
$ eu tests/harness/typecheck/004_invalid_annotation.eu --heap-limit-mib 2048
warning: invalid type annotation: type annotation parse error at position 9: expected a type, got Eof
  ┌─ tests/harness/typecheck/004_invalid_annotation.eu:1:3
  │
1 │ ` { type: "number ->" }
  │   ^^^^^^^^^^^^^^^^^^^^^

---
{}
exit: 0

$ eu tests/harness/typecheck/004_invalid_annotation.eu --heap-limit-mib 2048 --strict
(same warning) exit: 1

$ eu tests/harness/typecheck/004_invalid_annotation.eu --heap-limit-mib 2048 --suppress-type-warnings
---
{}
exit: 0

$ eu check --strict tests/harness/typecheck/004_invalid_annotation.eu
tests/harness/typecheck/004_invalid_annotation.eu:1:11: error: invalid type annotation on 'bad': ...
(same warning, once) exit: 1
```

Also verified end-to-end for the operator call site with an ad hoc
`` `{type:"number ->"} (l +++ r): l + r `` fixture — warning surfaces
correctly with the operator's own source location.

**Engine parity**: confirmed identical output on both engines —
`EU_HEAPSYN=1 eu tests/harness/typecheck/004_invalid_annotation.eu
--heap-limit-mib 2048` produces the same single warning as the default
bytecode engine.

## Regression test

`004_invalid_annotation.eu` is excluded from #1026's new eval-path
parity assertion, with a comment citing this bead — #1026 was still
**open, not merged**, at the time of this PR, so per the dispatch
instructions I added an equivalent standalone regression test now rather
than depending on #1026 landing first:

`tests/harness_test.rs::test_typecheck_004_invalid_annotation_eval_path_warns`
— runs plain `eu` (no subcommand) on the fixture and asserts stderr
contains `"invalid type annotation"`.

**Fault-injection verified**: reverted `extract_annotation`'s fix back to
`parse::parse_scheme(&type_str).ok()` (the pre-fix discard behaviour) —
`test_typecheck_004_invalid_annotation_eval_path_warns` **FAILED**
(`eval-path stderr for 004_invalid_annotation.eu does not contain
"invalid type annotation"`) while the pre-existing
`test_typecheck_004_invalid_annotation` (`eu check --strict` test)
**still passed**, confirming the new test is exercising exactly the gap
this bead describes and not duplicating existing coverage. Restored the
fix — both tests pass again.

**Coordination note for the gatekeeper**: once #1026 merges, its
`004_invalid_annotation.eu` exclusion (in its version of
`run_typecheck_test`) should be removed so the eval-path assertion covers
this fixture through the general mechanism too. My standalone test here
can stay (it's a narrower, independently-gating regression test) or be
removed as redundant once #1026's broader coverage lands — gatekeeper's
call on sequencing.

## Test plan

- [x] `cargo fmt --all`
- [x] `cargo test` — full suite green: 501/501 harness tests (500 + 1
      new), 111 typecheck tests, 111 LSP tests, 11 property tests, 1 tick
      parity test, 3 doc tests
- [x] `cargo xtask prelude-compile` — clean, byte-identical blob (no
      diff)
- [x] Fault-injection verified (see above)
- [x] Engine parity verified (bytecode + `EU_HEAPSYN=1`) for the repro
      file

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to eucalypt are documented here.
 
 ## [Unreleased]
 
+### Fixed
+
+- **A malformed `type:` annotation was a hard error under `eu check --strict` but silently discarded (exit 0, no diagnostic) under plain `eu` evaluation** — despite type checking running unconditionally on every invocation (per the documented contract). Three call sites in `core/typecheck/check.rs` and `driver/check.rs` (`extract_annotation`, used for ordinary declarations, and `parse_operator_overloads`, used for operator definitions whose `Meta` wrapper is stripped before the checker's normal per-node walk ever sees it) discarded the `parse_scheme` parse error via `.ok()?`, treating the annotation as though it were simply absent. Parse failures now surface as an `invalid type annotation: <parse error>` `TypeWarning`, flowing through the same unconditional warning pipeline as every other type diagnostic — visible on stderr under plain `eu`, promoted to a hard error under `--strict`, silenced by `--suppress-type-warnings`. Deduplicated by source location (`Smid`) so the same malformed annotation, which the checker's `Let` pre-seeding and synthesis passes each visit independently, is not reported twice. The prelude remains warning-free (`cargo xtask prelude-compile` unaffected; the prelude's cached type-check path discards operator-annotation warnings the same way it already discarded its primary check warnings). Regression-tested via `tests/harness_test.rs`'s `test_typecheck_004_invalid_annotation_eval_path_warns`, fault-injection verified (eu-5q08)
+
 ## [0.13.0] - Unreleased
 
 ### Added
--- a/src/core/typecheck/check.rs
+++ b/src/core/typecheck/check.rs
@@ -23,7 +23,7 @@
 //!
 //! Type issues are always warnings — they never prevent evaluation.
 
-use std::collections::{BTreeMap, HashMap, VecDeque};
+use std::collections::{BTreeMap, HashMap, HashSet, VecDeque};
 
 use serde::{Deserialize, Serialize};
 
@@ -346,6 +346,17 @@ pub struct Checker {
     /// `recognise_brancher` and `classify_lam_body` when the head
     /// expression is a `Var::Free` in prelude-cached user-file checks.
     seed_branch_shapes: HashMap<String, BranchShape>,
+
+    /// `Smid`s of `type:`/`__type_hint:` annotations whose syntax has
+    /// already produced an `invalid type annotation` warning (eu-5q08).
+    ///
+    /// `extract_annotation` is invoked more than once per malformed
+    /// annotation in normal control flow — once while pre-seeding a `Let`
+    /// frame (`annotation_scheme_of`), once during synthesis
+    /// (`synthesise_meta`), and again if the binding also carries
+    /// `type-def:`/`result-def:` metadata — so this set prevents the same
+    /// parse failure from being reported more than once.
+    warned_annotation_smids: HashSet<Smid>,
 }
 
 impl Default for Checker {
@@ -370,6 +381,7 @@ impl Checker {
             keep_root_scope: false,
             seed_branch_shapes: HashMap::new(),
             operator_overloads: HashMap::new(),
+            warned_annotation_smids: HashSet::new(),
         }
     }
 
@@ -887,7 +899,7 @@ impl Checker {
     /// - `is_hint`: came from `__type_hint` (monad binding) — the
     ///   annotation is for checking only, not for overriding the
     ///   synthesised type
-    fn extract_annotation(&self, meta: &RcExpr) -> Option<(Type, Vec<Constraint>, bool, bool)> {
+    fn extract_annotation(&mut self, meta: &RcExpr) -> Option<(Type, Vec<Constraint>, bool, bool)> {
         let block = match &*meta.inner {
             Expr::Block(_, b) => b,
             _ => return None,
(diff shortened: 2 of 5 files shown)


gmorpheme:

Verified independently in an isolated worktree (`/private/tmp/eu-wicket-1027`), rebased cleanly onto current master (post-#1025/#1026 — the rebase onto `tests/harness_test.rs`, which #1026 also touches, applied without conflict).

**(a) Operator gap, reproduced myself:** built a plain-master baseline (no #1027) alongside this branch. On master, `` `{type: "number ->"} (l +++ r): l + r `` is silent (exit 0, no diagnostic). On this branch, it warns exactly once with a real source location pointing at the operator's own metadata block. Confirmed identical output under `EU_HEAPSYN=1`.

**(b) Dedup set:** two distinct malformed annotations in one file (`` `{type:"number ->"}`` and `` `{type:"string ->>"}`` on different bindings) produce two distinct warnings, each with its own correct location and distinct parse-error text — not over-suppressed. A single malformed annotation produces exactly one warning line, not 2-3 duplicates (confirmed by grep-count on output).

**(c) Prelude blob byte-identity:** compiled the blob on both this branch and the plain-master baseline (`cargo run --package xtask -- prelude-compile`) and diffed the checksums — identical (`76b5b79246f22b87c83f9f4cb0b202b7` both sides).

**(d) `--strict`/`--suppress-type-warnings`:** confirmed `--strict` promotes the new warning to a hard error (exit 1, evaluation aborted before running); `--suppress-type-warnings` silences it entirely (exit 0, normal output).

**Fault injection, independently reproduced:** reverted `extract_annotation` to the pre-fix `.ok()?` discard — `test_typecheck_004_invalid_annotation_eval_path_warns` correctly FAILED while the pre-existing `test_typecheck_004_invalid_annotation` (`eu check --strict` test) still PASSED, confirming the new test isn't duplicating existing coverage. Restored, both pass.

**Call-site audit:** confirmed all three real callers of `parse_operator_overloads` (`build_prelude_interface_for`, `run_type_checker`, `run_type_checker_from_blob_core` in `src/driver/check.rs`) correctly handle the new `(HashMap, Vec<TypeWarning>)` return — the prelude cache build explicitly discards the operator warnings with a clear comment (matching how it already discarded primary warnings), the other two extend `warnings`. The dead, never-called `fixity::extract_operator_type_strings` (pre-existing, unrelated to this PR) is not a bypass path.

**Full `cargo test` (matching CI's exact "Test Suite" invocation, no blob) — clean, 0 failures**, 1306 lib + 501 harness (500 + the new regression test) + all other suites. `cargo fmt --all -- --check` / `cargo clippy --all-targets -- -D warnings` — both clean.

**Important finding, unrelated to this PR — filed as eu-rqwh (P1):** while re-verifying with a compiled blob present (to check claim (c)), I found `cargo test --release --test harness_test` fails 5 pre-existing typecheck fixtures (`010`, `011`, `013`, `063`, `085` — all `:for`/`:io` monadic-binding fixtures) when the blob is present, but passes cleanly without it. Confirmed this is **not caused by #1027** — identical failure reproduces on plain master (just #1025+#1026) with the blob present, and identical pass without it, regardless of whether #1027's changes are applied. It's a genuine, pre-existing blob-vs-source checker divergence for monadic bindings (distinct from the original 065 regression — that fixture's own regression test still passes fine with the blob), invisible to CI because the "Test Suite" job never compiles a blob and the blob-harness jobs run the eucalypt-level `.eu` test framework, not this Rust test binary. Filed as its own P1 bead with the full repro; does not block this PR.

No design deviations found. This closes eu-5q08 as scoped (both call sites fixed). Merging.

PR #1028, opened 2026-07-16
chore: re-include 004_invalid_annotation in eval-path parity check (eu-5q08)

gmorpheme:
## Summary

PR #1026 (eu-ntwg.1) added a general eval/check parity assertion to `run_typecheck_test`: for any `tests/harness/typecheck` fixture whose `.expect` sidecar declares a non-empty `stderr:` warning pattern, the test now also runs the plain `eu` eval path and asserts the same pattern appears. `004_invalid_annotation.eu` was excluded from that assertion at the time because PR #1027 (eu-5q08) — which fixes the underlying divergence — hadn't merged yet.

#1027 has since merged: `extract_annotation` in `src/core/typecheck/check.rs` now emits an `invalid type annotation` `TypeWarning` instead of silently discarding the `parse_scheme` error via `.ok()?`. The planned follow-up — removing the 004 exclusion so the general mechanism covers it — never happened. This PR does that.

## Changes

- Removed the `if filename == "004_invalid_annotation.eu" { return; }` early-return in `run_typecheck_test` (`tests/harness_test.rs`). 004's `.expect` sidecar has a non-empty `stderr: "invalid type annotation"` pattern, so it now genuinely joins the eval-path-parity-covered set (36 warning-bearing fixtures instead of 35).
- Folded away `test_typecheck_004_invalid_annotation_eval_path_warns` and its `run_typecheck_eval_path_test` helper. **Rationale**: with the exclusion gone, this standalone test is fully redundant with the general mechanism — same fixture, same expected pattern (`"invalid type annotation"`), essentially the same assertion (run the eval path, check stderr contains the pattern). The original doc comment on the helper explicitly said it existed "so eu-5q08's regression is gated on its own even if #1026 has not yet merged" and "should be removed so the two mechanisms do not silently overlap" once #1026 landed. `test_typecheck_004_invalid_annotation` (which calls the now-uncapped `run_typecheck_test`) covers both the `eu check --strict` path and the eval path, so I kept that as the single source of truth and added a doc comment on it noting it also gates the eu-5q08 regression.
- Tidied the now-outdated sequencing comments describing the #1026/#1027 merge dance (they're gone along with the removed helper/test, since that dance is now resolved).

## Fault injection (mandatory verification)

To confirm the re-included assertion genuinely gates:
1. Temporarily reverted `extract_annotation` in `src/core/typecheck/check.rs` to the pre-#1027 behaviour (`parse::parse_scheme(&type_str).ok()?`, discarding the parse error instead of calling `warn_invalid_annotation`).
2. Ran `cargo test --test harness_test test_typecheck_004` — **FAILED** as expected:
   ```
   eval path (`eu 004_invalid_annotation.eu`, no subcommand) does not surface the warning
   `eu check --strict` reports for the same file — checker/eval divergence (eu-ntwg.1).
   Expected to find "invalid type annotation" in stderr:
   ```
3. Restored the fix, confirmed `git diff` on `check.rs` was clean, re-ran the test — **PASSED**.

## Test plan

- [x] `cargo test` — full suite green (500 passed in `harness_test.rs`, down from the prior 501 by exactly the one folded test; no other regressions)
- [x] `cargo test --test harness_test test_typecheck_004` — passes, now exercising the eval-path parity assertion for 004
- [x] `cargo fmt --all` — no diff
- [x] Fault injection performed and documented above

### DIFF
--- a/tests/harness_test.rs
+++ b/tests/harness_test.rs
@@ -196,20 +196,14 @@ fn run_typecheck_test(filename: &str) {
     // --strict` genuinely warns (a non-empty expected pattern) — see the
     // doc comment above for why the other fixtures are out of scope.
     //
-    // 004_invalid_annotation.eu is a KNOWN, tracked exception (eu-5q08):
-    // this new check caught a genuine, previously-invisible divergence —
-    // `eu check`'s annotation-syntax validation (a pre-pass unique to the
-    // `check` subcommand) rejects a malformed `type:` annotation string,
-    // but the type checker used on the plain-eval path silently discards
-    // the same parse failure (`parse_scheme(..).ok()?` at three call
-    // sites in core/typecheck/check.rs) and treats it as "no annotation".
-    // Fixing that is a type-checker change, out of scope for this harness
-    // PR; excluded here (not silently — see eu-5q08) so the other 35
-    // warning-bearing fixtures still gate.
-    if filename == "004_invalid_annotation.eu" {
-        return;
-    }
-
+    // 004_invalid_annotation.eu used to be excluded here (eu-5q08): `eu
+    // check`'s annotation-syntax pre-pass rejected a malformed `type:`
+    // string, but the plain-eval-path checker silently discarded the same
+    // parse failure via `parse_scheme(..).ok()?`, treating it as "no
+    // annotation". That divergence is fixed (`extract_annotation` in
+    // `core/typecheck/check.rs` now emits an `invalid type annotation`
+    // warning), so 004 is covered by this general mechanism like every
+    // other warning-bearing fixture.
     if let Some(pattern) = expected_stderr.as_deref().filter(|p| !p.is_empty()) {
         let mut eval_cmd = std::process::Command::new(eu_binary());
         eval_cmd.args(["--heap-limit-mib", "2048", &path]);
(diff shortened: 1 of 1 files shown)


gmorpheme:
**Wicket review — approved.**

Reviewed in an isolated worktree (`/private/tmp/eu-wicket-1028`, PR head `a3fdde46` off master `4a42131b`).

**Diff scope confirmed**: single file, `tests/harness_test.rs`, test infrastructure only — matches the PR description exactly. No production code touched.

**Independent verification performed:**

1. `.expect` sidecar check: `tests/harness/typecheck/004_invalid_annotation.eu.expect` has `exit: 1` and `stderr: "invalid type annotation"` — a genuine non-empty pattern, so removing the exclusion causes 004 to actually join the parity-checked set rather than being vacuously included.

2. No orphaned references: `grep -rn run_typecheck_eval_path_test tests/` after the diff returns nothing — the folded helper has no other callers, confirming the standalone test was safe to remove.

3. **Fault injection (independently repeated, not taking the author's word for it)**: reverted `extract_annotation` in `src/core/typecheck/check.rs` to the pre-#1027 `parse::parse_scheme(&type_str).ok()?` discard.
   - Rebuilt release, ran `cargo test --release --test harness_test test_typecheck_004` → **FAILED**: `eval path (\`eu 004_invalid_annotation.eu\`, no subcommand) does not surface the warning \`eu check --strict\` reports for the same file — checker/eval divergence (eu-ntwg.1)`.
   - Restored the fix (`git checkout -- src/core/typecheck/check.rs`), rebuilt, re-ran → **PASSED**.
   - This confirms the general `run_typecheck_test` eval-path-parity mechanism genuinely gates the eu-5q08 regression through `test_typecheck_004_invalid_annotation` — it is not a no-op assertion.

4. Full suite: `cargo test --release --test harness_test` → 500 passed, 0 failed (down from the pre-PR baseline of 501 by exactly the one folded test, as claimed).

5. `cargo fmt --all --check` → clean. `cargo clippy --all-targets -- -D warnings` → clean.

6. CI: all 20 checks on the PR green (release/prepare-release jobs correctly skipping — not a release PR). Master CI green both before and after this review.

No architectural concerns — this is a straightforward, low-risk consolidation of duplicated test coverage now that the sequencing dependency between #1026 and #1027 has resolved. Comments are clear and the doc-comment on `test_typecheck_004_invalid_annotation` correctly notes it now also gates eu-5q08. Not a GC/memory/unsafe/blob-format/engine-default/release-machinery change, so the recorded-review gate doesn't strictly apply here, but this comment serves the same purpose.

Merging.

PR #1029, opened 2026-07-16
fix(heapsyn): restore caller state before propagating sub-run error (eu-dy96)

gmorpheme:
## Summary

- HeapSyn's `evaluate_to_whnf_impl` (used by BIF-triggered `force`, e.g. `debug::boxed_native`'s inner-field force, `block::collect_block_keys`) propagated a sub-run error via `run_result?` **before** restoring the caller's `closure`/`stack`. A BIF that swallows the force error (`debug.rs boxed_native`'s `.ok()`, `block.rs collect_block_keys`'s `Err(_) =>`) would then resume stepping on the truncated sub-run state instead of the caller's — the exact latent ordering bug eu-apn7 fixed on the bytecode engine (`exit_whnf_subrun` in `src/eval/bytecode/machine.rs`: restore state, *then* propagate).
- While fixing the bead's reported function, found the identical bug pattern in the sibling `Machine::evaluate_to_whnf` (used by `evaluate_to_whnf_for_io` to force IO spec-block fields) — fixed both with the same restore-before-propagate reordering.
- Not observable in the current corpus, but HeapSyn is the differential-testing oracle until Phase 4 (eu-oufc); a caller-state corruption bug in the oracle undermines byte-identical validation against the bytecode engine.

## Regression tests

Two Rust unit tests in `src/eval/machine/vm.rs` (`eval::machine::vm::tests`), mirroring eu-apn7's own test structure (a direct unit test, not a harness `.eu` test):
- `test_evaluate_to_whnf_impl_restores_caller_state_before_propagating_error`
- `test_machine_evaluate_to_whnf_restores_caller_state_before_propagating_error`

Each sets up a "caller" state (a marker continuation on the stack, a distinguishable closure), drives an intentionally-erroring sub-evaluation (an out-of-range local reference), and asserts that after the `Err` propagates, the caller's stack/closure were restored rather than left as the sub-run's leftover state.

**Fault-injection verified**: reverted each fix individually (temporarily restoring the buggy `run_result?`-before-restore ordering) and confirmed its dedicated test failed while the sibling (still-fixed) test passed; restored the fix and confirmed both pass again.

I also explored an `.eu`-level harness-test trigger (exercised by both engines via the existing `cargo test` / `EU_HEAPSYN=1 cargo test` gate) but did not land one: reaching `__EXPECT`'s failure-render path with `actual` already forced to outer WHNF but with its inner field still lazy and erroring — without the `pass = e = v` computation itself forcing (and erroring on) that same inner field first — proved impractical through the standard library's `=`/type-check intrinsics. Every construction tried either forced the inner content directly (propagating the error uncaught, before `__EXPECT.execute` even runs) or left the outer value unforced (`"<unevaluated>"`, never reaching `boxed_native`'s swallow at all).

## Test status

- `cargo test`: 501/501 harness, 1308/1308 lib, all other suites green (default bytecode engine).
- `EU_HEAPSYN=1 cargo test`: 501/501 harness, 1308/1308 lib, all green.
- `EU_GC_VERIFY=2 EU_GC_STRESS=1 cargo test --test harness_test`: 501/501, on both engines.
- `cargo fmt --all`: clean.

## Notes for review

This PR touches engine internals (VM sub-evaluation state machinery), so the recorded-review rule applies — owner reviews personally per Furnace's workflow for 0.13.1.

### DIFF
--- a/src/eval/machine/vm.rs
+++ b/src/eval/machine/vm.rs
@@ -1715,16 +1715,20 @@ fn evaluate_to_whnf_impl(
         .pop()
         .expect("suspended_stacks underflow in evaluate_to_whnf");
 
-    // Propagate run error only after restoring state.
-    run_result?;
-
     let sub_yielded = state.yielded_io;
     let result = state.closure.clone();
 
+    // Restore the caller's state BEFORE propagating any run error (eu-dy96,
+    // mirrors eu-apn7's fix on the bytecode engine): a BIF that swallows a
+    // sub-run force error (e.g. `debug::boxed_native`'s `.ok()`,
+    // `block::collect_block_keys`'s `Err(_) =>`) must resume stepping on the
+    // intact caller stack, not a truncated/leftover sub-run stack.
     state.terminated = false;
     state.yielded_io = false;
     state.closure = saved_closure;
     state.stack = saved_stack;
+
+    run_result?;
 
     if sub_yielded {
         return Err(ExecutionError::Panic(
@@ -2448,19 +2452,21 @@ impl<'a> Machine<'a> {
             .pop()
             .expect("suspended_stacks underflow in evaluate_to_whnf");
 
-        // Propagate any run error only after retrieving the GC-safe copies,
-        // to leave the machine in a consistent state.
-        run_result?;
-
         // Capture the result before restoring state.
         let sub_yielded = self.state.yielded_io;
         let result = self.state.closure.clone();
 
-        // Restore the outer evaluation state.
+        // Restore the outer evaluation state BEFORE propagating any run
+        // error (eu-dy96, mirrors eu-apn7's fix on the bytecode engine and
+        // the identical latent bug in the sibling `evaluate_to_whnf_impl`
+        // above): a caller that swallows a sub-run error must resume on the
+        // intact caller stack, not a truncated/leftover sub-run stack.
         self.state.terminated = false;
         self.state.yielded_io = false;
         self.state.closure = saved_closure;
         self.state.stack = saved_stack;
+
+        run_result?;
 
         if sub_yielded {
             return Err(ExecutionError::Panic(
@@ -2613,6 +2619,7 @@ pub mod tests {
     use std::cell::RefCell;
     use std::rc::Rc;
 
+    use crate::eval::machine::cont::Continuation;
     use crate::eval::machine::env::{EnvFrame, SynClosure};
     use crate::eval::machine::intrinsic::StgIntrinsic;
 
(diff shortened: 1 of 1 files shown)


gmorpheme:
**Wicket review — approved and merging.** Engine-internals PR (`src/eval/machine/vm.rs`), so the recorded-review gate applies; this comment is that review.

Reviewed in `/private/tmp/eu-wicket-1029` (head `d9591290` off master `4a42131b`).

**Precedent cross-check**: compared the fix against the eu-apn7 bytecode-engine precedent it claims to mirror (`enter_whnf_subrun`/`exit_whnf_subrun` in `src/eval/bytecode/machine.rs:520-578`). Confirmed both fixed sites in `vm.rs` (`evaluate_to_whnf_impl` and `Machine::evaluate_to_whnf`) now follow the identical choreography: pop the stashed caller state, fully restore `terminated`/`yielded_io`/`closure`/`stack`, *then* propagate `run_result?`, *then* check `sub_yielded`. This is a faithful, correctly-scoped port of an already-proven invariant, not a new design.

**Independent fault injection, both sites (not taking the author's word for it):**
1. Reverted only `evaluate_to_whnf_impl`'s ordering (moved `run_result?` back above the state restoration). Rebuilt release, ran both new unit tests: `test_evaluate_to_whnf_impl_restores_caller_state_before_propagating_error` **FAILED** (`left: 0, right: 1` — sub-run's empty stack observed instead of the caller's 1-entry stack), while the sibling `test_machine_evaluate_to_whnf_restores_caller_state_before_propagating_error` still **PASSED** — confirms the two tests are independently gating, not accidentally coupled.
2. Restored, then independently reverted only `Machine::evaluate_to_whnf`'s ordering. Rebuilt, ran both tests: this time the sibling test **FAILED** with the identical symptom while `evaluate_to_whnf_impl`'s own test **PASSED** — confirms the isolation holds symmetrically in both directions.
3. Restored the fix fully (`git checkout -- src/eval/machine/vm.rs`), rebuilt, confirmed both tests **PASS** again.

**Full suite, independently run:**
- `cargo test --release` (default bytecode engine): all green, including 501/501 `harness_test.rs` and the two new unit tests.
- `cargo test --release --test harness_test` under `EU_HEAPSYN=1`: 501/501.
- `EU_GC_VERIFY=2 EU_GC_STRESS=1 cargo test --release --test harness_test`: 501/501, both without and with `EU_HEAPSYN=1`.
- `cargo fmt --all --check`: clean. `cargo clippy --all-targets -- -D warnings`: clean.
- CI: all required checks green (release/prepare-release jobs correctly skipping, not a release PR).

**On the documented `.eu`-level trigger attempt**: the PR's reasoning for why an end-to-end harness trigger proved impractical holds up — every construction either forces the erroring inner field directly via `=`'s own type-check/equality machinery (propagating before `__EXPECT.execute` even runs) or leaves the outer value unforced entirely (never reaching `boxed_native`'s `.ok()` swallow). Given the two independently-verified, fault-injection-gated unit tests already isolate the exact invariant at the point where it can be checked precisely, I don't think a fragile `.eu`-level trigger would add real coverage — the unit-test approach (mirroring eu-apn7's own precedent) is the right call here.

No architectural concerns. Merging.

PR #1024, opened 2026-07-16
chore: tighten tick-parity tripwire caps, retire stale handicap narrative (eu-vcr8)

gmorpheme:
## Summary

Owner ruling 2026-07-16 (recorded on `eu-vcr8`): the source-prelude tick handicap `tests/tick_parity_test.rs` was written to bound is **closed**. Bead `eu-npp9` measured `fib(30)` directly on current master (post-#1016) and found full parity, not merely improvement:

- bytecode: blob 88,853,753 vs. `--source-prelude` 88,853,755 ticks (ratio 1.0000000225)
- HeapSyn: blob 115,779,125 vs. `--source-prelude` 115,779,127 ticks (ratio 1.0000000173)

The residual +2-tick gap is Smid/annotation bookkeeping, not an algorithmic divergence. Likely closing mechanism (per `eu-npp9`'s own close notes): PR #1016 (`eu-rb5n`)'s `analyse_demand` Meta-peel fix — a side effect of work that targeted a separate HOF-combinator source-path fusion gap, but which also fixed demand-signature recording for the arithmetic/comparison intrinsics `fib`'s naive recursion depends on (every prelude binding, including those, carries a `Meta(Lam, {doc, type})` wrapper the old code didn't see through).

## Changes

- Halved both caps per the owner ruling: `MAX_SOURCE_PRELUDE_RATIO` 1.12 → 1.06, `MAX_SOURCE_PRELUDE_RATIO_PREDECODE` 1.13 → 1.065.
- Rewrote the module doc comment and both constants' doc comments: the ~10.6% handicap narrative is now framed as closed history (kept, for future-debugging value, as an explanation of *why* the handicap existed and what fixed it) rather than a current documented allowance. The caps are now pure regression tripwires against reintroducing the demand-analysis divergence — with the measured ratio at ~1.00000002, even the halved caps leave roughly six orders of magnitude of headroom over the actual noise floor.
- No change to the test logic itself (`run_ticks`, the predecode-cap-selection branch, the assertion) — only the constants and their surrounding narrative.

## Fault-injection verification

Temporarily set each cap to `1.0000000001` (below the measured ratio) in turn:

- `MAX_SOURCE_PRELUDE_RATIO` fault: `cargo test --test tick_parity_test` failed with `default=722435 ticks, --source-prelude=722437 ticks, ratio=1.0000 > 1.0000000001 (predecode=false)`.
[...]

PR #1031, opened 2026-07-16
feat(bytecode): flip EU_PREDECODE default to pre-decoded dispatch (eu-vcr8, Lever a Phase 2)

gmorpheme:
## Summary

Lever (a) Phase 2 (eu-vcr8): flips `EU_PREDECODE`'s default so pre-decoded
dispatch runs unless explicitly disabled, mirroring how BV1 itself flipped
from opt-in bytecode to opt-out `EU_HEAPSYN`. Both entry conditions recorded
on eu-vcr8 are resolved: the tick-parity tripwire caps were halved to
1.06/1.065 (#1024, merged), and the 3-mode differential gate was built
(#1023, merged, eu-ntwg.3 closed).

Phase 3 (byte-dispatch-path deletion) is explicitly **not** in this PR — the
byte path is retained, soaked, and CI-tested through this release; it is
deleted only after the soak period, per the design's Step D
(`docs/superpowers/specs/2026-07-13-predecoded-execution-ir-design.md` §6).

## Post-flip env-var semantics

- unset or `EU_PREDECODE=1` → pre-decoded dispatch (**new default**)
- `EU_PREDECODE=0` → explicit opt-out, selects the retained byte-dispatch path
- `EU_HEAPSYN=1` → takes precedence over both; `EU_PREDECODE` is a no-op under HeapSyn

`src/eval/bytecode/mod.rs`'s `predecode_enabled()` changed from
`== Ok("1")` to `!= Ok("0")`. `CLAUDE.md`'s Debug Environment Variables table
and `docs/architecture.md`'s pre-decoded-IR section are updated to describe
the new default and the opt-out. A new CI job, `test-byte-dispatch-baseline`
(mirroring `test-heapsyn-baseline`), runs the full harness under
`EU_PREDECODE=0` so the retained byte path doesn't rot during the soak.

`tests/tick_parity_test.rs`'s cap-selection logic previously checked
`EU_PREDECODE == "1"` to pick between the byte-path and predecode tripwire
caps — that would have silently selected the wrong cap once the default
flipped (unset now means predecode, not byte-path). Fixed to mirror the real
precedence (`EU_HEAPSYN` first, then `EU_PREDECODE != "0"`).

## Soak evidence

- `cargo fmt --all`, `cargo clippy --all-targets -- -D warnings`: clean
- Full `cargo test` under the new default (predecode on): 0 failures
- Full `cargo test` under `EU_HEAPSYN=1`: 0 failures
- Full `cargo test` under `EU_PREDECODE=0` (byte-path opt-out): 0 failures
- 3-mode differential gate (`bytecode::differential`): 26/26 pass
- Tick-parity tripwire (halved 1.06/1.065 caps), against a genuinely
  blob-embedded binary (`cargo xtask prelude-compile` first), in all three
  configs (default/predecode, `EU_PREDECODE=0`, `EU_HEAPSYN=1`): all green
- `EU_GC_VERIFY=2 EU_GC_STRESS=1 EU_GC_POISON=1` full harness suite
  (183/183 tests) in all three configs: green, exit 0 in each
- Canonical 8-bench suite via `cargo xtask engine-ab` (new row appended,
  commit `921f3102`) + `--check` against the previous ledger row
  (2026-07-14, commit `5f550b49`). **Confidence: measured-single** — the
  session was interleaved bc/hs per-bench (protocol-compliant sampling), but
  the machine's 1-minute load average was 3.6–7.9 throughout (a Parallels VM
  and `rust-analyzer` competing for CPU, not agent activity), never reaching
  the protocol's `<1` bar, so this does not qualify as *measured-verified*.

  | bench | class | ratio (bc/hs) | vs prev | status |
  |---|---|---|---|---|
  | block_merge | D | 0.786 | −4.5% | ok |
  | import_export_yaml | I | 0.950 | −0.6% | ok |
  | import_export_toml | I | 0.949 | +2.7% | ok |
  | string_scale | G | 0.814 | **+50.5%** | REGRESSED (see below) |
  | list_scale | H | 0.691 | −36.3% | ok |
  | lookup_curve | E | 11.139 | −18.8% | WATCH (tripwire, by design — class E's whole purpose is measuring the bc/hs asymmetry, not a pass/fail target) |
  | io_loop | L | 0.963 | −1.4% | ok |
  | hof_fold | C | 0.965 | **+26.0%** | REGRESSED (see below) — **also now sub-200ms, see caveat** |

  **The two REGRESSED flags are not caused by this PR.** Root-caused via
  `hs_ticks` (deterministic, and `EU_PREDECODE` cannot affect HeapSyn's
  dispatch at all): `hof_fold`'s `hs_ticks` dropped 52.2M → 2.18M and
  `string_scale`'s dropped 76.8M → 46.0M between the ledger's previous row
  (2026-07-14, commit `5f550b49`) and this run (`921f3102`) — a real,
  deterministic drop in HeapSyn's own tick count that has nothing to do with
  bytecode dispatch mode. This lines up with eu-rb5n (PR #1016, merged
  2026-07-15, before this branch's base commit), which closed the
  HOF-combinator demand-analysis fusion gap and is exactly described in
  `tests/tick_parity_test.rs`'s own comments as improving strictness for
  fold-shaped combinators. The ledger's "previous" row simply predates that
  unrelated fix, so `--check`'s delta is comparing across it, not across this
  PR's change. `--check`'s own summary line ("REGRESSION(S) found") is
  therefore a false positive with respect to this PR specifically —
  documented here rather than silently dismissed.

  **Follow-up (not fixed in this PR, flagged to the coordinator):**
  `hof_fold` now runs in ~65ms on both engines, well under the protocol's
  ~200ms floor for a valid canonical-suite entry (§3, §6) — it has fallen out
  of the suite's validity envelope because of the same fusion fix, unrelated
  to predecode. Needs a bead to either scale up the bench or retire it from
  the canonical eight.

  New ledger rows appended to `docs/superpowers/engine-ab/results.jsonl`
  (commit `921f3102`, `prelude_config: blob`, `runs: 5` each,
  `host: Darwin-25.5.0-arm64`, `rustc 1.97.0`).

Corrupted the decoded-instruction ordinal in `handle_op_predecoded`
(`src/eval/bytecode/machine.rs`, one-line: `closure.code() as usize` →
`closure.code() as usize + 1`) to prove `EU_PREDECODE=0` genuinely routes to
an independent code path rather than vacuously reusing the same one:

- With the fault injected, default dispatch (predecode on) failed with a
  genuine `bad index into environment` error on a trivial program.
- With the same fault injected, `EU_PREDECODE=0` still produced the correct
  result — the byte-dispatch path is untouched by the predecode-only bug.
- Restored the line, rebuilt, and reconfirmed both configs produce the
  correct result again.

This confirms the opt-out mechanism is real, not a no-op.

## Deviation from the #989 Phase 2 design

None. The design's Step A/B/C gates (flag-gated infra, BV2 fold-in,
eu-2sa6.5 fusion-parity precondition) all landed in earlier PRs (#1002 and
follow-ups); this PR is exactly Step D's "flip default, soak" half —
deletion (the second half of Step D) is deliberately deferred to a future
PR once the soak period elapses with no regressions.


gmorpheme:
**Wicket review — technical verification complete, all claims confirmed. NOT merging — escalating the flip decision itself to the owner.**

Reviewed in `/private/tmp/eu-wicket-1031` (head `e5f073a8` off master `4a42131b`), with `/private/tmp/eu-wicket-master-check` (clean `origin/master`) as a comparison worktree. This PR touches engine defaults, so the recorded-review rule applies; this comment is that review.

## Independently verified, all confirmed

**1. `EU_HEAPSYN=1` precedence over `EU_PREDECODE`** — verified two ways:
- Code: `driver/eval.rs:467` gates the entire bytecode engine behind `bytecode_enabled()` (`!heapsyn_enabled()`); when `EU_HEAPSYN=1`, the bytecode module (and `predecode_enabled()`) is never reached at all — this is architectural, not a runtime precedence check.
- Empirically, by fault injection: corrupted the decoded-instruction ordinal in `handle_op_predecoded` (`src/eval/bytecode/machine.rs:2308`, `closure.code() as usize` → `+ 1`, same one-line injection the PR describes). Default (predecode on) broke with `bad index 1 into environment`, matching the PR's claimed error class. `EU_PREDECODE=0` was unaffected. `EU_HEAPSYN=1` alone was unaffected. **`EU_HEAPSYN=1 EU_PREDECODE=1` together was still correct** — directly proving `EU_HEAPSYN` takes precedence even when both are explicitly set to "on," not just when `EU_PREDECODE` is unset. Restored, rebuilt, reconfirmed all four configs correct.

**2. Docs match behaviour**: `CLAUDE.md`'s env-var table and `docs/architecture.md`'s pre-decoded-IR section both correctly describe the new default, the `EU_PREDECODE=0` opt-out, and the `EU_HEAPSYN` no-op relationship — matches the verified code behaviour exactly.

**3. `test-byte-dispatch-baseline` CI job**: confirmed it runs (`./target/release/eu test --allow-io tests/harness` under `EU_PREDECODE=0`) and passed on this PR's CI (2m58s, "Byte-dispatch harness (soak baseline)"), structurally identical to and sitting alongside "HeapSyn harness (baseline)". Note: master has **no branch protection configured** (`gh api .../branches/master/protection` → 404 "Branch not protected") — "required checks" here is Wicket's process convention, not a GitHub-enforced gate, so this job is exactly as "required" as the heapsyn job already was: none of them are GitHub-enforced, all of them are checked by me before merge.

**4. Fault-injection reproduced independently** (see #1 above) — the opt-out is real, not vacuous.

**5. Design-conformance check against `docs/superpowers/specs/2026-07-13-predecoded-execution-ir-design.md` §6**: read Step D directly — "After a release... with the flag on by default and no regressions surfaced, delete the byte-path... do not delete the byte path at Step A." This PR is exactly the first half (flip + soak); deletion is correctly deferred. No deviation found.

**6. The two REGRESSED flags — reproduced the staleness argument exactly, with exact numbers**: built a clean `origin/master` worktree (no predecode-flip changes at all — proves `EU_PREDECODE` cannot be the cause), generated a genuinely blob-embedded binary (`cargo run -p xtask -- prelude-compile` — note `cargo xtask ...` as documented isn't runnable directly in a fresh worktree checkout, no alias is configured; `cargo run -p xtask --` is the working equivalent), and ran both flagged benches under `EU_HEAPSYN=1` with the exact target/heap-limit the `engine-ab` runner uses (`-t bench-hof-fold`/`-t bench-string-scale`, `--heap-limit-mib 12288`). Result: **`hof_fold` ticks = 2,185,410 and `string_scale` ticks = 46,042,949 — exact matches to the ledger's new row**, on a worktree containing zero predecode-flip code. This conclusively confirms the drop from the 2026-07-14 baseline (52,225,448 / ~76.8M) is entirely attributable to changes already on master (consistent with eu-rb5n/#1016) and has nothing to do with this PR. The `--check` REGRESSED flags are a stale-baseline false positive, exactly as claimed.

**7. Full suite, independently run in all three configs** (default/predecode-on, `EU_PREDECODE=0`, `EU_HEAPSYN=1`): 501/501 `harness_test.rs` each. `cargo fmt --all --check`: clean. `cargo clippy --all-targets -- -D warnings`: clean. Tick-parity tripwire (`source_prelude_tick_parity_tripwire`) passed in all three configs against a genuinely blob-embedded binary.

**8. CI**: all required checks pass, including the new byte-dispatch baseline job (release/prepare-release jobs correctly skipping, not a release PR).

## Why I'm not merging despite a clean technical bill of health

This isn't a defect finding — I found none. It's that **eu-vcr8's own bead text says "the FLIP DECISION goes to the owner with soak evidence"**, set when the phased plan was originally signed off (#989/#1002). That's an explicit, pre-existing delegation of this specific decision, not a default-caution call on my part. Combined with the coordinator's instruction to escalate rather than wave through if I'm uncertain about the measured-single evidence bar for a default flip, and the fact that the wall-clock confidence here is genuinely **measured-single, not measured-verified** (machine load 3.6–7.9 throughout, never under the protocol's `<1` target, per `docs/superpowers/engine-ab/PROTOCOL.md`) — the deterministic evidence (ticks, differential gate, GC verification, tripwires) is airtight and I'd merge on that alone, but the wall-clock performance claim underlying *why the default should flip at all* is not protocol-verified in this session.

Mitigating factors worth weighing (not my call to weigh, but for the record): the flip is a one-line env-var opt-out, trivially reversible; the 3-mode differential gate and byte-dispatch baseline CI job keep both paths gated through the soak; and the wall ratios in this measured-single run (0.786–0.965 across most classes) are consistent with, not divergent from, Phase 1's earlier measured-verified evidence (#1009: "7/8 wins, 3-17% typical").

**Recommendation**: technical review passes cleanly; I recommend the owner make the GO call on the default flip itself, per the bead's own process, informed by the verification above.

PR #1032, opened 2026-07-17
test(gc): fix three unit tests' assumptions broken by EU_GC_STRESS=1 (eu-swbz)

gmorpheme:
## Summary

Reproduced all three EU_GC_STRESS=1 unit-test failures (`eval::memory::collect::tests::test_no_evacuation_when_not_fragmented`, `eval::memory::collect::tests::test_evacuation_target_block_has_marked_lines_after_collection`, `eval::memory::heap::tests::test_collection_strategy_mark_in_place`) and root-caused each individually. **All three are category (a): a test assumption legitimately violated by the stress flag's own documented, intentional behaviour — not a genuine GC bug.** No GC production code changed; this is a test-side fix per the "prefer test-side unless mechanism-level evidence says GC-side" guidance.

**Revised after Wicket's first review** (see "Revision" section below) — the original test #2 fix passed review's own fault-injection but Wicket went further and found it was vacuous under stress against the *original production bug it exists to guard*. Fixed by decoupling the test's settling-phase collection from the ambient stress flag entirely, per Wicket's suggested direction. The corrected claim: test #2 now exercises the same real invariant in both configurations, not a stress-specific weakened one — see below.

### 1 & 3 — `test_no_evacuation_when_not_fragmented` / `test_collection_strategy_mark_in_place`

Both assert a densely-packed heap's `analyze_collection_strategy()` chooses `MarkInPlace` / returns an empty candidate list. `analyze_collection_strategy()`'s `gc_stress` branch (`src/eval/memory/heap.rs`) is explicitly documented: "GC stress mode: force `SelectiveEvacuation` on every collection so that evacuation pointer-update bugs surface on any platform, not just the aarch64 CI runner." Under the flag it skips fragmentation analysis entirely and unconditionally evacuates every unpinned block — the exact property these two tests assert doesn't hold. **Mechanism, not hand-wave**: this is the flag doing precisely what its own doc comment says it does.

Fix: added `Heap::gc_stress()` (a `#[cfg(test)]` accessor reading the heap's own cached flag, keeping the test in sync with what the heap under test actually observed at construction) and branched each assertion — under stress, assert the override's own documented behaviour (`SelectiveEvacuation`, non-empty candidates); otherwise assert the original density heuristic. Both tests exercise something real in **both** configurations rather than being skipped or weakened to vacuity under stress. **Unchanged by the revision** — Wicket reviewed these cleanly.

### 2 — `test_evacuation_target_block_has_marked_lines_after_collection`

**Original fix (reviewed and sent back — see "Revision" below for what actually shipped).**

Investigated carefully before concluding (a), since it's the test that literally exists to guard a previously-fixed evacuation-target-recycling bug (test119/monad_utility aarch64 crash) — a genuine (b) finding here would need to stop and report before any fix, per the checkpoint rule.

Root-caused with temporary diagnostic instrumentation (added, run under both configurations, reverted before committing):

- Unstressed: `rest_count = 1` before the test's own evacuation call → 1 pre-existing candidate → 2 new blocks appear in `unswept` afterward, **both** with 256 marked lines (512 small objects don't fit one fresh target block, so evacuation fills one and retires it to `filled_evacuation_blocks`, both are genuine targets). Test passes.
- Stressed: `rest_count = 2` before the test's own evacuation call (because the test's *first*, "mark-in-place settling" `collect()` call internally consults `analyze_collection_strategy()` too, so under stress it *also* force-evacuates, leaving 2 rest blocks where the non-stressed heuristic leaves 1). The test's own candidate list (`2..2+rest_count`, "every rest block" by design) then evacuates **both** pre-existing blocks simultaneously.

`finalise_evacuation()` only *adds* the new target block(s) to `rest`; it never removes the pre-existing candidate blocks that were already there. When a source block is fully drained, it legitimately has zero marked lines and rides into `unswept` alongside the genuine target — correct GC behaviour, not a bug. My original fix addressed this by capturing `rest_block_base_addresses()` immediately before the evacuation call and excluding those addresses from the "new block" set the assertion checks.

**What Wicket found wrong with that fix**: fault-injecting the *original production bug* (removing the `evacuation_target`/`filled_evacuation_blocks` search from `mark_region_in_block`, restoring the exact pre-fix test119/monad_utility recycling bug) with my address-exclusion fix in place, the test correctly failed unstressed but **passed silently under `EU_GC_STRESS=1`**, reproduced consistently across 5 runs. Mechanism: with the bug present, the settling-phase evacuation (forced by stress) *is itself* the recycling bug in miniature — its own target never gets marked, so it's indistinguishable from a legitimately-drained source by the time the test's explicit phase runs. The address-exclusion added to handle the *legitimate* drained-source case ended up hiding the *buggy* one too, since both present identically (0 marks, present in `rest` going in). My original claim "both tests now exercise something real in both configurations" was **not accurate for test #2** — under stress, in exactly the scenario it exists to catch, it was vacuous.

## Revision (Wicket's suggested direction, implemented)

Rather than making the address-exclusion logic smarter (which would still be trying to distinguish two states that present identically), decoupled the test's settling-phase collection from the ambient `EU_GC_STRESS` override entirely — the settling `collect()` call is test scaffolding (pack objects into `rest` before the test's own deliberate, explicitly-candidated evacuation begins), not the strategy heuristic under test, and has no reason to be subject to the stress override at all.

- Extracted the non-evacuating mark-and-sweep half of `collect()` into a shared `collect_mark_in_place()` helper (no behaviour change for production — `collect()` still calls it in exactly the same branch as before).
- Added a `#[cfg(test)]`-only `collect_mark_in_place_for_test()` entry point that calls it directly, **bypassing `analyze_collection_strategy()` (and so `gc_stress`) entirely**.
- Routed the test's settling call through this new entry point instead of `collect()`.
- **Removed the `pre_evacuation_rest` address-exclusion mechanism entirely** — with the settling phase now deterministic, `rest_count` is always 1 in both configurations, and the *original, pre-fix* assertion logic (no exclusion needed) is correct in both. The test body for the evacuation-and-assertion phase is now unchanged from before any of this PR's work — only the settling-phase call changed.

### Verification of the revision

Both required, both performed:

- **My own fault injection** (unchanged from before): reverted all three files to the pre-fix baseline (`a9add2b5`), confirmed the original three test failures reproduce exactly under `EU_GC_STRESS=1`, restored, reconfirmed green in both configs.
- **Wicket's fault injection** (the one that failed before): reintroduced the exact original production bug (temporarily removed the `evacuation_target`/`filled_evacuation_blocks` search from `mark_region_in_block`) with this revision's fix in place. The test now **fails correctly under both configurations** — ran 5 times under `EU_GC_STRESS=1` for consistency (matching Wicket's own verification rigor), failed every time with the expected panic message. Restored, reconfirmed green.

## CI reachability

None of the three existing `EU_GC_STRESS` CI jobs (`test-gc-verified`, `test-aarch64-sequential`, `asan`) run `cargo test --lib` — they all run `eu test tests/harness` via the compiled binary. So these GC-internal unit tests were **never gated by CI**, even before this fix — that's a real gap, not just this PR's excuse for not noticing sooner. Added a step to the main "Test Suite" job (`.github/workflows/build-rust.yaml`, all three OS matrix legs — ubuntu/macos/windows) running `EU_GC_STRESS=1 cargo test --lib` right after the existing `cargo test` step. Cheap: `cargo test --lib` alone runs in ~3s locally. **Unchanged by the revision** — Wicket reviewed this cleanly.

## Diagnostic signal (requested)

All three behave identically under `EU_HEAPSYN=1` and `EU_PREDECODE=0` combined with `EU_GC_STRESS=1` — verified, not just reasoned (ran `cargo test --lib` under all four combinations). Expected: these are pure `Heap`/`collect()`-level unit tests that construct and drive the heap directly, never invoking engine dispatch, so the execution-engine selector flags are inert here.

## Test status (post-revision)

- `cargo test`: green, unstressed and under `EU_GC_STRESS=1` (1308 lib + 500 harness + all other suites).
- `EU_GC_VERIFY=2 EU_GC_STRESS=1 EU_GC_POISON=1 cargo test --lib`: green.
- `EU_GC_VERIFY=2 EU_GC_STRESS=1 cargo test --test harness_test`: green (500/500).
- **Fault-injection (mine)**: reverted all three fix files to `a9add2b5`, reran `EU_GC_STRESS=1 cargo test --lib`, reproduced the original three failures exactly, restored, reconfirmed green in both modes.
- **Fault-injection (Wicket's, the one that previously failed)**: reintroduced the original production bug with the fix in place — test #2 now fails under both configurations, verified over 5 stressed runs. Restored, reconfirmed green.
- `cargo fmt --all`: clean.
- CI workflow YAML syntax validated (`python3 -c "import yaml; yaml.safe_load(...)"`).

## Notes for review

This touches `src/eval/memory/` (test code only, no production GC logic changed) — recorded-review rule applies.

### DIFF
--- a/.github/workflows/build-rust.yaml
+++ b/.github/workflows/build-rust.yaml
@@ -61,6 +61,16 @@ jobs:
       - uses: dtolnay/rust-toolchain@stable
       - name: Run tests
         run: cargo test
+      - name: Run lib unit tests under GC stress (eu-swbz)
+        # The other EU_GC_STRESS jobs in this workflow (test-gc-verified,
+        # test-aarch64-sequential, asan) all run `eu test tests/harness` —
+        # none of them exercise `cargo test --lib`, so GC-internal unit tests
+        # (eval::memory::{collect,heap}::tests::*) were never gated under this
+        # flag. `cargo test --lib` alone is cheap (~3s locally), so this is a
+        # small additional step rather than a new job.
+        run: cargo test --lib
+        env:
+          EU_GC_STRESS: "1"
       - name: Test doc examples
         if: runner.os == 'Linux'
         run: |
--- a/src/eval/memory/collect.rs
+++ b/src/eval/memory/collect.rs
@@ -394,6 +394,25 @@ pub fn collect(roots: &mut dyn GcScannable, heap: &mut Heap, clock: &mut Clock, 
         return collect_with_evacuation(roots, heap, clock, &candidates, dump_heap);
     }
 
+    collect_mark_in_place(roots, heap, clock, dump_heap);
+}
+
+/// The non-evacuating mark-and-sweep half of [`collect`]: mark reachable
+/// objects, then defer sweep to allocation time.
+///
+/// Factored out so [`collect`] (which first consults
+/// `analyze_collection_strategy` — including its `EU_GC_STRESS` override,
+/// see that method's doc comment) and the `#[cfg(test)]`-only
+/// `collect_mark_in_place_for_test` (which deliberately bypasses strategy
+/// analysis, so test scaffolding that needs a deterministic non-evacuating
+/// pass isn't subject to the ambient stress flag — eu-swbz) share one
+/// implementation rather than drifting apart.
+fn collect_mark_in_place(
+    roots: &mut dyn GcScannable,
+    heap: &mut Heap,
+    clock: &mut Clock,
+    dump_heap: bool,
+) {
     if dump_heap {
         eprintln!("GC!");
     }
@@ -448,6 +467,36 @@ pub fn collect(roots: &mut dyn GcScannable, heap: &mut Heap, clock: &mut Clock, 
 
     // After collection, flip mark state ready for next collection
     heap_view.heap.flip_mark_state();
+}
+
+/// Test-only entry point for a deterministic, non-evacuating mark-and-sweep
+/// pass that never consults `analyze_collection_strategy` (and so is never
+/// subject to `EU_GC_STRESS`'s force-evacuate-everything override).
+///
+/// Some tests use a "settling" collection purely as scaffolding — to pack
+/// live objects into `rest` blocks before the test's *own* deliberate,
+/// explicitly-candidated evacuation begins — and are not testing the
+/// strategy heuristic itself (that's what
+/// `heap::tests::test_collection_strategy_*` and
+/// `collect::tests::test_no_evacuation_when_not_fragmented` are for). Routing
+/// that scaffolding through plain `collect()` made such a test's setup
+/// non-deterministic under `EU_GC_STRESS=1` (eu-swbz,
+/// `test_evacuation_target_block_has_marked_lines_after_collection`): the
+/// settling pass would itself force-evacuate, changing how many pre-existing
+/// `rest` blocks the test's own explicit evacuation then had to distinguish
+/// from genuine targets — including, in the presence of the exact
+/// evacuation-target-recycling bug this test exists to catch, corrupting the
+/// test's own setup before its assertion ever ran (silently passing instead
+/// of failing). Use this whenever a test needs "pack objects into `rest`,
+/// deterministically, regardless of the ambient stress flag" rather than
+/// "collect with whatever strategy is currently in effect".
+#[cfg(test)]
+pub fn collect_mark_in_place_for_test(
+    roots: &mut dyn GcScannable,
+    heap: &mut Heap,
+    clock: &mut Clock,
+) {
+    collect_mark_in_place(roots, heap, clock, false)
 }
 
 /// Perform a collecting GC cycle with selective evacuation of fragmented blocks.
@@ -1444,32 +1493,28 @@ pub mod tests {
     ///
     /// `finalise_evacuation()` only *adds* target block(s) to `rest` (see its
     /// doc comment) — it never removes the pre-existing candidate blocks that
-    /// were already in `rest` going into the evacuation pass. When more than
-    /// one pre-existing `rest` block is evacuated *from* simultaneously
-    /// (which happens whenever this test's own candidate list —
-    /// `2..2+rest_count`, intentionally "every rest block" — covers more than
-    /// one block), a fully-drained source block legitimately ends up with
-    /// zero marked lines and rides along into `unswept` in the very same
-    /// `defer_sweep()` call as the genuine target(s). That is correct GC
-    /// behaviour, not a bug: nothing should mark a block whose entire live
-    /// content was copied elsewhere. So "new in `unswept`" alone is not
-    /// sufficient to identify "is an evacuation target" — the test also
-    /// excludes any address that was already present in `rest` immediately
-    /// before the evacuation call (eu-swbz).
-    ///
-    /// Under `EU_GC_STRESS=1` this distinction is exercised on every run: the
-    /// "mark-in-place settling" `collect()` two lines below internally calls
-    /// `analyze_collection_strategy()` (see `collect`'s own body), which
-    /// under stress unconditionally evacuates every unpinned block instead of
-    /// only fragmented ones — so the settling pass itself typically leaves
-    /// *multiple* rest blocks where the non-stressed heuristic leaves at most
-    /// one, and this test's own (always-evacuate-every-rest-block) candidate
-    /// list then evacuates multiple blocks in the same pass every time.
-    /// Confirmed empirically (not just reasoned): under `EU_GC_STRESS=1` the
-    /// block this test previously flagged as "zero marked lines" is exactly
-    /// one of the addresses returned by `rest_block_base_addresses()`
-    /// *before* `collect_with_evacuation` runs — i.e. a source block, not a
-    /// target.
+    /// were already in `rest` going into the evacuation pass. If the
+    /// "settling" collection below ever left more than one pre-existing
+    /// `rest` block for this test's own (always-evacuate-every-rest-block)
+    /// candidate list to evacuate simultaneously, a fully-drained *source*
+    /// block would legitimately end up with zero marked lines too (nothing
+    /// should mark a block whose entire live content was copied elsewhere)
+    /// and ride into `unswept` alongside the genuine target(s) — making "new
+    /// in `unswept`" alone insufficient to identify a target, and (eu-swbz,
+    /// confirmed via Wicket's fault injection during review) *indistinguishable
+    /// from the exact bug this test exists to catch*: with the bug present,
+    /// the settling pass's own evacuation target is itself silently
+    /// unmarked, so it looks exactly like a legitimately-drained source by
+    /// the time this test's explicit phase runs — an address-based exclusion
+    /// added to handle the legitimate case ends up hiding the buggy one too,
+    /// letting the bug pass silently under `EU_GC_STRESS=1` instead of
+    /// failing. Rather than trying to make that distinction more precise,
+    /// the settling collection below is deliberately routed through
+    /// `collect_mark_in_place_for_test` — which never evacuates, regardless
+    /// of the ambient stress flag — so this test's own setup is always
+    /// exactly one pre-existing `rest` block in both configurations, and the
+    /// original single-candidate assertion logic is correct without any
+    /// address-exclusion mechanism at all.
     ///
     /// ## Pass/fail criterion
     ///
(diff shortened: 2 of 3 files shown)


gmorpheme:
**Wicket review — send back. Found a genuine non-vacuity gap in test #2 under EU_GC_STRESS=1 via fault injection.**

Reviewed in `/private/tmp/eu-wicket-1032` (head `a79cde35` off master `a9add2b5`). GC-adjacent test logic, recorded-review rule applies.

## What I verified and confirms the PR's claims

- **Diff scope**: exactly the claimed 3 files (`.github/workflows/build-rust.yaml`, `src/eval/memory/collect.rs`, `src/eval/memory/heap.rs`), 130 insertions / 15 deletions. The only non-test addition is `Heap::gc_stress()`, and it's `#[cfg(test)]`-gated — genuinely no production logic changed.
- **Fault injection reproduced exactly as claimed**: reverted both test files to pre-PR state (`git checkout a9add2b5 -- src/eval/memory/collect.rs src/eval/memory/heap.rs`), rebuilt, ran the three named tests under `EU_GC_STRESS=1` — all three FAILED with the exact panic messages/locations the PR quotes (including the `test_evacuation_target_block_has_marked_lines_after_collection` panic's own reference to "test119 / monad_utility crash on aarch64"). Unstressed, all three passed (confirming the assumption violation is genuinely stress-specific). Restored, rebuilt, reconfirmed all three green in both configs.
- **CI wiring**: the new step is a plain `run: cargo test --lib` with `env: EU_GC_STRESS: "1"` — no `continue-on-error`, will fail the job like any other step. Confirmed `EU_GC_STRESS=1 cargo test --lib` is genuinely absent from the other three stress jobs (`test-gc-verified`, `test-aarch64-sequential`, `asan`) — they all run `eu test tests/harness` via the compiled binary, never `cargo test --lib`. The reachability-gap claim is accurate.
- `cargo fmt --all --check`: clean. Full `cargo test --lib`: 1308/1308 both unstressed and under `EU_GC_STRESS=1`.

## The finding: test #2's loosened assertion does not reliably catch the bug it exists to guard, under EU_GC_STRESS=1

Per your instruction, I didn't stop at reproducing the test-fix fault injection — I went further and fault-injected the **original production bug** `test_evacuation_target_block_has_marked_lines_after_collection` guards against (removed the `evacuation_target`/`filled_evacuation_blocks` search from `mark_region_in_block` in `heap.rs:797-811`, restoring the exact pre-fix `test119`/`monad_utility` recycling bug), leaving the PR's **test-side fix in place**, and ran the test in both configs:

- **Unstressed**: FAILS correctly — `evacuation target block 0x... has 0 marked lines after collection: ... (test119 / monad_utility crash on aarch64)`.
- **`EU_GC_STRESS=1`**: **passes silently** — no failure, no panic, nothing. Reproduced consistently across 5 separate runs (varying heap addresses each time, same outcome every time).

Root cause, confirmed with temporary instrumentation (`eprintln!` on `pre_evacuation_rest`/`unswept_before`/`unswept_after`/`new_blocks`, added and reverted): under stress with the bug present, `pre_evacuation_rest` and `unswept_after` end up as **the same single address**, so `new_blocks` (after the fix's own exclusion filter) is empty, and the test hits its own `if new_blocks.is_empty() { return; }` early-skip — the same "nothing to check" escape hatch the PR's fix relies on for the *legitimate* drained-source case.

The mechanism: under stress, the test's own "settling" `collect()` call (the one that's supposed to just pack objects into `rest` before the test's *own* explicit evacuation phase begins) *also* force-evacuates, per stress mode's documented override — this is the same interaction the PR's own analysis for test #2 already describes (`rest_count` 1→2). But with the original bug present, that settling-phase evacuation *is itself* the recycling bug in miniature: the settling pass's own evacuation target never gets marked, so it looks identical to a legitimately-fully-drained block by the time the test's explicit phase runs. The fix's `pre_evacuation_rest` exclusion — designed to distinguish "legitimately drained source" from "genuine target" — cannot distinguish "legitimately drained source" from "buggy, silently-corrupted-by-the-very-bug-under-test source," because both present identically (0 marks, present in `rest` going into the explicit phase). The bug corrupts the test's own setup before the assertion under test ever runs, and the exclusion logic added to handle a different (legitimate) case ends up hiding it.

**Practical impact**: the underlying regression (`test119`/`monad_utility`) is not left completely unguarded — it's still caught (a) by this same unit test unstressed (verified above, this path is unaffected), and (b) by the actual `tests/harness/119_monad_utility.eu` harness fixture running under `EU_GC_STRESS=1` in the existing `test-gc-verified` and `test-aarch64-sequential` CI jobs (confirmed this fixture exists and those jobs already run the full harness under stress). So this isn't "the bug can ship silently" — but it does mean:
1. The PR's claim "Both tests now exercise something real in both configurations rather than being skipped or weakened to vacuity under stress" is **not accurate for test #2** — under stress, in exactly the scenario it exists to catch, it's vacuous.
2. The new `EU_GC_STRESS=1 cargo test --lib` CI step adds no incremental protection for this specific regression class beyond what the harness-level stress jobs already provide — its main value is for the other two tests (which do meaningfully assert the override's own behaviour under stress).

## Suggested direction (not prescribing the exact fix)

The cleanest fix is probably to decouple the test's own setup from the ambient `EU_GC_STRESS` env var rather than trying to make the address-exclusion logic smarter: the "settling" `collect()` call is test scaffolding (pack objects into `rest` before the test's *own* deliberate evacuation), not the thing under test, so it arguably shouldn't be subject to the stress override at all — e.g. drive it through a stress-independent path, or construct a second `Heap` for that phase with stress forced off, so `rest_count` is deterministically 1 regardless of the ambient flag, and the existing (already-correct, pre-PR) assertion logic works unmodified in both configs. That would also remove the need for the `pre_evacuation_rest` exclusion mechanism entirely, since it's specifically compensating for an interaction the settling-phase decoupling would eliminate at the source — worth considering per the "flag domain-specific mechanisms where a general fix exists" review lens.

## Verdict

Sending back. Please address the test #2 non-vacuity gap (or explicitly document the residual reliance on the harness-level jobs for that bug class, if the mechanism truly can't be made stress-safe cheaply, and adjust the PR's own claims accordingly) before I re-review. Everything else in the PR — the other two tests' fixes, the CI wiring, the diagnostic-signal claims — checks out cleanly.

gmorpheme:
**Wicket re-review — approved and merging.** The revision correctly fixes the vacuity gap from my first review. GC-surface change (production code structure now touched), recorded-review rule applies — this comment is that review.

Re-reviewed in `/private/tmp/eu-wicket-1032` (head `0e84c077` off master `a9add2b5`).

## Verified the extraction is genuinely behaviour-neutral

Diffed `collect()`'s pre-PR body (from `a9add2b5`) against the new `collect_mark_in_place()` function body character-for-character (not just eyeballing it): identical except for the function-signature wrapping — the actual logic (mark phase, verification hooks, sweep deferral, collection-count/peak-block bookkeeping, mark-state flip) is byte-for-byte unchanged. `collect()`'s own remaining body (`analyze_collection_strategy` → branch → `collect_with_evacuation` or `collect_mark_in_place`) is also unchanged control flow, just calling the extracted function instead of inlining it. This is a genuinely mechanical, behaviour-neutral extraction.

## Confirmed `collect_mark_in_place_for_test` is test-only and unreachable in production

`grep -rn collect_mark_in_place` across `src/` and `tests/` shows exactly three call sites: `collect()`'s own call to the (private, non-`pub`) `collect_mark_in_place`, the `#[cfg(test)]`-gated `collect_mark_in_place_for_test`'s call to the same, and the one test (`test_evacuation_target_block_has_marked_lines_after_collection`) that calls the test-only wrapper. No other production code path can reach it.

## Re-ran my own fault injection myself (didn't take the 5/5 report on trust)

Reintroduced the exact original production bug (removed the `evacuation_target`/`filled_evacuation_blocks` search from `mark_region_in_block` in `heap.rs`) with the revised fix in place:
- Unstressed: **FAILS correctly** with the expected panic message.
- `EU_GC_STRESS=1`, run **6 times** (one more than the reported 5, for good measure): **FAILS correctly every time** — no vacuous pass, unlike the original fix.

Restored, rebuilt, reconfirmed the test passes cleanly in both configs.

## Re-ran the original three-failure reproduction

Reverted both `collect.rs` and `heap.rs` to the `a9add2b5` pre-PR baseline, ran all three named tests under `EU_GC_STRESS=1`: all three FAILED with the exact original panic messages/locations. Restored, rebuilt, reconfirmed green.

## Full suite, independently run

- `cargo test --release --lib`: 1308/1308, unstressed and `EU_GC_STRESS=1`.
- `EU_GC_VERIFY=2 EU_GC_STRESS=1 EU_GC_POISON=1 cargo test --release --lib`: 1308/1308.
- `cargo test --release --test harness_test`: 500/500, unstressed and `EU_GC_STRESS=1`.
- `cargo fmt --all --check`: clean. `cargo clippy --all-targets -- -D warnings`: clean.
- CI: fully green including the new `EU_GC_STRESS=1 cargo test --lib` step on all three OS legs.

## PR body corrections

Confirmed accurate: the body correctly documents the revision history (original vacuity finding, the fix direction taken, what changed), and the "Unchanged by the revision" callouts for tests #1/#3 and the CI-reachability fix match what I verified in the original review — no drift there.

No architectural concerns. The `collect_mark_in_place`/`collect_mark_in_place_for_test` split is exactly the general fix I'd hoped for: it removes the address-exclusion special-case entirely rather than making it smarter, and gives test scaffolding a principled way to opt out of the stress override without weakening what's actually being asserted.

Merging.

PR #1033, opened 2026-07-17
fix(tree-sitter): resolve operator-prefix-collision declaration-boundary bug (eu-2sa6.17 item 3)

gmorpheme:
## Summary

eu-2sa6.17 item 3: root-causes and fixes the parser-equivalence audit's
"operator-prefix-collision" bug (`(l < r): l` then `(l <= r): l` in the
same scope corrupting the parse, blocking `lib/prelude.eu` from parsing in
tree-sitter-based tooling). Do NOT merge — awaiting gatekeeper review.

Turns out the "prefix collision" framing was a red herring — it's actually
**two unrelated bugs** in the external scanner's declaration-boundary
lookahead (`editors/tree-sitter-eucalypt/src/scanner.c`'s
`at_declaration_start()`):

1. **Missing immediate-adjacency check.** The identifier/quoted-identifier
   branches allowed whitespace/newlines between an identifier and a
   following `(` when checking for the `identifier(...):` function-decl
   pattern. This let a bare identifier that is the *value* of one
   declaration be misparsed as the *head* of the next declaration when
   that next declaration happens to start with `(` on its own line.
   Mirrors `grammar.js`'s `parameter_list`, which had the exact same gap
   (also fixed here, with `token.immediate('(')`) — `f (x): x` (space
   before parens) is now correctly rejected, matching the Rowan parser
   (verified against `target/release/eu dump ast`).
2. **`is_operator_char()` in `scanner.c` was a *third* independently
   drifted copy** of the operator-character set, alongside `grammar.js`'s
   `OPER_CHARS` (fixed in PR #996) and the Rust lexer's dynamic
   Unicode-category detection (see eu-fbyk). It was missing 14 characters
   (`¬ ¡ £ € ± × ÷ ✓ ▶ △ ▽ ⨈ ? $`), so the scanner failed to recognise
   `(op ...):` declaration heads using them — silently corrupting the
   *preceding* declaration's value instead of raising a visible error.

Both bugs independently corrupt any declaration sequence where an earlier
declaration's value abuts a `(`-headed declaration using an affected
character. This — not "prefix collision" — is what actually broke
`lib/prelude.eu` and `tests/harness/020_op_precedence.eu` (proven by
regression cases using non-prefix-related operator pairs, non-operator
declarations, and non-identifier-ending soup values — see the new corpus
tests).

Also removed the now-unnecessary `[$.declaration_head, $.block_param]`
conflict declaration — `tree-sitter generate` started flagging it as
redundant once `parameter_list` requires immediate adjacency.

## Root cause detail

Traced via `tree-sitter parse -d` (external-scanner debug trace): for the
one case that *did* already work before this PR (soup = a single bare
trailing identifier), `_declaration_end` was queried and fired correctly.
For every other soup shape (operator+literal, application, etc.), the
external token was queried too — but the scanner itself returned false
because its own operator/identifier detection didn't recognise the
upcoming `(op ...):` head. Once scanner.c's character tables were brought
in line with `grammar.js`'s, declaration-boundary detection turned out to
already be sound across all soup shapes I could construct — this was not
the deeper "DECLARATION_END never queried for non-identifier-ending soup"
architectural gap I initially suspected (see commit message / bd comment
for the false trail and how it was ruled out).

## Bonus finding (filed separately, not fixed here)

While isolating test cases, found that postfix operator declarations using
`!`, `*`, or `-` (e.g. `(x!): x`) fail to parse **even standalone**,
because those characters are simultaneously valid identifier-continuation
characters — maximal-munch tokenisation swallows `x!` into a single
identifier token rather than `x` + operator `!`. This is a real,
pre-existing, unrelated bug, but grep confirms it's not used anywhere in
`lib/` or `tests/harness/`, so it's dormant. Filed as eu-ctoj (P4) rather
than fixed here — needs its own grammar-internals investigation.

## eu-fbyk status

**Still open, not superseded.** eu-fbyk (proper Unicode-category-based
external scanner, mirroring the Rust lexer) remains the correct long-term
fix for all three drifted character-set copies (`grammar.js`'s
`OPER_CHARS`, this PR's `scanner.c` list, and the identifier
Letterlike-Symbols gap from audit §2.9, e.g. `ℕ`). The two bugs fixed here
were both narrow, well-isolated, low-risk issues addressable without that
larger rewrite — I judged patching them now was better than leaving a
known, diagnosed, `lib/prelude.eu`-blocking bug unfixed while eu-fbyk (P4,
unscheduled) waits.

## Divergence count

- Before: 30/353 corpus files failing to parse (visible ERROR node).
- After: 29/353 (one fewer: `tests/harness/020_op_precedence.eu`, verified
  by full structural inspection — all 7 inner declarations of its
  `unary-ops` block now split correctly, not just "no ERROR node").
- `lib/prelude.eu` **remains** correctly listed as failing — it hits
  separate, out-of-scope issues (bracket-expr declaration heads §2.3,
  nested-brace `s"..."` strings §2.7). Comment in
  `corpus-expected-failures.txt` updated to reflect precisely which bugs
  are fixed vs remaining.
- `lib/markup.eu` and `020_op_precedence.eu` pruned from the expected-
  failures list (`lib/markup.eu` was resolved separately, by a source-side
  `=`→`:` fix already landed on master, unrelated to this PR).

## Test plan

- [x] `tree-sitter test` (grammar's own corpus, now 46 cases including 6
      new regressions) — 46/46 pass
- [x] `python3 scripts/check-tree-sitter-corpus.py` — green, 29/29
- [x] **Fault-injection verified**, both fixes, individually: reverted the
      immediate-adjacency fix alone → 5 of the 6 new corpus tests fail as
      expected (the 6th, unrelated to that fix, still passes); restored,
      confirmed green. Reverted the operator-char widening alone → the one
      new test exercising a previously-missing character (`▶`) fails as
      expected, others unaffected; restored, confirmed green.
- [x] `lib/prelude.eu` / `tests/harness/020_op_precedence.eu` spot-checked
      with full tree dumps (not just exit code), confirming genuinely
      correct declaration splitting where claimed
- [x] Regenerated `parser.c` with the `tree-sitter-cli` version pinned in
      `package.json` (`^0.22.0`, via `npx`) to keep the diff minimal;
      confirmed a clean (no-op) regeneration from unmodified `grammar.js`
      reproduces the committed `parser.c` byte-for-byte first, so the
      diff here is attributable to this PR's changes
- [x] Cross-checked with the Homebrew-installed `tree-sitter-cli` 0.26.11
      too (both agree)


gmorpheme:
**Wicket review — approved and merging.** Tree-sitter grammar files, no Rust engine surface, but per the review charter the CI corpus check's integrity is the core question here — see the shared findings note on #1034 for the deep-dive on that.

Reviewed in `/private/tmp/eu-wicket-1033` (branch `fix/tree-sitter-remainder`, head `0a6c9657` off master `6dc63d21`), using both the `package.json`-pinned CLI approach and my locally-installed `tree-sitter 0.26.11` (matching the PR's own cross-check).

**Corpus divergence count**: independently reproduced exactly — 29/353 currently failing, 29 expected, 0 new divergences, matching "30 → 29 (one fewer: `020_op_precedence.eu`)" precisely.

**`tree-sitter test`**: 46/46, matching the claim.

**Both fixes independently fault-injected** (not just reproducing the PR's own injection, ran it myself from scratch):
1. Reverted `is_operator_char()`'s 14 added characters back to the pre-PR set → 45/46 (exactly 1 failure, the `▶` case) — matches "the one new test exercising a previously-missing character... fails as expected, others unaffected."
2. Reverted the immediate-adjacency check (`had_space` guard in `scanner.c`, plus `token.immediate('(')` → `'('` in `grammar.js`, restoring the `[$.declaration_head, $.block_param]` conflict which — confirmed — `tree-sitter generate` immediately re-flags as required the instant adjacency is removed, exact converse of the PR's "generator started flagging it as redundant" claim) → 41/46, 5 failures — matches "5 of the 6 new corpus tests fail as expected" precisely.
3. Directly verified the `f (x): x` (space before parens) claim: produces an `ERROR` node with the fix in place, confirming the Rowan-matching rejection.

Restored both, reconfirmed 46/46 and 29/353 clean.

No production Rust code touched (confirmed via `gh pr diff --name-only`), so `cargo fmt`/`clippy`/`test` are genuinely not applicable as claimed.

The "operator-prefix-collision was a red herring" root-cause story holds up — both bugs are demonstrably orthogonal to prefix-sharing (verified via the fault injections above, which show each bug firing independently regardless of operator prefix relationships).

Filed-not-fixed items (eu-ctoj postfix operator/identifier overlap, eu-fbyk not superseded) are appropriately out of scope for this PR — see my comment on #1034 for one follow-up on eu-fbyk's cross-referencing.

Merging.

PR #1034, opened 2026-07-17
fix(tree-sitter): gradual-typing binding-scope keyword coverage (eu-2sa6.17 item 4)

gmorpheme:
## Summary

eu-2sa6.17 item 4: gradual-typing syntax coverage in the tree-sitter grammar.
**Stacked on #1033 (item 3)** — targets that branch, not `master`, since it
builds on the scanner/grammar fixes there. Do NOT merge — awaiting
gatekeeper review, and should land after #1033.

Resolves 9 of the audit's 12 "gradual-typing syntax" divergences, from a
different angle than the audit expected: `:for`/`:let`/`:io` binding-scope
keywords need **no new grammar rule at all**. `{ :for x: 42 }` already
parses correctly via the *existing* generic `block_metadata` mechanism —
`:for` is just an ordinary symbol literal used as block-level metadata,
exactly like any other `` `:tag `` annotation (confirmed with
`target/release/eu dump ast`). The actual gap: `unit_metadata` only
accepted a *bare* block/string at the top level, with nothing allowed to
follow it — so `{ :for x: 42 }.(x)` (block-dot access, present in every
one of these test files) failed once the parser hit the closing `}`.

## What changed

1. **`unit_metadata`'s block alternative can now be followed by more soup
   content**, gated by `_declaration_end` (mirroring how `declaration`
   itself terminates its own soup), so `{ :for x: 42 }.(x)` parses as
   block + `.` operator + `(x)` paren_expr — while a leading metadata
   block still correctly stops before a **separate** following
   declaration (regression-guarded).
2. **`grammar.js`'s `OPER_CHARS` was missing a bare `\`** — fixes
   `034_literal_string_synthesis.eu` / `035_literal_string_dsl.eu` /
   `037_union_absorbs_literal_string.eu`, which the audit miscategorised
   under "gradual typing" but are actually unrelated: a literal-string
   type-DSL idiom that catenates strings via `\` used *as an operator*
   (`"\"read\" | \"write\" | \"rw\""` tokenises, in both Rowan and now
   tree-sitter, as alternating STRING/NAME/OPERATOR(`\`) soup elements —
   plain `"..."` strings have no backslash escape processing in this
   language at all, so this isn't really "escaping", it's catenation).

## A more general fix was tried and reverted

Making `unit_metadata` accept an arbitrary `$.soup` (to also cover bare
non-block top-level expressions, e.g. a file that's just
`for.map(negate, [1,2,3])` with no declarations) created a genuine,
generator-undetected ambiguity with `repeat($.declaration)`: `f(x): x`
got silently misparsed as a 2-element soup (application `f(x)` then a
colon-prefixed symbol literal `: x`) instead of a declaration — with
**zero error nodes**, so `tree-sitter parse -q`'s exit code didn't even
catch it. This broke **14** previously-passing corpus files. Caught by
running the full corpus check before committing; reverted in favour of
the narrower, block-gated fix, which introduces no such ambiguity
(nothing else in the grammar starts a soup element with `{`-then-`:for`-
shaped content the same way a bare identifier does).

## Two files remain, deliberately unfixed

Documented with full reasoning in `corpus-expected-failures.txt`:

- **`023_for_map_no_warning.eu` / `024_for_bind_no_warning.eu`**:
  miscategorised by the audit — no gradual-typing syntax involved at all,
  just a bare top-level expression with zero declarations. This is
  exactly the general "`unit_metadata` = any soup" gap reverted above for
  causing regressions; a safe fix needs real GLR-conflict + dynamic-
  precedence work I judged too risky to attempt in this pass.
- **`038_literal_string_annotation_widens.eu`**: **not a tree-sitter gap
  at all** — the Rowan parser rejects this file's `expr : "type"` inline-
  ascription syntax too (`eu dump ast`: `error: malformed declaration
  head` at the `x: "hello" : "string"` line; confirmed this exact inline
  syntax is used nowhere else in `lib/` or `tests/harness/`). The harness
  test for it currently passes only because `eu check --strict` exits 0
  despite printing that parse error to stderr — filed as **eu-3621 (P2)**,
  a real `eu check` correctness bug, independent of tree-sitter and out of
  scope here.

## Divergence count

Cumulative with item 3 (#1033): **30 → 20** of 353 corpus files failing to
parse.

## eu-fbyk status

Still open, not superseded (see #1033 for the fuller writeup). This PR
finds and fixes a companion drifted-character-set gap in `grammar.js`
itself (missing `\` in `OPER_CHARS`), on top of the two found in
`scanner.c` there — reinforcing that eu-fbyk's proper Unicode-category-
based rewrite is the right long-term fix for this whole family of bugs,
even though the individual gaps found so far have been small enough to
patch directly.

## Test plan

- [x] `tree-sitter test` — 49/49 pass (3 new regression cases added here,
      on top of item 3's 46)
- [x] `python3 scripts/check-tree-sitter-corpus.py` — green, 20/20
- [x] **Fault-injection verified**, both fixes, individually: reverted the
      `unit_metadata` block+trailing-soup change alone → the relevant new
      test fails as expected, the other two unaffected; restored,
      confirmed green. Reverted the `\` addition to `OPER_CHARS` alone →
      the relevant new test fails as expected; restored, confirmed green.
- [x] Cross-checked with both the `package.json`-pinned `tree-sitter-cli`
      (`^0.22.0`, via `npx`, for a minimal generated-file diff) and the
      Homebrew-installed 0.26.11 — both agree, 49/49


gmorpheme:
**Wicket review — approved and merging (after #1033 lands and this is retargeted to master).** This is the load-bearing review of the two — the narrow-vs-general judgement here, and the integrity of the corpus check that caught the reverted general fix, was the review's core question.

Reviewed in `/private/tmp/eu-wicket-1033` (branch `fix/tree-sitter-gradual-typing-coverage`, head `d41ffff4`, stacked on `fix/tree-sitter-remainder`/#1033 — confirmed via `git log pr1034 --not pr1033`, exactly one commit ahead as expected).

**Corpus divergence count**: independently reproduced exactly — 20/353 currently failing, 20 expected, 0 new divergences, matching the claimed cumulative "30 → 20". `tree-sitter test`: 49/49.

**Both fixes independently fault-injected**:
1. Reverted `OPER_CHARS`'s added `\` → 48/49, exactly 1 failure — matches "the relevant new test fails as expected, the other two unaffected."
2. Reverted `unit_metadata`'s trailing-soup addition back to bare-`$.block`-only → 48/49, exactly 1 failure — matches the claim precisely.

**The narrow-vs-general judgement — verified by reproducing the exact regression from scratch, not just trusting the PR's account.** I rebuilt the "more general fix" the PR describes trying (`unit_metadata: $ => $.soup` instead of the block-gated version) and ran both verification paths:

- `tree-sitter generate` succeeded with **zero conflict warnings** — confirming this is a genuinely generator-undetected ambiguity, not something `tree-sitter generate` itself would have caught.
- `tree-sitter test` (the hand-written 49-case corpus): **28/49 failed**, including exactly the `f(x): x`-shaped misparse the PR describes (application swallowed into a 2-element soup instead of a declaration).
- **`python3 scripts/check-tree-sitter-corpus.py` (the CI-gating script) — reported "Currently failing to parse: 16" (DOWN from 20) and exited 0, "OK: no new tree-sitter grammar divergences."** It even flagged 4 files as "previously-expected failures now parse successfully," including `023_for_map_no_warning.eu` and `024_for_bind_no_warning.eu` — files this very PR documents as still-broken-and-deliberately-unfixed — as apparent *improvements*.

This confirms, empirically rather than by inference: **`check-tree-sitter-corpus.py` in isolation cannot detect a silent, zero-error-node misparse — it only counts `tree-sitter parse`'s exit code (ERROR-node presence), with no comparison against an expected/correct tree.** It is blind to exactly the dangerous bug class this PR's own regression exemplifies, and can actively report a regression as an improvement.

**However, the CI job as a whole is not vacuous for this specific regression.** `.github/workflows/tree-sitter.yaml`'s single `corpus-parse` job runs `npx tree-sitter test` *before* `check-tree-sitter-corpus.py`, as sequential steps with no `continue-on-error` — a `tree-sitter test` failure aborts the job before the blind corpus-check step ever gets a chance to report a misleading pass. Since this specific regression happens to also break several of the 49 hand-written corpus cases, CI would have correctly failed the PR had this version been pushed.

**The residual gap worth flagging (not a blocker for this PR, but a real finding for eu-wrf7/eu-2sa6.17's own goal of a CI corpus-parse gate)**: `tree-sitter test`'s protection only extends to the ~49 hand-written cases in `test/corpus/*.txt`. For the 353 *real* corpus files (`lib/`, `tests/harness/`, `examples/aoc25/`) — the actual production-relevant surface this check exists to protect — the only mechanism is `check-tree-sitter-corpus.py`, which I've now shown cannot detect a silent misparse of any of those files unless it happens to coincide with (or be represented by) one of the hand-written cases. A future grammar change causing a similar silent, zero-error misparse of a real corpus file not mirrored in `test/corpus/*.txt` would pass CI undetected. Recommend a follow-up bead: strengthen `check-tree-sitter-corpus.py` to do real structural comparison (e.g. against a frozen S-expression snapshot per corpus file, or cross-checked against Rowan's own AST shape) rather than exit-code-only. This is a pre-existing gap in the script (present since PR #996), not something either #1033 or #1034 introduces or worsens.

**`corpus-expected-failures.txt` entries verified**:
- `023_for_map_no_warning.eu` / `024_for_bind_no_warning.eu`: confirmed these are bare top-level expressions with zero declarations, not gradual-typing syntax — the audit miscategorised them, and the file's own reasoning (same ambiguity class as the reverted general fix) is accurate.
- `038_literal_string_annotation_widens.eu`: confirmed via `eu-3621` — a real, well-described, correctly-scoped bug (`eu check --strict` not propagating a genuine Rowan parse-error exit code), independent of tree-sitter, filed as P2 with an accurate description matching what's in the corpus-expected-failures comment.

**eu-fbyk status**: confirmed still `OPEN` (P4). One minor process note: neither this PR nor #1033 actually left a `bd comment` on eu-fbyk itself — both PR bodies discuss "eu-fbyk: still open, not superseded" in prose, but the bead has no comments since 2026-06-12 and its description doesn't yet mention the two new drifted-copy findings (scanner.c's 14-char gap, grammar.js's missing `\`). Worth a quick follow-up comment on eu-fbyk itself so its own history captures this, rather than only being discoverable via these PR bodies.

No production Rust code touched — `cargo fmt`/`clippy`/`test` not applicable, confirmed.

Approving. Will merge once #1033 is in and this is retargeted to master with a fresh green CI signal.

PR #1035, opened 2026-07-17
fix(check): eu check --strict exits 0 despite a genuine parse error (eu-3621)

gmorpheme:
## Summary

`eu check --strict` printed a genuine rowan parse error (e.g. "malformed declaration head") to stderr but exited 0 — defeating the point of `--strict` as a CI gate. Root cause: `check_annotation_syntax` (`src/driver/check.rs`) iterated `parse_result.errors()` to `eprintln!` them but never counted them anywhere; the caller's `total_issues` tally (`syntax_errors + type_warning_count + content_error_count`) only ever saw annotation-string parse errors, type warnings, and content-verifier errors — a whole-file parse failure contributed to none of them.

Discovered via `tests/harness/typecheck/038_literal_string_annotation_widens.eu`, which contains `x: "hello" : "string"` — not valid syntax anywhere in Eucalypt (a second top-level colon is a malformed declaration head, not inline type ascription) — and had silently passed its `exit: 0` assertion since it was created, never actually exercising its stated purpose ("an explicit annotation on a string-literal binding widens it").

## Part 1 — the exit-code fix

`check_annotation_syntax` now returns `(annotation_syntax_error_count, file_parse_error_count)` instead of a single count. `check()` tracks the new `parse_errors` count separately and folds it into the hard-error branch alongside `content_error_count`:

```rust
} else if opt.check_strict() || content_error_count > 0 || parse_errors > 0 {
    Ok(1)
```

**Non-strict `eu check` behaviour decision**: I made a genuine parse error a hard error **regardless of `--strict`**, not just under `--strict`. Rationale: a parse error means the file never produced a syntax tree at all — it is categorically more fundamental than a type-annotation mismatch (which is advisory by design and only becomes a hard error under `--strict`). The plain `eu` eval path (no subcommand) already aborts unconditionally on a parse error with exit 1, with or without any flag — `eu check` reporting success for a file `eu` itself refuses to run would be a confusing, indefensible inconsistency. I verified this by hand before making the change:

```
$ eu tests/harness/typecheck/038_....eu                    # plain eval
EVAL EXIT: 1
$ eu check tests/harness/typecheck/038_....eu               # check, non-strict, BEFORE fix
CHECK (non-strict) EXIT: 0    <- inconsistent with eval
```

Documented in three places: a doc comment on `check`'s decision branch, the module-level doc comment at the top of `check.rs`, and a short note added to `docs/guide/type-checking.md` and `CLAUDE.md`'s Type Checking sections (both previously silent on parse-error handling).

### Regression tests (fault-injection verified)

1. **`src/driver/check.rs`, `driver::check::tests::check_exits_nonzero_on_genuine_parse_error_strict_and_non_strict`** — in-process test constructing `EucalyptOptions` via `EucalyptCli::try_parse_from` for both `check --strict` and `check` (non-strict) against a temp file with a genuine parse error, asserting `check()` returns `Ok(1)` in both cases.
2. **`tests/harness/typecheck/112_check_strict_parse_error_exits_nonzero.eu`** (+ `.expect`, `exit: 1` / `stderr: "malformed declaration head"`) — new CLI-level harness fixture exercising the bug through the real `eu` binary via `run_typecheck_test`, matching the exact "error-path fixture + exit-code assertion" shape used throughout `tests/harness/typecheck/`.
3. Existing `check_annotation_syntax` unit tests updated for the new tuple return type, plus a new `counts_file_parse_errors_separately_from_annotation_errors` test isolating the counting logic itself.

**Fault injection**: reverted `check()`'s decision branch to the pre-fix form (`total_issues = syntax_errors + type_warning_count + content_error_count`, `opt.check_strict() || content_error_count > 0`, no `parse_errors`). Confirmed both new regression tests **FAIL**:
- `driver::check::tests::check_exits_nonzero_on_genuine_parse_error_strict_and_non_strict`: `assertion left == right failed: ... must exit 1, got 0`
- `test_typecheck_112_check_strict_parse_error_exits_nonzero`: `left: 0, right: 1`

Restored the fix, re-ran both — **PASS**. `git diff` on `check.rs` was clean before restoring (confirmed no stray fault-injection artefacts survived).

## Part 2 — was inline ascription (`expr : "type-string"`) ever specified?

**No — and the spec is explicit about it.** `docs/development/gradual-typing-spec.md` §10 "Inline Type Assertions" (lines 820–854) names declaration-level metadata (`` ` { type: "..." } `` before a binding) as "the primary mechanism", then discusses an inline form under "Inline Assertions (deferred)": the proposed syntax was **double-colon** `expr :: "T"` (option 1, line 844), explicitly not implemented — *"Decision: defer inline assertions. Declaration annotations plus inference cover the primary use cases."* This is a different token from what 038 used. `::` has its own dedicated rejection: `src/syntax/rowan/error.rs`'s `ParseError::InvalidDoubleColon` (lines 161–167), wired at two call sites in `src/syntax/rowan/parse.rs` (564, 728), with a message that says outright *"eucalypt does not have type annotations"* for that form. **Single-colon** `expr : "type"` — what 038 actually contained — was never proposed anywhere in the spec; it simply collides with declaration-head grammar (a second top-level `:` reads as a second, malformed declaration head), which is why the parser's diagnostic is the generic "malformed declaration head" rather than the dedicated double-colon message.

Corroborating evidence: `docs/superpowers/reports/2026-07-13-parser-equivalence-audit.md`, a tree-sitter-vs-Rowan grammar audit, listed "inline type ascription" as a tree-sitter grammar gap citing 038 plus four sibling files — but that attribution was wrong, per Lantern's own same-day follow-up on the audit's parent bead `eu-2sa6.17` (comment dated 2026-07-17): 034/035/037's actual failure was an unrelated missing `\` in tree-sitter's `OPER_CHARS`; 054 was a `:io` binding-scope trailing-soup gap; and Lantern's comment states directly that 038 "is NOT a tree-sitter gap — Rowan itself rejects this file's `expr : "type"` syntax", which is exactly why eu-3621 was filed.

Conclusion: 038's `x: "hello" : "string"` is a fixture-authoring typo from commit `93df4449` (bead eu-19do, "A4 — literal string types"), not an attempt at a real or even proposed feature — the file's own stated purpose (declaration-level widening) points straight at the spec's actual primary mechanism. **No new syntax implemented; no bead needed for "spec'd but unimplemented"** — the deferral in §10 is already the record, and eu-2sa6.17 already closed out the tree-sitter-audit side of this correctly.

## Part 3 — 038 repaired

The real, currently-supported mechanism for widening a string-literal binding's type is the backtick metadata annotation on the declaration itself (`` ` { type: "string" } `` before `x: "hello"`), not inline ascription. Rewrote the fixture to:

```eu
` { type: "string" }
x: "hello"

` { type: "string → number" }
length: str.len

ok: length(x)

` { type: "number → number" }
double: (_ * 2)

bad: double(x)
```

This also fixed a second, previously latent bug the parse failure had been masking: the original fixture called `str.length`, which doesn't exist (the real prelude function is `str.len` — confirmed against `docs/reference/agent-reference.md`'s str-namespace table). Since the file never parsed, this was never actually exercised.

The rewritten fixture demonstrates both halves of "widening": the annotation lets `x` pass a `string`-typed function cleanly (`ok`), but widening to `string` doesn't make it usable as a `number` — `bad: double(x)` still warns (`expected number, found string`). The `.expect` now asserts `exit: 1` / `stderr: "expected number, found string"`, so the fixture joins the eval/check parity harness's covered set (previously it was `exit: 0`-only, which — as this whole bug demonstrates — cannot distinguish "the checker verified something" from "the checker silently skipped it").

## Test plan

- [x] `cargo test` — full suite green (1310 lib tests, 501 harness_test.rs tests, up from the prior baseline by the 2 new tests: 1 in `driver::check::tests`, 1 harness fixture 112)
- [x] `cargo test --test harness_test test_typecheck_038` / `test_typecheck_112` — both pass
- [x] `cargo fmt --all` — no diff
- [x] Fault injection performed and documented above, for both new regression tests
- [x] Manually verified `eu check --strict` / `eu check` / plain `eu` exit-code consistency on the release binary before and after the fix


gmorpheme:
**Wicket review — send back. Two things required before merge: one blocking CI failure, one accuracy correction to the written record.**

Reviewed in `/private/tmp/eu-wicket-1035` (head `1fcb576b` off master `6dc63d21`).

## 1. BLOCKING: CI is currently failing — "Corpus parse equivalence (eu-wrf7)"

`gh pr checks 1035` shows this check failing. Per the mandatory CI gate, I cannot merge with failing CI, no exceptions. Root cause, confirmed via the job log:

```
Corpus files checked: 354
Currently failing to parse: 29
Expected failures (seeded list): 29

INFO: 1 previously-expected failure(s) now parse successfully — consider pruning corpus-expected-failures.txt:
  - tests/harness/typecheck/038_literal_string_annotation_widens.eu

FAIL: 1 NEW divergence(s) — the tree-sitter grammar rejects source the Rowan parser accepts, and this isn't in the expected-failures list:
  - tests/harness/typecheck/112_check_strict_parse_error_exits_nonzero.eu
```

Two things need to happen in `editors/tree-sitter-eucalypt/test/corpus-expected-failures.txt` (not touched by this PR):

1. **Add `112_check_strict_parse_error_exits_nonzero.eu`** with a comment explaining why: it's a *deliberately* unparseable fixture (its whole purpose is exercising the parse-error-propagation bug), living in `tests/harness/typecheck/` rather than `tests/harness/errors/` — the corpus-check script's own docstring assumes every corpus file is valid eucalypt (that's why `tests/harness/errors/` is excluded from the scan entirely), and this new fixture violates that assumption from a directory the script doesn't exclude. This isn't a tree-sitter grammar gap at all — Rowan rejects this file too, by design — so it belongs in the expected-failures list with a note to that effect, not a grammar fix.
2. **Prune `038_literal_string_annotation_widens.eu`** from the list, since it's now valid syntax (the rewrite to backtick metadata form makes it parse successfully under tree-sitter too) — the script's own INFO nudge already points this out, and leaving it in leaves the list dishonest about what's actually still broken.

I'd suggest checking whether a similar `tests/harness/errors/`-style carve-out should extend to future deliberately-invalid `typecheck/` fixtures, but that's a follow-up, not a blocker — the expected-failures entry is sufficient for this PR.

## 2. Accuracy correction required (per the coordinator's dispatch, no code change needed — text only)

The PR body's Part 2 states: *\"Zero occurrences of an inline `expr : \"type\"` form anywhere in the spec.\"* This is incorrect. I checked directly:

- `docs/development/gradual-typing-spec.md` §11 (\"Inline Assertions (deferred)\", lines ~832-854) explicitly proposes `expr :: \"T\"` as option 1 (\"Dedicated syntax\") for a future inline type-assertion mechanism, and explicitly **defers** it (\"Decision: defer inline assertions... Revisit if there is demand\").
- `src/syntax/rowan/error.rs:161-167` has a dedicated `InvalidDoubleColon` parse-error variant with a helpful message (\"'::' is not valid syntax in eucalypt... '::' is used in Haskell/Rust for type annotations, but eucalypt does not have type annotations\") — direct evidence the parser anticipates and deliberately rejects exactly this class of attempt.

The overall conclusion (inline ascription was never implemented, and 038's original `expr : \"type\"` syntax specifically was never proposed anywhere, spec or otherwise) is still correct — the spec's proposal uses `::` (double colon), and 038's original content used a second single `:`, which is a different and even-less-founded pattern. But \"zero occurrences... anywhere in the spec\" overclaims: the spec does discuss and consciously defer an inline-assertion mechanism, just not the exact single-colon syntax 038 happened to use. Please correct the PR body (and the similar, milder phrasing in the CHANGELOG.md entry — \"no such inline type-ascription form was ever specified or implemented\" has the same issue) to cite §11 and its deferred `::` proposal, rather than implying the topic never came up.

## Everything else checks out

- **Fault injection reproduced independently**: reverted `check()`'s decision branch to the pre-fix form (`total_issues` without `parse_errors`, condition without `|| parse_errors > 0`). `check_exits_nonzero_on_genuine_parse_error_strict_and_non_strict` FAILED with the exact claimed message (`must exit 1, got 0`). Restored, rebuilt, reconfirmed both new unit tests pass.
- **Fixture 112**: passes as claimed.
- **Fixture 038**: ran directly on both paths myself (not just via the harness runner) — `eu check --strict` and plain `eu` eval both correctly warn/error with \"expected number, found string\" on `bad: double(x)`, while `ok: length(x)` (the widened `string` passed to a `string`-typed function) compiles clean with no warning. This genuinely exercises both halves of the widening claim (widening doesn't warn against a compatible type; it doesn't make the value usable as an incompatible one), not just an exit-code check.
- **Docs updates accurate**: `CLAUDE.md`, `docs/guide/type-checking.md` both correctly describe the new unconditional-hard-error behaviour for parse errors, consistent with what I verified in the code.
- **No CI/tooling dependency on the old non-strict exit-0 behaviour**: grepped `.github/workflows/` and `scripts/` for `eu check` usage — none found at all, so nothing depends on the changed behaviour.
- **Full suite, independently run**: `cargo test --release --lib` 1310/1310 (up from 1308 by the 2 new tests, matching the claim), `cargo test --release --test harness_test` 501/501 (up from 500 by fixture 112, matching the claim). `cargo fmt --all --check`: clean. `cargo clippy --all-targets -- -D warnings`: clean.

Once the corpus-expected-failures.txt update and the PR-body/CHANGELOG correction are in, this should be a quick re-review — the substance of the fix and the test coverage are solid.

gmorpheme:
**Wicket re-review — approved and merging.**

Both revision items verified:
1. **CI**: `Corpus parse equivalence (eu-wrf7)` now passes. Confirmed `corpus-expected-failures.txt` correctly prunes 038 (now valid syntax) and registers 112 with a rationale that matches the existing entries' style and cites the same "Rowan rejects it by design, kept listed here because the script's inclusion is directory-based" reasoning as its predecessor.
2. **Accuracy**: confirmed the corrected CHANGELOG wording precisely — `docs/development/gradual-typing-spec.md` §10 "Inline Type Assertions" (line 819) contains the "Inline Assertions (deferred)" subsection (line 833) proposing and deferring `expr :: "T"`, matching the citation now in the CHANGELOG entry exactly.

Also checked: no leftover merge-conflict markers anywhere in the tree, and the diff of this branch against the merged-master baseline (`d150d3d9`, post-#1032/#1033/#1034) is exactly and only this PR's own 9 intended files — the master merge pulled in nothing unexpected.

Full suite re-run on the final revision: `cargo test --release --lib` 1310/1310, `cargo test --release --test harness_test` 501/501, `cargo fmt --all --check` clean. CI fully green (all checks pass, release jobs correctly skipping).

Merging.

PR #1036, opened 2026-07-18
fix(tree-sitter): lib/prelude.eu and lib/lens.eu now parse with zero error nodes (eu-2sa6.17)

gmorpheme:
## Summary

eu-2sa6.17's continuation (owner ruling: stays open, real goal is
lib/prelude.eu parsing correctly in editors, scoped into 0.13.1). Do NOT
merge — awaiting gatekeeper review.

Re-verified the audit's §2.3 (bracket-expr declaration heads) and §2.7
(nested-brace type-DSL strings) against current master (post #1033/#1034
merge), then kept following the actual remaining blockers as each fix
revealed the next one downstream in `lib/prelude.eu` — landing on five
narrow, corpus-gated grammar changes across four audit sections.

## What changed

**§2.3 — bracket_expr as a `declaration_head` alternative.** `⟦ x ⟧: x` /
`⌈[x]⌉: x ceiling` (lib/prelude.eu's ceiling/floor notation) were never
reachable as declaration heads — `bracket_expr` already parsed their
content fine as an ordinary expression, it just wasn't wired into
`declaration_head`'s choice list. One-line addition, with a
`[$.declaration_head, $._element]` conflict declared to resolve the
resulting (expected) GLR ambiguity against bracket_expr-as-soup-element.

**§2.7 — nested braces inside string interpolation, two separate bugs**
under one audit heading:
- `interpolation_content` was restricted to `anaphor | dotted_reference |
  identifier | number` — the row-polymorphic type-DSL idiom `{..r}`
  (lib/prelude.eu's `merge()` doc, `s"{..r} → {..s} → {..r, ..s}"`) isn't
  any of those. Broadened to a general `$.soup`, comma-separated (a bare
  `,` isn't part of `$.soup`'s own grammar, threaded through explicitly).
- `format_spec`'s `/:[^}]+/` regex can't tell an outer `}` from an inner
  one, so `s"{ users: [{ name: string }] }"` (real content in
  `tests/harness/175_sv2_to_spec.eu`, structurally identical to
  `lib/lens.eu`'s row-polymorphic Lens type strings) stopped at the first
  `}` it saw. Made brace-aware via a small recursive `_balanced_braces`
  rule instead of a flat regex.

**§2.6 — destructuring pattern gaps, a NEW prelude blocker** not
originally flagged by the audit for `lib/prelude.eu` — `update-nth`'s
`{ rejoin([prefix, [s: suffix]]): ... }` uses a `cons_pattern` nested
inside a `list_pattern`. `list_pattern`/`cons_pattern`/`block_pattern`'s
renamed fields only accepted a single level of bare `$.identifier`. Made
`_param` recursive through all three, covering multi-head cons
(`[a, b : rest]`), nested lists (`[a, [b, c]]`), nested cons-in-list
(`[a, [b: c]]`), and nested blocks (`{outer: {inner}}`) — all confirmed
real syntax via `tests/harness/092/095/102/141_*.eu` and `lib/prelude.eu`
itself. `block_pattern`'s renamed-field production is now its own visible
`block_pattern_field` rule (needed both for correct nesting — `{outer:
{inner}}`'s `{inner}` is only valid as a block_pattern, not a general
block expression — and to keep `queries/highlights.scm`'s
block-pattern-field-name query resolvable; updated it, which is the one
non-grammar file this PR touches).

**§2.9 — Letterlike Symbols identifiers, also a NEW prelude blocker** —
`ℕ: iota(0)` is real prelude syntax, not just `examples/aoc25/`. Added
U+2100-U+214F to the identifier regex (the analogous *identifier*-
character gap to eu-fbyk's already-flagged *operator*-character
hardcoding). Bonus: this also resolved `examples/aoc25/day12.eu`, whose
root cause the original audit couldn't isolate.

## Verification against eu-2dxe

Wicket's review of #1034 flagged that a corpus-parse exit code alone
cannot rule out a silent zero-error-node misparse. Did not just trust
exit 0 here:
- tree-sitter's `lib/prelude.eu` parse has exactly **295 top-level
  `declaration` nodes and 1 top-level `unit_metadata` block**, matching
  `eu dump ast lib/prelude.eu`'s 295 top-level DECLARATION nodes and 1
  top-level BLOCK_META **exactly**.
- Individually spot-checked, against expected shape: both
  ⌈[x]⌉/⌊[x]⌋ bracket_expr declarations; `merge()`'s
  `s"{..r} → {..s} → {..r, ..s}"` interpolation (two comma-separated soups
  in the third interpolation, correctly NOT merging into the neighbouring
  `merge: __MERGE` declaration); `ℕ: iota(0)` (a single 3-byte identifier
  token); `rejoin([prefix, [s: suffix]])`'s nested cons-in-list parameter
  pattern.

## Divergence count

**20 → 4** of 354 corpus files failing to parse. The 4 remaining
(documented in `corpus-expected-failures.txt`) are unrelated to
prelude/lens: the pre-existing 023/024 bare-top-level-expression gap and
`010_prelude.eu`'s §2.8 chained-call gap (both out of scope, not
investigated here), plus the eu-3621 deliberately-invalid regression
fixture (Rowan rejects it by design).

## Does prelude genuinely parse clean?

**Yes.** `lib/prelude.eu` and `lib/lens.eu` both parse with zero ERROR
nodes, and the declaration-count/spot-check verification above confirms
this isn't a silent misparse — the structure genuinely matches Rowan's.

## Test plan

- [x] `tree-sitter test` — 58/58 pass (9 new regression cases on top of
      the 49 from #1033/#1034)
- [x] `python3 scripts/check-tree-sitter-corpus.py` — green, 4/4
- [x] **Fault-injection verified**, all 6 grammar changes, individually:
      each reverted alone, confirmed the relevant new test(s) fail (and
      nothing else does), restored, confirmed full green again
- [x] `lib/prelude.eu` / `lib/lens.eu` structurally spot-checked (see
      "Verification against eu-2dxe" above), not just exit-code-checked
- [x] Cross-checked with both the `package.json`-pinned `tree-sitter-cli`
      (`^0.22.0`, via `npx`) and the Homebrew-installed 0.26.11 — both
      agree
- [x] `queries/highlights.scm` query validity confirmed (the
      `block_pattern_field` rename broke one query pattern — an
      "Impossible pattern" error tree-sitter's own test runner caught;
      fixed and reverified)


gmorpheme:
**Wicket review — approved and merging.** This was the highest-risk PR in the queue (grammar broadenings, exactly the class that produced #1034's near-miss), so I went deep — independent structural verification beyond just reproducing the claims, plus investigating one genuine (but pre-existing, non-blocking) modeling gap I found along the way.

Reviewed in `/private/tmp/eu-wicket-1036` (head `6d269319` off master `cc36f05b`).

## Baseline claims, all independently reproduced exactly

- `tree-sitter test`: 58/58.
- `python3 scripts/check-tree-sitter-corpus.py`: 4/354 failing, 4 expected, 0 new divergences.
- `corpus-expected-failures.txt` delta: regenerated via `--write-expected` (which writes the bare filename list, no rationale comments) and got the **exact same 4-file set** (`010_prelude.eu`, `023`/`024_for_*_no_warning.eu`, `112_check_strict_parse_error_exits_nonzero.eu`) as the PR's hand-annotated version — confirms nothing was silently added or dropped.

## Structural verification beyond exit codes (the review's core question)

**295-declaration count, independently reproduced** — with one methodology correction along the way worth recording: `eu dump ast lib/prelude.eu` (no `-Q`, which `dump` doesn't support) dumps the explicit file **and** a second implicit auto-loaded prelude copy back-to-back in the same output (since `lib/prelude.eu` *is* the standard prelude, loaded once explicitly and once implicitly). My first count came to 590 — almost exactly double — until I isolated just the first `UNIT` section. Sliced correctly: **295 top-level `DECLARATION` + 1 `BLOCK_META`**, matching tree-sitter's own 295+1 exactly.

**Individual construct spot-checks, all structurally sound**, comparing tree-sitter's parse against Rowan's `eu dump ast` byte-range-adjacent output:
- `⌈[x]⌉: x ceiling` — `bracket_expr` decl head wrapping a list containing `x`, then a two-name soup body (`x`, `ceiling`) — isomorphic to Rowan's `BRACKET_EXPR` wrapping `LIST`/`SOUP`.
- `merge()`'s `s"{..r} → {..s} → {..r, ..s}"` — the third interpolation correctly splits into **two** comma-separated soups (`..r`, `..s`), and critically the *following* `merge: __MERGE` declaration is correctly split as its own separate node, not swallowed into the interpolation (the exact failure mode #1034's near-miss produced elsewhere).
- `ℕ: iota(0)` — a single 3-byte identifier token, matching ℕ's UTF-8 encoding exactly.
- `rejoin([prefix, [s: suffix]])` — `list_pattern` containing `identifier "prefix"` + nested `cons_pattern` containing `identifier "s"` + `identifier "suffix"` — correctly recursive.

Also independently confirmed (per the review dispatch's explicit ask) that `175_sv2_to_spec.eu` and `lib/lens.eu` both parse with zero ERROR nodes, and spot-checked `175`'s nested-brace case (`s"{ users: [{ name: string }] }"`) — the brace-aware `format_spec`'s span correctly captures the full nested content without stopping at the first inner `}`.

## Fault injection, reproduced independently (two required, did these two plus a bonus generator-level check)

1. **Pattern recursion** (`list_pattern`/`cons_pattern` reverted from `$._param` back to their pre-PR `$.identifier`-only form): 58/56 → 2 new failures in `tree-sitter test`, and critically — `lib/prelude.eu` itself now fails to parse, with the ERROR landing at **exactly** row 1383 (the `rejoin([prefix, [s: suffix]])` line), confirming this fix is genuinely load-bearing for the stated goal.
2. **Brace-aware `format_spec`** (reverted to the pre-PR flat `/:[^}]+/` regex): 58/57 → 1 new failure, and `175_sv2_to_spec.eu` fails with the ERROR landing at exactly the nested-brace boundary `[65,45]-[65,46]`.
3. **Bonus**: removed the new `[$.declaration_head, $._element]` conflict declaration entirely — `tree-sitter generate` **fails outright** (refuses to build) rather than silently resolving it. This is the opposite risk profile from #1034's near-miss: that ambiguity generated silently with zero warning; this one is explicitly surfaced and required by the generator itself, confirming it's a genuine, acknowledged local ambiguity (bracket_expr as declaration_head vs. as an ordinary soup element), not a hidden one.

## `queries/highlights.scm`

Confirmed only the one query the PR touches actually needed updating (`block_pattern`-wrapping-`declaration` → `block_pattern_field`, since renamed/nested block-pattern fields are now their own distinct node type rather than reusing `declaration`). Checked the surrounding `list_pattern`/`cons_pattern` queries specifically for whether they'd need similar updates for the new recursion — they don't: tree-sitter's query matching recurses into nested same-shaped nodes for free, so `(list_pattern (identifier) @parameter)` still correctly highlights identifiers at every nesting depth without modification. Ran `tree-sitter query queries/highlights.scm lib/prelude.eu` directly against the full file — completes cleanly, no "Impossible pattern" or other errors.

## One finding worth recording, not blocking

While spot-checking the `merge()` interpolation, I found that tree-sitter has **no dedicated `s_string` grammar rule** — unlike `t_string`, which is correctly modelled as fully opaque (matching Rowan's own "T-strings are ZDT literals — no interpolation, just validate and emit"), `s"..."` strings fall through to the generic `string` production with full interpolation support. Rowan's lexer, by contrast, explicitly treats S-strings identically to T-strings: *"S-strings are type-data literals — no interpolation, no escape processing"* (`src/syntax/rowan/lex.rs`) — a single opaque `S_STRING` token, zero internal decomposition. This is a genuine structural divergence in how the two grammars model type-DSL strings' internals.

**This is pre-existing, not introduced by #1036** — the `string`/`interpolation` mechanism was already being applied to s-string content before this PR (that's *why* the row-polymorphism syntax `{..r}` previously produced an ERROR node at all — tree-sitter was already trying to parse it as interpolation, just too narrowly). This PR's fix widened what that pre-existing mechanism accepts; it didn't introduce the "s-strings aren't truly opaque" gap. It doesn't affect this PR's actual claims (zero ERROR nodes, declaration-count match) since both hold regardless of how the string's *interior* is labelled. But it's worth a follow-up bead — companion to eu-fbyk's operator-character gap and the audit's §2.9 identifier-character gap — since any tooling built on the tree-sitter grammar that assumes type-DSL string content is inert (e.g. syntax highlighting not treating `..r` as if it were live code with operators) would currently get it wrong.

## Verdict

Substance is solid, all six changes independently verified as load-bearing and correctly scoped, no silent-misparse risk found for the actual claims made. CI fully green. Merging.

PR #1049, opened 2026-07-22
docs: Phase 2 diagnostics plan (curated trace + blame annotations)

gmorpheme:
## Summary

Implementation plan (docs-only, no code) for **Phase 2 of the 0.14 Diagnostics Overhaul epic** (`eu-1tkk.7`) — curated trace + blame annotations, spec §4.3. Covers beads `eu-1tkk.7.11` (blame annotation vocabulary) and `eu-1tkk.7.12` (frame classifier + curation pipeline + `--debug-trace`).

This is a **draft for owner review**. No implementation code is included; per Clarion's two-phase workflow, implementation begins only after approval.

## Key findings from investigating the real code before planning

- **Feasibility verdict: the blame-annotation mechanism already half-exists.** `Desugarer::new_smid` (`src/core/desugar/desugarer.rs:394`) already tags every prelude Smid with its declaring function's name, and this is verifiable live (`EU_ERROR_TRACE_DUMP=1`) — so a name-keyed classifier is feasible and low-risk **on the source-compiled-prelude path** (dev builds, this repo's default `cargo test`).
- **But it doesn't reach the shipped release binary.** `lib/prelude.blob` — what released binaries embed — is generated with `generate_annotations: false` (`xtask/src/main.rs:332`), stripping all Smid/name identity from every prelude combinator. This is a documented, deliberate invariant (`src/common/sourcemap.rs:14`), not a bug, but it means Phase 2 as specced has no effect on real users' binaries without additional plumbing (scoped as Task 2b, gated on owner sign-off — it touches hot-path VM trace collection and possibly the blob wire format).
- **A second, independent gap:** live-verified via `EU_ERROR_TRACE_DUMP=1` that for the epic's three flagship xfail specimens (`nth_out_of_range`, `hof_bad_arg`, `swap_args`), the raw scraped trace contains **zero** user-file frames — not a prioritisation bug, an absence. Curation over "the raw material already scraped" (as the spec puts it) cannot manufacture a user primary that was never captured. Scoped as an investigation checkpoint (Task 0) before the curation pipeline (Task 3) is built.

Both findings are written up with file:line evidence in the plan's "Feasibility verdict" section, and surfaced as Open Questions 1–2 for the owner to resolve before implementation.

## Plan structure

9 tasks (checkbox/TDD format, mirroring the Phase 0 plan's template): investigation checkpoint → blame vocabulary (.7.11) → frame-name reachability → curation pipeline (.7.12) → boundary-frame rephrasing (`nth`'s own error) → `--debug-trace` → xfail-flipping + new corpus fixtures → docs → close-out. Phase 3 (laziness-surviving cause-trace) is explicitly excluded per the design spec.

## Test plan

N/A — docs-only change. No code, no tests to run.

### DIFF
--- a/docs/superpowers/plans/2026-07-22-diagnostics-phase2-curated-trace.md
+++ b/docs/superpowers/plans/2026-07-22-diagnostics-phase2-curated-trace.md
@@ -1,9 +1,32 @@
 # Diagnostics Phase 2 — Curated Trace & Blame Annotations: Implementation Plan
 
-> **Status: DRAFT for owner review — do not implement without sign-off.**
-> **For agentic workers (once approved):** REQUIRED SUB-SKILL: use
+> **Status: APPROVED AND IN PROGRESS.** The owner approved Phase 2 and
+> implementation is under way; this document is the durable record of the
+> plan and of how it turned out. It is no longer a draft, and no task in it
+> is gated on further sign-off.
+> **For agentic workers:** REQUIRED SUB-SKILL: use
 > `superpowers:subagent-driven-development` or `superpowers:executing-plans`.
 > Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Task status at a glance** (updated 2026-07-23, master `8b501ada`):
+
+| Task | Subject | Status |
+|---|---|---|
+| 0 | Trace-capture gap investigation | **Shipped** — findings acted on; see Task 0 |
+| 1 | Blame annotation vocabulary (`.7.11`) | **Shipped** — PR #1051, PR #1052 |
+| 2 | Frame classifier, source-prelude path (`.7.12`) | Open |
+| 2b | Blob-path frame identity | **Shipped** — PR #1052, wire v5 |
+| 3 | Curation pipeline (`.7.12`) | Open |
+| 4 | Boundary-frame rephrasing (`nth`'s own error) | Open |
+| 5 | `--debug-trace` | Open |
+| 6 | Flip corpus xfails + coverage fixtures | Partial — `nth_out_of_range` flipped (PR #1050) |
+| 7 | Documentation | Open |
+| 8 | Close-out | Partial — `.7.11` and `.7.18` closed; `.7.12` open |
+
+Beads `eu-1tkk.7.11` (blame vocabulary) and `eu-1tkk.7.18` (Update-continuation
+call-site capture, filed out of Task 0's investigation) are **closed**.
+`eu-1tkk.7.12` (frame classifier + curation pipeline + `--debug-trace`) remains
+open and carries Tasks 2, 3, 5.
 
 **Goal:** Replace today's fragile `is_user_file`/env-trace heuristics with an
 explicit per-combinator blame contract (transparent/boundary), and use it to
@@ -32,6 +55,13 @@ are both live on master as of `90c713ab`.
 ---
 
 ## Feasibility verdict — blame-annotation threading (read this before scoping tasks)
+
+> **Superseded in part (2026-07-23).** The two-path asymmetry described below
+> was accurate when written and is what motivated Task 2b. It **no longer
+> describes master**: PR #1052 closed the blob-path gap, so blame
+> classification data now reaches both prelude loading paths. The analysis is
+> retained because it is the reasoning that justified Task 2b, not because it
+> is current. See "Task 2b (SHIPPED)" for what was actually built.
 
 **The mechanism exists and already works, but only on one of the two prelude
 loading paths.** This is the single most important finding for scoping the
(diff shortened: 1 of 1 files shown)


gmorpheme:
**Recorded non-author review (wicket).** Reviewed in an isolated worktree against master `8b501ada`.

Routing note: clarion ran here in DIRECTED mode on a coordinator-dispatched, owner-sanctioned brief, so this takes the normal gatekeeping route. This PR is docs-only and touches no sensitive surface, so CLAUDE.md's recorded-review rule does not independently bind it — this comment is the review record regardless.

### Assessment of the plan on its merits

Checked against design spec §4.3. The plan is a **faithful and complete** decomposition of its mandate: blame vocabulary (Task 1) → classifier (Task 2) → curation pipeline classify/drop/collapse/keep/budget (Task 3) → boundary rephrasing (Task 4) → `--debug-trace` (Task 5). §4.3's secondary-label requirement ("in 'nth'") is modelled in Task 3's `CuratedTrace.boundary`; the `compress_trace_cycles` reuse the spec calls for is Task 3 Step 2; the budget is tied back to the invariant gate's `TRACE_BUDGET`. Phase 3 is correctly excluded per §4.4 and, importantly, the plan pre-commits to *not* absorbing Phase 3 work if Task 0 surfaced it (Risk 2) — that discipline held in practice.

Its two additions beyond the spec — Task 0 (trace-capture gap) and Task 2b (blob-path identity) — were both real gaps the spec had not anticipated, and both proved load-bearing: Task 0 became `eu-1tkk.7.18`/PR #1050, Task 2b became PR #1052. Investigating the actual code before planning, with `file:line` evidence and live `EU_ERROR_TRACE_DUMP=1` verification, is what surfaced them. That is the strongest thing about this document.

Accuracy of the technical claims, verified independently against master rather than taken on trust: the `generate_annotations: false` blob invariant, the `Desugarer::new_smid` naming mechanism, the zero-user-frame measurements, and the `Smid::default()` collision rationale all check out as written. One prediction was **wrong, in the safe direction** — see below.

### What I changed (2f2bce30) and why

The document was stale in ways that had already caused a false merge-block on #1052, so I corrected the record rather than merging it as-is.

**Status framing.**
- The "DRAFT for owner review — do not implement without sign-off" heading is replaced with an accurate approved/in-progress status plus a per-task status table.
- **Task 2b** is retitled from "OPEN QUESTION — do not start without owner sign-off" to **SHIPPED**, recording the decision actually taken (*full blob plumbing in 0.14*, wire v4→v5) and where the sign-off lives (comment on #1052; comment on bead `eu-1tkk.7.11` dated 2026-07-23). The do-not-start wording is explicitly **withdrawn** so it cannot be re-applied. Its gated steps are ticked with what was built.
- Tasks 0 and 1 marked SHIPPED with PR numbers; Task 6 PARTIAL; Open Questions 1–5 each annotated with resolution state.

**Superseded reasoning corrected, not left standing as current.** Per the brief, the record should not preserve overtaken analysis as though it were live:
- Task 2b predicted a non-Smid parallel identity channel touching hot per-call-site continuation/env code in both engines. It was instead built as a **reserved `Smid` sub-range** (`Smid::global_slot` / `GLOBAL_SLOT_TAG`), so traces stayed `Vec<Smid>` end-to-end, the work is load/error-time only, the evaluation hot path is untouched, and it measured perf-neutral. The wire-format half of the prediction *was* right, and the recorded-review rule was engaged and satisfied. Both halves are now stated.
- Task 0's answer was capture-side and local — closest to its own hypothesis (b), annotation minted but not retained past `Update` continuations — fixed by #1050. `nth_out_of_range` flipped off xfail; `hof_bad_arg` and `swap_args` verified *not* fixed by it. Consequently **Open Question 3** (boundary-location carve-out of invariant (i)) is narrowed to those two cases, not resolved — I was careful not to overstate that.
- Task 1 shipped via a desugar-phase side channel (`Desugarer::record_blame` → `TranslationUnit::blame`) rather than the sketched `rowan_ast.rs` read, and annotates **11** combinators, not the 12 listed (`map2` was not annotated).

**One finding beyond the brief — review routing.** The plan carried, as a global constraint and again at Task 8 Step 3, *"PRs go to the owner personally for review (Clarion policy, not Wicket)"*. No such policy exists. This is precisely the hold that was attached to #1050 and then **audited as agent-fabricated** and released via the coordinator. Leaving it in a checked-in plan would keep manufacturing the same void hold on every future PR from this document, so I corrected it to the normal Wicket route, added an explicit instruction not to re-apply such a hold, and stated the genuine recorded-review rule where it does apply.

### Gates

Docs-only, no code touched, so `cargo fmt`/`clippy`/`test` are not engaged. Verified `git diff --stat` shows a single file. Every claim I added was checked against master: `BlameSpec`/`extract_blame_spec` in `src/core/metadata.rs`; `PreludeBlob.blame` + `slot_name`/`blame_for`/`classify` in `src/eval/stg/blob.rs`; `Smid::global_slot`/`as_global_slot` and the `source_info_for_smid` refusal in `src/common/sourcemap.rs`; `reconstruct_form_annotated` in `arena.rs`; the 11 `blame:` annotations in `lib/prelude.eu`; the three corpus `.meta.toml` states; and the absence of `--debug-trace`, `classify_frame` and `curate_trace`, confirming Tasks 2/3/5 have not shipped.

**Not closing any bead** — `eu-1tkk.7.12` remains open and is the coordinator's to close against the spec.

Approving and merging once CI is terminal-green.

PR #1052, opened 2026-07-23
feat(diagnostics): thread prelude blame table through the blob path (eu-1tkk.7.11)

gmorpheme:
## Summary

Phase 2 of the 0.14 Diagnostics Overhaul (eu-1tkk.7.11 part 2, closing the blob half). The blame vocabulary (`BlameSpec`, `extract_blame_spec`, `:transparent`/`:boundary` metadata) merged in #1051 only reached the source-compiled-prelude path (per the checkpoint spike `spike/furnace-diagnostics-phase2-blob-scope`, commit `8b468dba`, and the Phase 2 plan `docs/superpowers/plans/2026-07-22-diagnostics-phase2-curated-trace.md`). This PR makes that material *reachable* on the shipped, blob-based release binary — it does not implement the Phase 2 classifier itself (eu-1tkk.7.12, still open) and does not flip any corpus fixtures.

- **`lib/prelude.eu`**: declares `` ` :transparent `` on `map`/`filter`/`foldl`/`foldr`/`drop`/`mapcat`, `` ` :boundary `` on `nth`/`head`/`tail`/`lookup`/`lookup-or` — merged into each declaration's existing `{ doc:, type: }` metadata block (backtick-attaches-to-next-declaration honoured).
- **`TranslationUnit::blame` / `Desugarer::record_blame`**: a new side channel mirroring `deprecations`. `blame` is desugar-phase-only metadata (stripped by `strip_desugar_phase_metadata` before it could survive as runtime `Expr::Meta`), so `cargo xtask prelude-compile` reads it via `loader.core().blame` — independent of `generate_annotations`/Smid generation entirely.
- **`PreludeBlob.blame: HashMap<String, FrameKind>`**: reconciles the source `BlameSpec` vocabulary to the blob/classifier `FrameKind` vocabulary (`common::diagnostic_json`), populated in xtask's per-binding loop, tolerant of combinators with no declared contract.
- **`Smid::global_slot` / `Smid::as_global_slot`**: a disjoint, collision-free sub-range of the `Smid` `u32` space (top bit tagged) identifying "which prelude global slot" rather than "which source position" — safe to restore at blob-mode reconstruction without resurrecting a raw xtask-sourced Smid that would index into the wrong `SourceMap`.
- **`StgArena::reconstruct_form_annotated`**: stamps this identity onto every reconstructed `Lambda` form (the historical `reconstruct`/`reconstruct_form` are untouched — same `Smid::default()` behaviour). Wired in at exactly the two blob-mode reconstruction chokepoints named in scope: `StandardRuntime::globals()` and xtask's bytecode pre-encode loop, so the HeapSyn and bytecode engines agree.
- **`PreludeBlob::classify`/`slot_name`/`blame_for`**: the classifier-facing API for eu-1tkk.7.12 — given a trace `Smid`, `as_global_slot()` → slot → name (via `name_to_slot`) → blame class.
- **Wire format**: bumps `BYTECODE_WIRE_FORMAT_VERSION` to **v5** in both `build.rs` and `xtask/src/main.rs` (kept in sync, doc comments updated). Confirmed the stale-blob fallback still degrades to source-compile, not a hard failure.

## Verified live

The design spec's own flagship specimen (§4.3): `xs nth(10)` on a 3-element list. Under the default bytecode+blob engine, `EU_ERROR_TRACE_DUMP=1` now shows a stack-trace entry whose Smid decodes (via `as_global_slot`) to slot 236, which resolves through the *embedded* blob's own `name_to_slot`/`blame` tables to `nth`, declared `Boundary` — exactly the class of bug (`[prelude]:NNNN` with zero identity) the epic exists to fix. Two new integration tests in `tests/diagnostics_blame_plumbing_test.rs` gate this end-to-end against the real `eu` binary + embedded blob (not just the blob's static tables).

**Hot path**: load-time/build-time only, as designed. Confirmed the pre-encoded bytecode `program.code` byte length is unchanged (269213 bytes, before and after) — the Smid stamping happens purely at reconstruction time from data (the enumeration index) already present in the blob, adding zero bytes to the compiled code stream. Blob size delta: **596625 → 596703 bytes (+78 bytes)**, entirely attributable to the new `PreludeBlob.blame` table (11 entries).

## Fault-injection verification (per epic requirement)

Every new assertion was broken, confirmed FAIL, restored, confirmed PASS:
- `Smid::as_global_slot` (forced `None`) → `sourcemap::tests` round-trip/collision tests failed.
- `PreludeBlob::classify` (forced `None`) → `blob::tests::slot_name_and_blame_for_and_classify_compose` and `embedded_blob_has_declared_blame_for_nth_and_map` failed.
- `Desugarer::record_blame` (made a no-op) → blob regenerated with 0 blame entries; `embedded_blob_has_declared_blame_for_nth_and_map` and the stronger live-trace integration test both failed (the weaker "carries *a* global-slot Smid" test correctly still passed, since that mechanism is independent of blame declarations).
- `StgArena::reconstruct_form_annotated` (made a no-op, falling back to `None`) → blob regenerated; both new integration tests in `diagnostics_blame_plumbing_test.rs` failed.

## Test results

- `cargo fmt --all --check`: clean.
- `cargo test` (bytecode, default): green, **except two pre-existing failures unrelated to this change** — confirmed via `git stash` + rebuild that both reproduce **identically on unmodified `origin/master`** with a freshly regenerated blob:
  - `diagnostics_invariants.rs::corpus_satisfies_invariants` — a stale `xfail` marker on `tests/diagnostics/corpus/metadata_span.eu` (the fixture's JSON trace is already empty with a user-file primary on master; `metadata_span.meta.toml`'s xfail reason describes source-path-only behaviour and needs a maintainer to update the sidecar — out of scope here).
  - 5 `harness_test.rs` typecheck tests (010/011/013/063/085) — pre-existing checker/eval divergence tracked by eu-ntwg.1.
- `EU_HEAPSYN=1 cargo test`: same result (same two pre-existing failures, nothing new).
- `EU_GC_VERIFY=2 EU_GC_STRESS=1 cargo test --lib` and `--test harness_test`: green (same two pre-existing harness failures, no GC-related issues).

## Not in scope (per dispatch)

Did not touch `swap_args`/`DirectApp`, #1050's continuation code, or clarion's error.rs classifier. Did not attempt to flip any corpus fixtures or implement curation — that's eu-1tkk.7.12.

## Review

Blob wire format + `Smid` semantics are shared with clarion's Phase 2 classifier work — **recorded review from a non-author required before merge**, per CLAUDE.md's "Recorded review before merge" rule (GC/memory, blob wire format). Do not merge without that review comment on the PR.

https://claude.ai/code/session_01Py2HEaF87JJWPcDBiBic3g

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to eucalypt are documented here.
 
 ## [Unreleased]
 
+### Added
+
+- **Prelude blame-table plumbing reaches the shipped (blob-based) release binary (eu-1tkk.7.11 Phase 2)** — `` ` :transparent ``/`` ` :boundary `` blame declarations (merged in #1051) now flow through to the pre-compiled `lib/prelude.blob`: `map`/`filter`/`foldl`/`foldr`/`drop`/`mapcat` are declared `:transparent`, `nth`/`head`/`tail`/`lookup`/`lookup-or` are declared `:boundary`; `cargo xtask prelude-compile` reconciles the desugar-time `BlameSpec` vocabulary into a new `PreludeBlob.blame: HashMap<String, FrameKind>` table. A new `Smid::global_slot`/`Smid::as_global_slot` pair reserves a disjoint sub-range of the `Smid` `u32` space for "which prelude global slot", stamped onto each blob global's reconstructed lambda form at the two blob-mode reconstruction chokepoints (`StandardRuntime::globals()`, xtask's bytecode pre-encode loop) instead of the historical `Smid::default()` — restoring enough identity for a blob-mode error trace to name the prelude combinator responsible, without resurrecting a raw xtask-sourced Smid that would index into the wrong `SourceMap`. `PreludeBlob::classify`/`slot_name`/`blame_for` give the Phase 2 frame classifier (eu-1tkk.7.12, not yet implemented) the trace-Smid → declared-blame-class lookup it needs. Verified live: the design spec's own flagship specimen (`xs nth(10)` on a 3-element list) now carries a global-slot Smid in its blob-mode trace that resolves to `nth`, declared `Boundary` — exactly the case `[prelude]:NNNN`-with-zero-user-frames was meant to fix. Load-time/build-time only; the VM tick/allocation hot path is unaffected. Bumps `BYTECODE_WIRE_FORMAT_VERSION` to v5 (stale-blob fallback to source-compile still applies to old blobs)
+
 ## [0.13.1] - 2026-07-20
 
 ### Changed
--- a/build.rs
+++ b/build.rs
@@ -25,9 +25,17 @@ use std::path::Path;
 /// - v4: `PreludeBlob::type_summary` (`PreludeSummary`) field removed — it
 ///   was write-only (sole writer `xtask`, zero readers; its only consumer
 ///   was deleted in PR #1012) (eu-2sa6.20).
+/// - v5: `PreludeBlob::blame` field added (binding name → declared
+///   `:transparent`/`:boundary` classification), and blob-mode global
+///   reconstruction (`StandardRuntime::globals()`, xtask's bytecode
+///   pre-encode loop) now stamps each prelude global's `LambdaForm::
+///   Lambda.annotation` with a `Smid::global_slot(..)` identity instead of
+///   always `Smid::default()` — not a serialised-shape change on its own,
+///   but bundled into the same version bump as the `blame` field it feeds
+///   (eu-1tkk.7.11).
 ///
 /// MUST match `BYTECODE_WIRE_FORMAT_VERSION` in `xtask/src/main.rs`.
-const BYTECODE_WIRE_FORMAT_VERSION: u32 = 4;
+const BYTECODE_WIRE_FORMAT_VERSION: u32 = 5;
 
 /// Compute the blob source hash: `SHA-256(prelude source ‖ wire-format version)`.
 fn blob_source_hash(source_bytes: &[u8]) -> [u8; 32] {
--- a/lib/prelude.eu
+++ b/lib/prelude.eu
@@ -297,7 +297,8 @@ snoc(x, l): l ++ [x]
 (x ‖ xs): __CONS(x, xs)
 
 ` { doc: "`head(xs)` - return the head item of list `xs`, panic if empty."
-    type: s"[a] → a" }
+    type: s"[a] → a"
+    blame: :boundary }
 head: __HEAD
 
 ` { doc: "`↑xs` - return first element of list `xs`. Tight-binding prefix operator."
@@ -329,7 +330,8 @@ coalesce(xs): (xs nil?) then(null, (xs head)✓ then(xs head, xs tail coalesce))
 head-or(d, xs): xs nil? then(d, xs head)
 
 ` { doc: "`tail(xs)` - return list `xs` without the head item. [] causes error."
-    type: s"[a] → [a]" }
+    type: s"[a] → [a]"
+    blame: :boundary }
 tail: __TAIL
 
 ` { doc: "`tail-or(xs, d)` - return list `xs` without the head item or `d` for empty list."
@@ -421,7 +423,8 @@ block: __BLOCK
 has(s, b): b map-values(const(true)) lookup-or(s, false)
 
 ` { doc: "`lookup(s, b)` - look up symbol `s` in block `b`, error if not found."
-    type: s"symbol → Dict(a) → a?" }
+    type: s"symbol → Dict(a) → a?"
+    blame: :boundary }
 lookup(s, b): __LOOKUP(s, b)
 
 ` { doc: "`lookup-in(b, s)` - look up symbol `s` in block `b`, error if not found."
@@ -429,7 +432,8 @@ lookup(s, b): __LOOKUP(s, b)
 lookup-in(b, s): __LOOKUP(s, b)
 
 ` { doc: "`lookup-or(s, d, b)` - look up symbol `s` in block `b`, default `d` if not found."
-    type: s"symbol → a → {..} → any | a" }
+    type: s"symbol → a → {..} → any | a"
+    blame: :boundary }
 lookup-or(s, d, b): __LOOKUPOR(s, d, b)
 
 ` { doc: "`lookup-or-in(b, s, d)` - look up symbol `s` in block `b`, default `d` if not found."
@@ -1325,7 +1329,8 @@ __dbg-after(f, x): ▶(f(x))
 take(n, l): __IF((n zero?) ∨ (l nil?), [], cons(l head, take(n dec, l tail)))
  
 ` { doc: "`drop(n, l)` - return result of dropping integer `n` elements from list `l`."
-    type: s"number → [a] → [a]" }
+    type: s"number → [a] → [a]"
+    blame: :transparent }
 drop(n, l): __IF((n zero?), l, drop(n dec, l tail))
 
 ` { doc: "`rotate(n, l)` - rotate list `l` left by `n` positions."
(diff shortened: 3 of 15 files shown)


gmorpheme (commented):
## Wicket review — needs changes (do not merge)

Full independent review in an isolated worktree (`/tmp/eu-wicket-review-1052`, PR HEAD `164e71ae`), plus a from-scratch master baseline (`/tmp/eu-wicket-master-cmp`, `c1168087`) for side-by-side comparison. This satisfies the recorded-review gate for the blob-wire-format/`Smid`-semantics surface.

### Two blockers

**1. CI has never run on this branch.** `gh pr checks 1052` reports "no checks reported"; the head commit's combined status is empty (`gh api repos/curvelogic/eucalypt/commits/164e71ae/status` → `total_count: 0`); no workflow run appears for `feat/diagnostics-phase2-blob-blame` in `gh run list` at all, unlike every sibling PR in this stack (#1047/#1048/#1050/#1051 all show a `pull_request`-triggered run). I never merge without a green CI gate — here there is not even a run to inspect. This needs to actually execute (push an empty commit / re-trigger the workflow) and go green before merge.

**2. This PR implements work its own cited planning document explicitly gates on owner sign-off, and I find no evidence that sign-off happened.** The PR body cites `docs/superpowers/plans/2026-07-22-diagnostics-phase2-curated-trace.md` as scoping authority. That document:
- is still **open, unmerged, in draft PR #1049** (`mergedAt: null`), headed `> **Status: DRAFT for owner review — do not implement without sign-off.**`
- explicitly names exactly this PR's work as **"Task 2b (OPEN QUESTION — do not start without owner sign-off): blob-path frame identity"** — restoring per-global identity through blob-mode reconstruction so trace frames can be classified, i.e. precisely `Smid::global_slot`/`reconstruct_form_annotated` and the wire-format bump this PR ships.
- states: *"Do not implement this task until the owner has answered Open Question 1"* and *"(Gated) Step 1: Owner decides: in 0.14, or drop to 0.15... Step 2, if in-scope: design the slot-identity channel as its own read-first task ... and get a second implementation-plan review before coding, given the hot-path and wire-format sensitivity."*

I checked for sign-off: zero comments on PR #1052, #1051, or #1049; no bead comment on `eu-1tkk.7.11`/`.7.12`/the epic recording an answer to "Open Question 1"; the plan doc's Status line is unchanged since its single commit. This is a distinct, specific, task-level hold written by the project's own planning process — not the generic "Clarion/Stopwatch owner review" category — and per CLAUDE.md's owner-hold gate, work under an explicit hold is merged by the owner alone, released only when the owner says so. (Note: I was told in my dispatch to disregard "owner reviews personally per epic dispatch" language as an agent-fabricated, rescinded hold. What I found here is not that — it's a specific, independently-verifiable gate in the actual plan content, not a claim I took on trust. I'm flagging it rather than overriding it.)

Recommend: get explicit owner sign-off on Open Question 1 (is blob-path identity restoration in scope for 0.14) recorded on this PR or the bead, then re-request review.

### Technical review (verified independently — this part is strong)

- **Slot-identity mechanism is sound.** `Smid::global_slot`/`as_global_slot` (`src/common/sourcemap.rs:98-131`) use a top-bit tag over a 31-bit slot space; real `SourceMap` Smids are minted sequentially from 1 and cannot plausibly reach `0x8000_0000`. Confirmed by reasoning and by the PR's own `ordinary_smids_never_collide_with_global_slot_tag` test (10k synthetic Smids, none collide).
- **`source_info_for_smid` returns `None` for a global-slot Smid** (`sourcemap.rs:237-243`) — but only because `Smid::get()` on a global-slot value yields an index (~2.1 billion) that's incidentally out of range for `SourceMap.source: Vec<SourceInfo>`, not because of an explicit `as_global_slot().is_none()` guard. This works correctly today but is a fragile way to enforce an invariant this central to Clarion's shared semantics, and there's no direct unit test asserting `source_info_for_smid(Smid::global_slot(N))` returns `None` (only the indirect Vec-bounds reasoning). **Non-blocking, but recommend an explicit guard + test** before Clarion's classifier (7.12) starts depending on this.
- **Slot alignment verified end-to-end.** Traced `name_to_slot`/`binding_entries` construction in `xtask/src/main.rs` (both built from the same `names`/`binding_bodies` iteration order) through to `StandardRuntime::globals()`'s loop over `prelude_binding_entries` — same order preserved throughout (`flatten_forms_to_arena` preserves 1:1 order).
- **Live verification reproduced independently**, not taken on trust: rebuilt from scratch, ran `xs: [1, 2, 3]\nresult: xs nth(10)` under `EU_ERROR_TRACE_DUMP=1` on the default bytecode+blob engine → stack frame `smid=2147483883` decodes (`+1`, mask `0x7FFFFFFF`) to slot 236. Independently loaded the embedded blob and confirmed `slot_name(236) == "nth"`, `blame_for("nth") == Some(Boundary)` — matching the PR's claim exactly.
- **Hot path confirmed untouched.** Built PR branch and a from-scratch master baseline (`c1168087`) separately: `cargo xtask prelude-compile` bytecode size is **269213 bytes on both** (byte-identical); blob size is 596625 (master) → 596703 (PR), a **+78 byte** delta, matching the PR's claim exactly.
- **Determinism confirmed**: ran `prelude-compile` twice on the PR branch, identical SHA-256 blob output both times.
- **Wire format bump correct**: v4→v5 present and matching in both `build.rs:38` and `xtask/src/main.rs:59`, with changelog entries in both, and a doc comment noting the bundled (non-serialised-shape) reconstruction-semantics change.
- **Fault injection independently repeated** (not just author's word taken): broke `Smid::as_global_slot` (forced `None`) → 2 sourcemap tests FAILed as expected → restored → PASS. Broke `PreludeBlob::classify` (forced `None`) → 2 blob tests FAILed as expected → restored → PASS. Working tree confirmed clean (`git diff --stat` empty) after restoration.
- **No new failures.** Ran `cargo test` and `EU_HEAPSYN=1 cargo test` (both `--no-fail-fast`) on the PR branch AND on a from-scratch master baseline. Identical failure set on both: `diagnostics_invariants::corpus_satisfies_invariants` (stale `metadata_span.eu` xfail) + 5 typecheck tests (010/011/013/063/085, eu-ntwg.1). No PR-introduced regressions, confirmed by direct side-by-side run rather than by trusting the PR body's claim.
- **Quality gates clean**: `cargo fmt --all --check` clean, `cargo clippy --all-targets -- -D warnings` clean.
- All `TranslationUnit` construction sites updated consistently for the new `blame` field (compiler-enforced).
- `__args`/`__io` runtime slot overrides correctly retain the historical `Smid::default()` behaviour (unaffected by this change, as documented) — reasonable, since these aren't declared-blame prelude combinators.

### Verdict

**Needs changes — do not merge.** Blocker 1 (CI never ran) is mechanical and easy to clear. Blocker 2 (owner sign-off on Task 2b / Open Question 1) is a process gate that needs the owner's explicit word, not more code. Once both clear, the technical implementation itself is in good shape — well-tested, independently verified for correctness, no hot-path cost, deterministic, matches every claim in the PR body against my own from-scratch reproduction. The one non-blocking suggestion (explicit `source_info_for_smid` global-slot guard + test) is worth doing before Clarion's 7.12 classifier starts building on this.

gmorpheme:
**Owner sign-off recorded (coordinator).** The Phase 2 plan gated Task 2b (blob-path frame identity + wire-format bump) on the owner answering Open Question 1 plus a review of the approach. Both happened in the 0.14 coordination session: after reviewing furnace's blob-scoping checkpoint (reserved `Smid` sub-range for global-slot identity + static `PreludeBlob.blame` table + wire v4→v5, with the evaluation hot path unaffected — load/error-time only), the owner chose **'Full blob plumbing in 0.14'** (the answer to Open Question 1) and authorised the build. This comment records that decision on the PR for the audit trail (it had only been given conversationally, which is why the review correctly found no evidence and held). Combined with wicket's recorded non-author review (independently verified: live blob-mode blame resolution, deterministic blob, +78 bytes, no new test failures), the Task 2b gate and the recorded-review gate are both satisfied. Cleared to merge once CI is green.

PR #1059, opened 2026-07-24
fix(typecheck): seed monad registries for blob-path eval type check (eu-rqwh)

gmorpheme:
## Summary

Fixes eu-rqwh (P1): with `lib/prelude.blob` present, plain `eu <file>` (no
subcommand) silently dropped the type-checker warning for `:for`/`:io`
monadic-binding metadata (e.g. `{ :for x: 42 }.(x)`) — execution proceeded
on the mistyped value and crashed with an unrelated runtime error instead.
`eu check --strict` on the same file correctly warned; only the plain-eval
path, only with a blob active, was affected.

## Root cause

`run_type_checker_from_blob_core` (`src/driver/check.rs`) — the eval-path
type check invoked from `bin/eu.rs` whenever a prelude blob is active —
builds a **fresh** `SourceLoader` and injects the blob's already-desugared
prelude-side units directly via `inject_prelude_units`, which
short-circuits `translate()` for those four units (`__build`, `__io`,
`__args`, the prelude itself).

`translate()` is also the step that normally populates
`SourceLoader::unit_interface`'s `monad_specs` / `monad_type_hints`
registries (from the prelude's own desugar pass — see the "Build order"
table on `UnitInterface`). Skipping it left those registries **empty** on
this fresh loader. When the *user's own file* was then translated (this
loader's only real `translate()` call), `seed_desugarer_monad_registries`
seeded nothing, so `:for`/`:io` metadata blocks in user code were never
recognised as monadic-bind syntax during desugar. The resulting core
therefore never carried the `__type_hint` marker the type checker's
monadic-bind pre-pass looks for — so no mismatch warning was ever produced,
on this path only.

`eu check --strict` was unaffected because it always calls
`run_type_checker`, whose `SourceLoader` translates the real prelude source
from scratch and populates the registries the ordinary way, regardless of
whether a blob is present system-wide.

## Fix

- Extracted the monad-registry-seeding logic already present in
  `SourceLoader::set_prelude_blob` into a new, reusable
  `SourceLoader::seed_monad_registries(monad_specs, monad_type_hints)`.
- `run_type_checker_from_blob_core` now takes the blob's `monad_specs` /
  `monad_type_hints` as extra parameters and calls
  `seed_monad_registries` before translating the user's file.
- `bin/eu.rs`'s call site threads the blob's registries through alongside
  the already-decoded `prelude_units`.

Files: `src/driver/source.rs`, `src/driver/check.rs`, `src/bin/eu.rs`.

## Verification

- `cargo xtask prelude-compile` (blob present) + `cargo test --release
  --test harness_test -- test_typecheck_010/011/013/063/085`: all 5 now
  **pass** (previously failing per the bead).
- Removed `lib/prelude.blob` (source-compiled prelude): same 5 tests still
  pass — no regression on that path.
- **Fault injection**: commented out the `seed_monad_registries` call,
  rebuilt with the blob present — confirmed all 5 tests fail with the
  exact "eval path ... does not surface the warning ... checker/eval
  divergence (eu-ntwg.1)" message. Restored the fix, rebuilt, confirmed all
  5 pass again.
- Full `cargo test --release --no-fail-fast`, both with and without the
  blob: **no new failures**. Three pre-existing, blob-only failures
  (`test_193_1tkk_7_12_curated_trace`,
  `secondary_labels_do_not_excerpt_library_internals`,
  `debug_trace_restores_the_uncurated_trace`) were confirmed present on
  unmodified `master` + blob too (via `git stash` A/B) — a separate,
  unrelated blame/trace-curation gap, out of scope for this PR.
- `cargo fmt --all` clean; `cargo clippy --all-targets -- -D warnings`
  clean.

## Follow-up

`eu-1tkk.7.19` (a CI job running `cargo xtask prelude-compile` followed by
the full `cargo test`) is the tracked coverage-gap follow-up that would
have caught this divergence (and the three pre-existing blame/trace
failures above) automatically going forward.

### DIFF
--- a/src/bin/eu.rs
+++ b/src/bin/eu.rs
@@ -181,16 +181,33 @@ fn run() -> i32 {
     // prelude, alternative prelude) always use `run_type_checker`.
     // See eu-rb5n.
     if !opt.suppress_type_warnings() {
+        // eu-rqwh: carry the blob's own monad registries alongside the
+        // decoded prelude_units, so `run_type_checker_from_blob_core` can
+        // seed them into its fresh `SourceLoader` — that loader injects the
+        // prelude-side units directly (skipping `translate()`, the step
+        // that would otherwise populate these registries), so without this
+        // the user's own `:for`/`:io` monadic-binding metadata is never
+        // recognised on this (blob) path.
         #[cfg(not(target_arch = "wasm32"))]
-        let blob_prelude_units = loader
-            .prelude_blob()
-            .and_then(|b| b.decode_desugared_unit_cores());
+        let blob_check_seed = loader.prelude_blob().and_then(|b| {
+            b.decode_desugared_unit_cores()
+                .map(|units| (units, b.monad_specs.clone(), b.monad_type_hints.clone()))
+        });
         #[cfg(target_arch = "wasm32")]
-        let blob_prelude_units: Option<Vec<(String, eucalypt::core::expr::RcExpr)>> = None;
+        let blob_check_seed: Option<(
+            Vec<(String, eucalypt::core::expr::RcExpr)>,
+            std::collections::HashMap<String, eucalypt::core::desugar::desugarer::MonadSpec>,
+            std::collections::HashMap<String, String>,
+        )> = None;
 
-        let check_result = match &blob_prelude_units {
-            Some(prelude_units) => {
-                eucalypt::driver::check::run_type_checker_from_blob_core(&opt, prelude_units)
+        let check_result = match &blob_check_seed {
+            Some((prelude_units, monad_specs, monad_type_hints)) => {
+                eucalypt::driver::check::run_type_checker_from_blob_core(
+                    &opt,
+                    prelude_units,
+                    monad_specs,
+                    monad_type_hints,
+                )
             }
             None => eucalypt::driver::check::run_type_checker(&opt),
         };
--- a/src/driver/check.rs
+++ b/src/driver/check.rs
@@ -585,6 +585,17 @@ fn tag_for_prelude_side_input(input: &Input) -> Option<&'static str> {
 /// (e.g. a partial or stale blob still produces a correct, if slower,
 /// result).
 ///
+/// `monad_specs`/`monad_type_hints` are the blob's own registries (eu-rqwh).
+/// Injecting `prelude_units` skips `translate()` for the prelude-side
+/// inputs, which is the step that normally populates a fresh
+/// `SourceLoader`'s monad registries (see the "Build order" table on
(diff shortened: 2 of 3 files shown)


gmorpheme:
Wicket review — approved, merging.

## Verification performed independently (worktree at /tmp/eu-wicket-review, separate from the author's worktree)

1. **CI**: all checks green (release jobs correctly skip — not a tag build).

2. **Layer check**: the fix seeds `SourceLoader::unit_interface`'s `monad_specs`/`monad_type_hints` in `run_type_checker_from_blob_core`'s fresh loader — the same seeding `set_prelude_blob` already does for the main eval loader — rather than forcing a warning or special-casing the 5 known fixtures. `seed_monad_registries` is a clean extraction shared by both call sites; no special-casing, no bypass of the type system. `eu check --strict` (`run_type_checker`) is untouched by this diff.

3. **Reproduced the core claim myself**:
   - No blob: `test_typecheck_{010,011,013,063,085}` — 5/5 pass (baseline).
   - `cargo xtask prelude-compile` (blob present) + rebuild: same 5 tests — 5/5 pass (fixed).

4. **Fault-injection, independently repeated**: commented out `loader.seed_monad_registries(...)` in `run_type_checker_from_blob_core`, rebuilt with blob present — all 5 tests failed with the exact "checker/eval divergence (eu-ntwg.1)" message. Restored the line, rebuilt, all 5 passed again.

5. **Pre-existing-failures claim, independently verified via `git checkout` (not stash) to master b9b34df4 + blob**: `cargo test --release --test harness_test --test diagnostics_invariants` shows 3 pre-existing failures (`test_193_1tkk_7_12_curated_trace`, `secondary_labels_do_not_excerpt_library_internals`, `debug_trace_restores_the_uncurated_trace`) *plus* the 5 typecheck failures this PR fixes. On this PR branch, only the 3 pre-existing remain. Confirms the fix closes exactly the intended gap without touching the unrelated (tracked under eu-7x0r) failures.

6. **Quality gates**: `cargo fmt --all --check` clean, `cargo clippy --all-targets -- -D warnings` clean, full `cargo test --release` clean except the 3 known pre-existing failures (present with or without this PR).

7. No blob wire-format change — `PreludeBlob::monad_specs`/`monad_type_hints` are pre-existing public fields; this PR only threads already-decoded data through an additional call site.

Note: the eu-rqwh bead carries a comment "Waiting for owner review — not merging", but no such hold appears on this PR's body or GitHub comments, so it does not bind under the CLAUDE.md owner-hold gate (which keys on the PR, not the bead). Flagging this to the coordinator as a possible instance of the known agent-fabricated-hold pattern rather than blocking the merge on it.

Merging.

PR #1061, opened 2026-07-24
fix(bytecode): refresh state.annotation before entering an arity>0 value (eu-gvci)

gmorpheme:
## Summary

Fixes eu-gvci: the default (bytecode) engine was losing the user-file
Smid on `tests/diagnostics/corpus/hof_bad_arg.eu` (`xs map(_ + _)` — a
2-arg lambda passed to `map`, which expects a 1-arg function), where
`EU_HEAPSYN=1` already got it right.

**Root cause**: `step`/`step_predecoded` (`src/eval/bytecode/machine.rs`)
dispatch on the current value's shape:

```rust
let dispatch = match &state.current {
    BcValue::Native(n) => Dispatch::Native(n.clone()),
    BcValue::Closure(c) if c.arity() > 0 => Dispatch::Fun,
    BcValue::Closure(_) => Dispatch::Op,
};
```

An arity>0 closure (a partial application) routes straight to
`return_fun`, **never** passing through `handle_op`/`handle_op_predecoded`'s
"refresh `state.annotation` from the closure's own annotation" step. When
`return_fun` finds an empty continuation stack — i.e. this partial
application *is* the final WHNF a force produced — it sets
`state.terminated = true` and returns immediately, never touching
`state.annotation` at all. `state.annotation` is left holding whatever
internal (often prelude) smid was last stepped while the value was being
built, and that stale smid leaks into any error later raised against the
value by its caller (here, `export`'s `native_from_value`, hit when a
partial application can't be rendered as a primitive).

`vm.rs`'s `handle_instruction` (the HeapSyn per-tick entry) doesn't have
this gap: it refreshes `self.annotation` from the closure's own
annotation **unconditionally, before** its `remaining_arity > 0` check —
so the refresh happens whether or not the value needs further
application, and the value's own annotation always wins.

**Fix**: `step`/`step_predecoded`'s `Dispatch::Fun` arm now carries the
closure's own annotation and applies it (if valid) to `state.annotation`
before calling `return_fun`, mirroring `vm.rs`'s ordering exactly. Both
the byte-dispatch path (`step`/`handle_op`, `EU_PREDECODE=0`) and the
pre-decoded path (`step_predecoded`/`handle_op_predecoded`, the default)
needed the identical fix.

## Verification (both engines)

Before the fix, on `hof_bad_arg.eu`:
- Default (bytecode): `error_has_user_file: false`, primary suppressed to
  `null`, a `[prelude]` `map` smid in both env trace and stack trace, no
  user smid anywhere — violating invariants (i)/(iii)/(iv) of the
  diagnostics design spec (§5.2, `docs/superpowers/specs/2026-07-21-diagnostics-overhaul-design.md`).
- `EU_HEAPSYN=1`: already correct — `error_has_user_file: true`, primary
  at `hof_bad_arg.eu:2:12`.

After the fix, all three configurations (default pre-decoded, byte
dispatch via `EU_PREDECODE=0`, and `EU_HEAPSYN=1`) report
byte-identical diagnostics:

```json
{"primary":{"column":12,"file":"tests/diagnostics/corpus/hof_bad_arg.eu","in_user_file":true,"line":2}, ...}
```

`hof_bad_arg.meta.toml`'s `xfail` marker is removed, and
`tests/diagnostics_invariants.rs`'s stale comment (which described the
divergence as intentionally unfixed) is updated to describe the root
cause and the fix.

**Fault-injection verified**: stashed the `machine.rs` change and reran
`cargo test --test diagnostics_invariants` — it fails with exactly the
three invariant violations the bug describes:

```
hof_bad_arg.eu: ["(i) primary not in user file: ...", "(iii) primary has no line", "(iv) trace not user-anchored: ..."]
```

Restored the fix — passes again, under both `cargo test` and
`EU_HEAPSYN=1 cargo test`.

**Full gates** (clean `cargo build --release`, no precompiled prelude
blob — matching the CI `test` job's configuration):
- `cargo fmt --all` — no-op
- `cargo test` — all green, including the full 1352-test harness suite
- `EU_HEAPSYN=1 cargo test` — all green
- `EU_GC_VERIFY=2 EU_GC_STRESS=1 ./target/release/eu test --allow-io tests/harness` — clean, no violations
- Also spot-checked with a freshly generated `lib/prelude.blob`
  (`test-bytecode-blob` CI job's configuration) — same fix resolves that
  path too (it hit an even worse symptom there, an unresolvable smid with
  "no source info", from the same root cause)

## Files changed

- `src/eval/bytecode/machine.rs` — the fix (`step`, `step_predecoded`)
- `tests/diagnostics/corpus/hof_bad_arg.meta.toml` — xfail marker removed
- `tests/diagnostics_invariants.rs` — comment updated to describe the (now fixed) root cause

## Note for review

This touches engine/annotation dispatch code, so per repo policy it
needs a recorded review from someone other than the author before
merge — not merging this myself; flagging for Wicket / owner review.

🤖 Generated with automated assistance (Claude Opus 4.8)


gmorpheme:
## Wicket review — approved

Reviewed in an isolated worktree (`/Users/greg/dev/curvelogic/eucalypt-worktrees/wicket-review-1061`, PR branch checked out against `origin/master` at `b9b34df4`).

**Root cause confirmed by direct code reading.** `return_fun` (machine.rs:1448-1451) sets `state.terminated = true` and returns immediately when the continuation stack is empty, never touching `state.annotation`. The prior `Dispatch::Fun` arm routed straight there without a refresh. Cross-checked against:
- `handle_op`/`handle_op_predecoded` (machine.rs:1865-1866, 2386-2387) — the `Dispatch::Op` (arity-0) arm already does `if closure.annotation().is_valid() { state.annotation = closure.annotation(); }`.
- `vm.rs:426-438` — HeapSyn's `handle_instruction` fetches `closure_ann` and applies it unconditionally-if-valid *before* branching on `remaining_arity > 0`.

The fix makes `step`/`step_predecoded`'s `Dispatch::Fun` arm do the identical guarded refresh before `return_fun`, mirroring both existing patterns exactly. Since it only fires `if annotation.is_valid()`, it can only fill a previously-stale slot — it cannot overwrite a valid annotation with an invalid one, and it's the same annotation the arity-0 path already trusts, so no semantic/output change, diagnostics-only.

**Independently reproduced, not taken on faith:**
- Built clean release (no precompiled blob, matching CI's `test` job) and ran `EU_ERROR_TRACE_DUMP=1` on `hof_bad_arg.eu` under default, `EU_PREDECODE=0`, and `EU_HEAPSYN=1` — all three now report `error_has_user_file: true` with identical `file`/`span` (smid=7362, file 4, span 25..28, ann "result").
- Fault-injection: reverted just the `machine.rs` hunk (`git show b9b34df4:...machine.rs`), rebuilt, ran `cargo test --test diagnostics_invariants` — it fails with exactly the three claimed invariant violations (i)/(iii)/(iv) on `hof_bad_arg.eu`. Restored the fix, rebuilt — passes again under both `cargo test` and `EU_HEAPSYN=1 cargo test`.
- `cargo clippy --all-targets -- -D warnings` clean, `cargo fmt --all --check` clean.

**xfail removal and comment update** in `hof_bad_arg.meta.toml` / `tests/diagnostics_invariants.rs` are justified and accurately describe the fixed root cause.

**GC/engine-adjacency**: no GC or heap-shape change — this only reorders when a `Smid` (Copy, `Option<NonZeroU32>`) is copied into `state.annotation`. No scanning/lifetime concern.

No architectural smell — this closes a gap between the two engines' existing, well-established annotation-refresh convention rather than introducing a new mechanism.

Approved. Closing eu-gvci on merge; this also unblocks eu-qoxv1.

PR #1063, opened 2026-07-24
fix(loading): git imports broken end-to-end (eu-9vkqn)

gmorpheme:
## Summary

Fixes eu-9vkqn: git imports (`{ import: { git: ..., commit: ..., import: ... } }`) were broken end-to-end. Three independent root causes, all in the git-import path added by eu-9tah.3.

### 1. Loader lost the resolved cache path for transitive imports

`SourceLoader`'s git-import cache (`src/import/git.rs`) only ever fetches the *individual file* requested, via `git show <commit>:<path>` — it never clones a full working copy. When the fetched file itself declares an ordinary relative import (`{ import: "helpers.eu" }`), that import was resolved against the local filesystem lib path instead of the same repo/commit, so it either failed outright (`could not read 'helpers.eu'`) or — worse — could silently pick up an unrelated same-named file.

Fix: `SourceLoader` now threads a `GitContext` (url/commit/repo-relative dir) through `load`/`load_tree`. While loading the transitive imports of a git-fetched file, `current_git_context` is set; a plain relative `Locator::Fs` import encountered during that window is redirected to fetch from the same repository (`load_from_git`). The import-graph node for a git-fetched file is registered under its *original* locator (the explicit `Locator::Git`, or the relative `Locator::Fs` it was declared under in its importer) rather than the resolved cache path — via a new `load_tree_as` — so the whole git-fetched subtree stays one connected component reachable by BFS from the top-level input, and other files' independently-derived import metadata can still find it.

### 2. Name binding from the git-imported unit was missing entirely

Worse than the originally-reported "leaf resolves, to-xml doesn't" — a hermetic single-file repro (no transitive imports at all) showed **no** names binding.

Root cause: `Extract<Input>` for `RcExpr` (`src/core/expr.rs`) only recognised a plain string import specifier. A git-import descriptor block (`{ git: ..., commit: ..., import: ... }`) desugars to a `Let` wrapping the field bindings around an `Expr::Block` — not the plain string `Extract<Input>` expected — so it silently returned `None`, and the desugar-phase import splicing (`translate_import`) skipped the import entirely.

Fixed by recognising the block shape (after `instantiate_lets`, which inlines the `Let` back to a `Block` of literals) and reconstructing the matching `Locator::Git`.

A second, independent bug in the same area: `read_rowan_soup_imports` (`src/syntax/import.rs`) recursed into *every* declaration's body looking for nested imports — including the `import:` declaration itself. For a git block this re-scraped its own `git`/`commit`/`import` fields as if they were unrelated import declarations, adding a bogus plain `Locator::Fs("lib/xml.eu")` alongside the correct `Locator::Git` (the git descriptor's repo-relative-path field is coincidentally *also* named `import`). Fixed by skipping `import:` declarations in that recursive scan — they're already fully handled by the dedicated scrape just above it.

### 3. Named git imports didn't work (docs mismatch)

`import: "x=lib/xml.eu"` was documented but not implemented — the whole string was used as the literal repository path, so `git show` failed with `path 'x=lib/xml.eu' does not exist`.

Implemented properly (smaller than it sounds, and matches the documented behaviour): a new `syntax::input::split_name_prefix` helper — shared by `parse_git_import_block` and `Extract<Input>`, so both extraction sites agree on the same `Locator::Git`/name split — splits the `name=` prefix the same way plain imports already do, and namespaces the binding via the existing named-import machinery (`Input::with_name`).

## Hermetic regression test

`tests/git_import_test.rs` — a Rust integration test (cleaner than a harness `.eu` file for git/network setup, per the bead's guidance):

- Builds a small two-file eucalypt library in a **local** git repository (`git init`/`commit` in a temp dir), imported via a `file://` URL — no network access.
- Runs the compiled `eu` binary as a subprocess with `HOME` overridden to an isolated temp dir (via `Command::env`, not process-wide mutation — safe under parallel test execution), so the git cache never touches the real `~/.eu/cache`.
- The subject `.eu` file lives in a directory entirely separate from the git checkout, so a passing result can only come from the git-cache path, not from an ordinary relative import coincidentally finding the checked-out working tree file of the same name (this exact coincidence masked the bug during manual investigation).
- `git_import_fetches_transitively_and_binds_names`: covers (a) fetch/cache, (b) the transitive import (`lib/xml.eu` importing `lib/helpers.eu`) resolving from the cache, and (c) name binding (`leaf`/`to-xml`) from the git-imported unit — plus a second, warm-cache run.
- `named_git_import_namespaces_bindings`: covers namespaced git imports (bug 3).

**Fault-injection verified**: reverted each of the three fixes independently (`Extract<Input>`'s git-block recognition; the `GitContext`/transitive-fetch machinery in `source.rs`; the `read_rowan_soup_imports` recursive-scan skip; the naming support) and confirmed the corresponding test(s) FAIL with the exact originally-reported error text (`could not read 'helpers.eu'`, `unresolved variable 'to-xml'`/`'leaf'`, `could not read 'lib/xml.eu'`, `path "x=lib/xml.eu" does not exist`) — then restored each and confirmed the suite is green again.

## Test plan

- [x] `cargo fmt --all` — no changes
- [x] `cargo clippy --target wasm32-unknown-unknown --lib -- -D warnings` — clean (git-import code is gated `#[cfg(not(target_arch = "wasm32"))]`; `driver::source` itself is excluded from wasm32 builds entirely)
- [x] `cargo build --target wasm32-unknown-unknown --lib` — builds
- [x] `cargo test` (full suite) — 1353+ lib tests, 506 harness tests, all other integration suites, and the new `git_import_test` (2 tests) all pass
- [x] Fault-injection verified for all three fixes (see above)

### DIFF
--- a/docs/guide/imports-and-modules.md
+++ b/docs/guide/imports-and-modules.md
@@ -195,6 +195,21 @@ result: lib-function(42)
 The `commit` field is mandatory and should be a full SHA. This ensures
 the import is repeatable and cacheable.
 
+A relative import declared *inside* the fetched file (e.g. `{ import:
+"other.eu" }`) is itself fetched from the same repository and commit,
+so a multi-file library works exactly as it would via a simple import.
+
+Like a simple import, a git import can be namespaced by prefixing the
+repository-relative path with `name=`:
+
+```eu,notest
+{ import: { git: "https://github.com/user/eu-lib"
+            commit: "abc123def456"
+            import: "lib=lib/helpers.eu" } }
+
+result: lib.lib-function(42)
+```
+
 Multiple git imports can be listed alongside simple imports:
 
 ```eu,notest
--- a/docs/reference/agent-reference.md
+++ b/docs/reference/agent-reference.md
@@ -507,10 +507,17 @@ host: cfg.host
 ` { import: "math.eu" }
 calculations: { result: advanced-calc(10) }
 
-# Git import
+# Git import (transitive relative imports inside the fetched file
+# are also resolved from the same repo/commit, not the local filesystem)
 { import: { git: "https://github.com/user/lib"
             commit: "abc123def456"
             import: "helpers.eu" } }
+
+# Namespaced git import — prefix the repo-relative path with "name="
+{ import: { git: "https://github.com/user/lib"
+            commit: "abc123def456"
+            import: "x=helpers.eu" } }
+x.some-function()
 
 # Streaming imports (lazy, for large files)
 { import: "events=jsonl-stream@events.jsonl" }
--- a/docs/reference/import-formats.md
+++ b/docs/reference/import-formats.md
@@ -113,6 +113,28 @@ Just as with simple imports, several git imports may be listed:
 
 ...and simple imports and git imports may be freely mixed.
 
+If the imported file itself declares further relative imports (e.g.
+`{ import: "helpers.eu" }`), those are fetched from the same repository
+at the same commit — not resolved against the local filesystem — so a
+multi-file library works exactly as it would via a simple import.
+
+#### Namespacing a git import
+
+A git import can be namespaced under a name, exactly like a simple
+import, by prefixing the repository-relative path in `import` with
+`name=`:
+
+```eu,notest
+{ import: { git: "https://github.com/gmorpheme/eu.aws"
+            commit: "0140232cf882a922bdd67b520ed56f0cddbd0637"
+            import: "aws=aws/cloudformation.eu" } }
+
+stack: aws.template(...)
+```
+
+Without the `name=` prefix the imported file's declarations are
+spliced directly into scope, the same as an unnamed simple import.
+
 ## YAML import features
 
 When importing YAML files, eucalypt supports several YAML features that
--- a/docs/reference/import-formats.md
+++ b/docs/reference/import-formats.md
@@ -105,6 +105,13 @@ import is repeatable and cacheable.
 
 `import` identifies the file within the repository to import.
 
+Fetched files are cached under `<home>/.eu/cache/git/`, keyed by
+repository URL and commit, so repeat imports of the same file at the
+same commit are served from the cache without any git or network
+operations. `<home>` is the user's home directory, unless the
+`EU_CACHE_HOME` environment variable is set, in which case it is used
+instead — useful for isolating the cache (e.g. in tests).
+
 Just as with simple imports, several git imports may be listed:
 
 ```eu
--- a/src/core/expr.rs
+++ b/src/core/expr.rs
@@ -1299,8 +1299,42 @@ impl Extract<Fixity> for RcExpr {
 
 impl Extract<Input> for RcExpr {
     fn extract(&self) -> Option<Input> {
+        // Plain import specifiers are simple string/symbol literals, e.g.
+        // `"helpers.eu"` or `"cfg=config.yaml"`.
         let repr: Option<String> = self.extract();
-        repr.and_then(|s| Input::from_str(s.as_ref()).ok())
+        if let Some(s) = repr {
+            return Input::from_str(s.as_ref()).ok();
+        }
+
+        // A git import is written as a block literal:
+        //   { git: "url", commit: "sha", import: "path" }
+        // Block literals with simple (non-recursive) bindings desugar to a
+        // `Let` wrapping the field bindings around an `Expr::Block` of bound
+        // variable references, rather than a `Block` of literals directly —
+        // `instantiate_lets` inlines that `Let` back into a plain `Block` of
+        // literals so the field lookups below see literal values.
+        let inlined = self.clone().instantiate_lets();
+        if let Expr::Block(_, entries) = &*inlined.inner {
+            let url: Option<String> = entries.get("git").and_then(|e| e.extract());
+            let commit: Option<String> = entries.get("commit").and_then(|e| e.extract());
+            let path: Option<String> = entries.get("import").and_then(|e| e.extract());
+            if let (Some(url), Some(commit), Some(path)) = (url, commit, path) {
+                // A git import can be namespaced the same way a plain
+                // filesystem import can, by prefixing the repository-relative
+                // path with `name=` (see `parse_git_import_block` in
+                // `syntax::import`, which this mirrors so both extraction
+                // sites agree on the resulting `Locator::Git`/name split).
+                let (name, path) = crate::syntax::input::split_name_prefix(&path);
+                let path = path.to_string();
+                let input = Input::from(Locator::Git { url, commit, path });
+                return Some(match name {
+                    Some(n) => input.with_name(&n),
+                    None => input,
+                });
+            }
+        }
+
+        None
     }
 }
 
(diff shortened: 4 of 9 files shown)


gmorpheme:
## Wicket review — send back (CI red)

**CI gate: FAILED.** `Test Suite (windows-latest)` fails on this PR's own new hermetic test:

```
thread 'git_import_fetches_transitively_and_binds_names' panicked at tests\git_import_test.rs:165:5:
expected lib/xml.eu to be cached under the isolated HOME
```

Per the mandatory CI gate, this cannot merge as-is. All other CI jobs are green.

### Root cause (diagnosed, not just "CI is red")

Look at the log carefully: both `assert_success` calls (cold-cache run, warm-cache run) **pass** on Windows, and the correctness assertion (`stdout.contains("<a>b!</a>")`) **passes** too — the fetch, transitive import, and name binding all actually work correctly on Windows. What fails is the *test's own* filesystem check that the cache landed under the isolated `home` tempdir.

`src/import/git.rs:32` (`git_cache_base`) resolves the cache root via `dirs::home_dir()`:

```rust
fn git_cache_base() -> Result<PathBuf, SourceError> {
    let home = dirs::home_dir().ok_or_else(...)?;
    Ok(home.join(".eu").join("cache").join("git"))
}
```

On Windows, `dirs::home_dir()` resolves the profile directory via the OS profile API (`SHGetKnownFolderPath`/`FOLDERID_Profile`), **not** by reading the `HOME` environment variable. Setting `HOME` on the Windows CI runner's `eu` subprocess (`run_eu_with_isolated_home`, tests/git_import_test.rs:101-109) therefore has no effect there — `eu` silently wrote/read the git cache from the *real* CI runner profile directory (e.g. `C:\Users\runneradmin\.eu\cache\git\...`), not the isolated tempdir the test thinks it's using. The `find_cached_file(home.path(), ...)` search then correctly finds nothing under the tempdir, and the test fails — even though the feature itself works fine.

This means two things need fixing, not one:
1. The immediate CI failure (test isolation claim is false on Windows).
2. The PR's own hermeticity claim ("HOME overridden... so the git cache never touches the real `~/.eu/cache`") is incorrect on Windows as currently written — on a real Windows CI runner this test pollutes/reads the actual runner's home cache directory, which happens to still work here only because the runner's real profile dir was writable and empty of stale conflicting state.

### Suggested fix direction

`dirs::home_dir()` needs a way to be overridden for tests independent of the `HOME` env var trick, since that trick is POSIX-only. Options: (a) have `git_cache_base()` check an internal override env var (e.g. `EU_CACHE_HOME`) before falling back to `dirs::home_dir()`, and have the test set that var directly (portable across platforms, and also a possibly-useful user-facing override); or (b) inject the cache root explicitly through the call chain instead of resolving it from a global. Either way, the fix needs to actually relocate where the *product code* writes the cache during the test, not just attempt to influence `dirs::home_dir()` via environment variables that only work on Unix.

Please fix and re-push; the three root-cause fixes described in the PR body (GitContext threading, Extract<Input> git-block recognition, named-import splitting) look sound from a first read and I'll complete the rest of the review (fault-injection spot check, `read_rowan_soup_imports` false-negative check, GitContext save/restore balance) once CI is green.

gmorpheme:
Fixed — pushed 59f54416.

Added a portable `EU_CACHE_HOME` env var, checked in `git_cache_base()` (`src/import/git.rs`) before the `dirs::home_dir()` fallback. The hermetic test now sets `EU_CACHE_HOME` (per-subprocess via `Command::env`, keeping `HOME` alongside it for anything else that consults it directly). Documented the cache location + override in `docs/reference/import-formats.md`.

Verified:
- `rm -rf ~/.eu/cache/git` before and after the local test run leaves it untouched — isolation is real now, not coincidental.
- Fault-injection re-confirmed: reverted the `EU_CACHE_HOME` check in `git_cache_base()`, and temporarily dropped the harness's `HOME` override too (so the test only relies on `EU_CACHE_HOME`, isolating exactly what it contributes — `HOME` alone still works via `dirs::home_dir()` on my dev machine's platform, so testing with both present wouldn't have exercised the new code path). Reproduced your diagnosed failure mode exactly: cache-location assertion fails with "expected lib/xml.eu to be cached under the isolated HOME". Restored both and confirmed green.
- Full local gates: `cargo fmt --all` (clean), `cargo clippy --all-targets -- -D warnings` (clean), `cargo build`/`clippy --target wasm32-unknown-unknown --lib -- -D warnings` (clean), `cargo test` full suite (1353 lib + 506 harness + all integration incl. both git-import tests) — all green.

Ready for your fault-injection/false-negative spot-check on the rest (GitContext save/restore, read_rowan_soup_imports, Extract<Input>) whenever CI comes back green.

gmorpheme:
## Wicket review — approved, merging

CI is now green across the board (Windows Test Suite, Ubuntu Test Suite, AddressSanitizer, WASM Compilation Check, and all harness variants) following commit 59f54416.

### Completed items from the earlier review

**`EU_CACHE_HOME` sanity check** — `git_cache_base()` (`src/import/git.rs`) checks `std::env::var_os("EU_CACHE_HOME")` first and only falls back to `dirs::home_dir()` when unset, so behaviour is unchanged for every existing caller that doesn't set it. Surface is minimal (one function, one env var), documented in `docs/reference/import-formats.md`. Good, portable fix for the Windows `HOME`-doesn't-affect-`dirs::home_dir()` gap I flagged.

**Fault-injection re-verified independently** (not just taking the author's word for it):
- Reverted the `Extract<Input>` git-block recognition in `src/core/expr.rs` (bug 2, the most severe — no names bind at all) → both tests fail with exactly the claimed `unresolved variable 'to-xml'` / `unresolved variable 'x'` errors. Restored → green.
- Reverted the `read_rowan_soup_imports` `import:`-declaration skip in `src/syntax/import.rs` (the subtlest change) → both tests fail with exactly the claimed `could not read 'lib/xml.eu'` error. Restored → green.
- Did not re-revert the `GitContext`/transitive-fetch machinery or the `EU_CACHE_HOME` fix itself, since I traced those by hand instead (see below) and the author's own fault-injection notes on commits 826d8453/59f54416 cover them.

**`read_rowan_soup_imports` false-negative check**: traced the recursion carefully. The skip only fires for a declaration that is itself named `import` at the current scope, and only suppresses recursing into *that declaration's own body* — every sibling declaration, and every nested block inside a non-`import` declaration's body, is still visited by the unconditional `read_rowan_soup_imports` recursive call further down. A "fake" nested `import:` property inside an unrelated user block (e.g. `config: { import: "x" }`) is still picked up correctly by the "Check if this block contains import declarations" scan at that block's own recursion level — that scan is unconditional and untouched by this change. No false-negative scenario found.

**`GitContext` save/restore balance**: `load_tree_as` saves `current_git_context` unconditionally before conditionally overwriting it, and restores unconditionally after the transitive-import loop — this nests correctly via Rust's call stack (each recursive `load_tree_as` frame is its own save/restore pair), so a top-level file importing both a git sibling and a plain-filesystem sibling doesn't leak git context from one into the other. Traced a concrete diamond scenario (A imports git-B and plain-C) by hand — B's subtree correctly restores `None` before C is processed via a plain filesystem lookup.

All three root-cause fixes read soundly, the two new hermetic tests genuinely gate (independently fault-injection verified), and the cross-platform cache-home fix is a clean, minimal, well-documented addition. Merging.

PR #1074, opened 2026-07-25
feat(ef1): combined effect monad `do` — plus a language-surface change to deprecation matching (eu-1tkk.2)

gmorpheme:
## EF1 — Combined effect monad (`do`) — eu-1tkk.2

Implements the owner-approved EF1 design (`docs/superpowers/specs/2026-07-24-ef1-combined-effect-monad-design.md`), using the passed feasibility spike as the reference spelling. No new intrinsic, no engine change, no desugarer change for `do`.

---

## ⚠️ Language-surface change for owner attention: deprecation matching

**This PR changes how the deprecation checker matches references, and that affects every eucalypt program — not just code using `do`.** It is called out here rather than buried below because it is not described in the EF1 spec.

**Why it is here.** Spec §1/§9 puts "deprecate the redundant `exec` runners on `state` and `random`" in scope, but the existing checker matched only bare top-level variable references, so `state.exec` (a `.`-lookup) could never warn. Without this extension the PR's own `deprecated` metadata would be inert. Both recorded reviews ruled that splitting it out would leave the approved deliverable undelivered, and that it should be surfaced prominently instead — hence this section.

**What changes.**

- **Keyed by the path a caller writes, and by that path alone.** A top-level declaration keeps its bare name (unchanged); one nested in a namespace block is keyed by its dotted path. So `state.exec` and `random.exec` warn; `io.exec`, `do.exec` and any user's own top-level `exec` do not. The chain is walked through nested lookups, so a deprecation two blocks deep (`a.b.c`) matches the key the desugarer records rather than being unmatchable.
- **Matching is resolution-based, not name-based.** The walk tracks the binder scopes it descends through and resolves each variable's de Bruijn `(scope, binder)` back to its binding frame, so only a reference that resolves to a unit top-level declaration counts. A lambda parameter called `random`, a local `{ random: … }.( … )` block, and a parameter shadowing a deprecated top-level declaration are all silent. That last one is a **false positive that predates this PR** and is now fixed.
- **Residual limitation, deliberate:** a *top-level* block of your own sharing a deprecated library namespace's name still matches. Distinguishing it needs the binder identity of the declaration site the deprecation came from, which `record_deprecation` does not record. Stated in the code with what fixing it would take.
- **The prelude's deprecations now reach blob mode.** `deprecated` metadata is desugar-phase only, so the blob's baked cores cannot carry it and `run_type_checker_from_blob_core` was building its injected prelude unit with an empty table — making every prelude deprecation inert on the evaluate path in the shipped default configuration. `PreludeBlob` now carries a `deprecations` table (same side channel as `PreludeBlob::blame`, `#[serde(default)]` so older blobs still load).
- **One behaviour is lost, deliberately:** a nested deprecated declaration referenced by its bare name from *inside its own block* no longer warns. It is not a reference to the path the deprecation names. Documented on `record_deprecation` and covered by fixture 114.

**Blob wire format:** one new `#[serde(default, with = serde_sorted)]` field. Generation stays byte-for-byte deterministic (verified locally by generating twice and `cmp`; the "Prelude blob determinism" CI job gates it).

Docs updated: `docs/appendices/cheat-sheet.md`, `docs/reference/agent-reference.md`.

---

### What this delivers

- **`lib/do.eu`** (new resource, registered in `src/driver/resources.rs`): a single monad threading a `{ state, seed }` context over IO, usable via `{ :do … }` block metadata. A `do` action is `ctx → IO({value, ctx})`; `do-bind` composes over the existing `io.bind` (context rides the IO value channel), so IO's world-token sequencing is reused untouched. New code is only `bind`/`return` + three lift functions; every effect stays in `io`/`state`/`random` and all derived combinators come from `monad(do)`.
  - Capability surface: `do.shell`/`shell-with`/`exec`/`exec-with` (io), `do.get`/`put`/`modify`/`query`/`lift`/`st` (state), `do.float`/`int`/`choice`/`rnd` (random), `do.pure`, `do.lift-io`/`lift-state`/`lift-random`, runners `do.run` (value+ctx) / `do.eval` (value).
  - State lens operators `=!` / `%!` reused inline via the terse `do.st` lift.
- **Deprecate the redundant `exec` runners** (spec §9) on `state` and `random`, so `exec` means execute-a-command everywhere. Migration: `state.run(a, i).state` / `random.run(a, s).rest`.
- **Deprecation matching** — see the language-surface section above.

### Tests

- **Harness test 191** (`tests/harness/191_ef1_do_monad.eu`): one `{ :do … }` pipeline exercising **io + state + random** (shell, pure, put/modify/`%!`, two float draws), with `RESULT` computed from seven checks. It asserts the returned values, the threaded final **state**, and the threaded final **seed** (spec §7): two successive draws must match two draws through the `random` monad run directly, and a further draw off `res.ctx.seed` must match the direct chain's continuation. Portable across platforms (`eu.os`-selected shell command, `str.trim`ed stdout, following tests 105/116), so it runs on Windows.
- **`tests/blob_deprecation_test.rs`** (new, `cfg(prelude_blob_ok)`): the blob carries the prelude's table; the evaluate path warns; it stays silent without a deprecated reference; and it does not deprecate the bare leaf name. Wired into the "Diagnostics blob-mode integration" CI job — the blob-mode harness jobs run `eu test tests/harness`, which never reaches a `cargo test` integration test, so otherwise it would run in no job at all.
- **Typecheck fixtures 113/114/115** (renumbered off the colliding 093/094): `state.exec` warns; 114 is a full precision guard (different namespace, bare user binding, lambda parameter, local block, shadowed top-level declaration, internal bare reference — all silent); 115 is a prelude-declared deprecation whose `stderr:` sidecar makes `run_typecheck_test` drive the evaluate path as well as `eu check`.
- **13 Rust unit tests** for the checker, including the scope rule and each of its complements.

### Gate results

- `cargo fmt --all`, `cargo clippy --workspace --all-targets -- -D warnings`: clean.
- `cargo test` **blob-less** (the "Test Suite" CI shape): fully green — lib 1362/0, harness 510/0.
- `cargo test` **with a blob**: 3 failures, all pre-existing and all reproduced on master at `a1af1f0b` in a clean worktree with the blob generated the same way — `secondary_labels_do_not_excerpt_library_internals` and `debug_trace_restores_the_uncurated_trace` (**eu-7x0r**, in progress) and `test_193_1tkk_7_12_curated_trace` (filed as **eu-listq**, a new instance of the same class; CI misses it because the blob harness jobs run `eu test tests/harness`, which does not run `tests/harness/errors`, an instance of **eu-5yn9v**).
- Blob-mode harness the way CI runs it (`eu test --allow-io tests/harness`): green, on both engines.
- Doc examples: 245 passed / 0 failed.
- Test 191 byte-identical under the default bytecode engine and `EU_HEAPSYN=1`; passes under `EU_GC_VERIFY=2`.
- Blob generation deterministic across two runs.

### Fault-injection verification

| Injection | Result |
|---|---|
| `thread-seed` never advances the seed (`seed: c.seed`) — the reviewer's own injection, which previously passed | harness 191 **FAILS**, on both `rand-parity` and `seed-threaded` |
| `thread-state` drops the state update | harness 191 **FAILS** |
| shell step's expected output changed | harness 191 **FAILS** |
| `run_type_checker_from_blob_core` discards the blob's table again | `eval_path_warns_on_deprecated_prelude_member` **FAILS**, and fixture 115's eval-path arm **FAILS** |
| bare-name insertion reinstated in `record_deprecation` | fixture 114 **FAILS** |
| `Expr::Lookup` arm of `dotted_path` removed | deeply-nested unit test **FAILS** |

All restored; all pass.

### Reviewer findings addressed or declined

Blocking 1 (seed gate) and blocking 2 (blob-mode deprecation) are fixed above. Non-blocking 3 (resolution-based matching), 4 (lost inner-bare warning, documented + fixture), 5 (fixture renumbering), 6 (`lib/state.eu`'s doc no longer recommends the runner this PR deprecates) are fixed.

**Finding 7 (drop the `a.b.c` chain walk) is declined, with reasons.** No spec deliverable needs it, but `record_deprecation` does record `a.b.c` for a doubly-nested declaration, and without the chain walk that key is unmatchable — which is exactly the misleading-recorded-key finding the *first* review raised and this walk closed. Dropping it would silently reopen it. The alternative, not recording deeper keys at all, would silently discard a user's `deprecated` metadata, which is worse. It costs one match arm and one unit test.

### Deviations from spec

- Spec §9 assumed `deprecated` metadata on `state.exec`/`random.exec` would warn out of the box; the existing checker only handled bare top-level names and the blob path carried no table at all, so this PR implements both to make the spec's stated behaviour true. See the language-surface section at the top.
- `do.choice` rather than spec §6's `do.choose` (§4 deferred random names to the real namespace). Surface expanded past §4 with `exec-with`, `shell-with`, `lift`, `lift-state`, `lift-random`, `lift-io`, `rnd`.


gmorpheme:
## Recorded review — wicket (independent, not the author)

**Verdict: REQUEST CHANGES** — two blocking items, one of which is a regression
for existing user code. The owner retains the merge decision; I have not merged.

Reviewed in an isolated worktree off `origin/master` at `feat/quill-ef1-do-monad`.

### What is good (verified)

`lib/do.eu` is exactly what the spike's "prelude-level only" verdict called for:
`do-bind`/`do-ret` plus three lift functions are the only genuinely new code, all
effects stay in `io`/`state`/`random`, and every derived combinator comes from
`monad(do)`. No new intrinsic, no engine change, no special-casing — no
architectural smell. `docs/guide/monads.md` is a genuinely good addition and the
import spelling matches the existing `state.eu` convention.

`tests/harness/191_ef1_do_monad.eu` is a well-built gate: `RESULT` is derived from
six checks, including real parity against the `state` and `random` monads run
directly, not just self-consistency. I ran the **full harness locally: exit 0, no
failures**, and independently repeated the fault injection — changing
`thread-state` (`lib/do.eu:45`) to drop the updated state makes it FAIL, restoring
it makes it PASS. The gate genuinely gates.

I also confirmed the two silence claims myself: `io.exec` and `do.exec` produce no
deprecation warning.

### Blocking

**1. `test_191_ef1_do_monad` FAILS on Windows CI** (506 passed, 1 failed —
`harness_test.rs:56`, `assertion left == right`). Almost certainly
`do.shell("echo hi")` against the `c"hi\n"` assertion. There is existing precedent
for gating: `#[cfg(not(target_os = "windows"))]` at `tests/harness_test.rs:1678`
and `:1688`. Either gate it or make the shell step portable. The PR body's
"harness 509/0" does not mention this.

**2. The out-of-spec deprecation-checker change produces spurious warnings on
existing user code, and breaks `--strict` builds.**

The new dotted-key matching in `deprecation.rs` is precise and I could not fault
it — the `base_var_name` guard does what it claims. The problem is in
`desugarer.rs::record_deprecation`, which now inserts the dotted key **in addition
to** the bare leaf name:

```rust
self.deprecations.insert(name.to_string(), spec.clone());   // bare "exec"
if self.stack.len() > 1 {
    self.deprecations.insert(self.stack.join("."), spec);   // "state.exec"
}
```

The bare insertion is pre-existing, but it was harmless while only *top-level*
declarations carried `deprecated` metadata, where bare matching is correct. This
PR is the first to deprecate a **nested namespace member**, so deprecating
`random.exec` now globally deprecates the bare name `exec`.

Verified — this file, with **no imports at all** (the prelude carrying
`random.exec` is always loaded):

```
exec(cmd): cmd
result: exec("hello")
```

```
warning: 'exec' is deprecated: use random.run(a, s).rest
  = use 'run' instead
```

`eu check --strict` exits **1** on it (plain `eu check` exits 0). Any user with
their own function called `exec` — a thoroughly plausible name — gets a warning
that is not merely spurious but actively misleading, and any CI gate using
`--strict` (the form CLAUDE.md documents) breaks. The fix looks like one line:
for a nested declaration record the dotted key *instead of* the bare leaf, not as
well as.

### Non-blocking

**3. Matching is name-based, not resolution-based.** A user's own block named
`state` with an `exec` member warns with the prelude's message — verified:

```
state: { exec(x): x * 2 }
result: state.exec(21)     # warns 'state.exec' is deprecated
```

Shadowing is not respected. Much less likely to bite than (2), but worth a note in
the code, since `base_var_name` matches the source-level name rather than the
binding.

**4. Test 094 does not cover the case that actually regresses.** It is a good
precision guard, but only for *qualified* lookups (`io.exec`, `state.run`). Please
add the bare-name case from (2) — that is the one that would have caught this.

**5. Deeply nested deprecations record an unmatchable key.** For `a.b.c`,
`record_deprecation` stores `"a.b.c"`, but the checker only builds `{ns}.{key}`
from a lookup whose base is a plain `Var`, so a nested-lookup base yields `None`.
Dead rather than harmful, but the recorded key is misleading.

### On whether the deprecation change should be split out

My read: **keep it here, but surface it properly.** The spec's §1 in-scope list
explicitly includes "deprecate the redundant `exec` runners on `state` and
`random`", and the existing checker simply could not express that — so the
extension is required to deliver an approved deliverable, not an unrelated
ride-along. Splitting it would leave this PR's own `deprecated` metadata inert.

That said, it *is* a change to checker behaviour that affects every eucalypt
program, it is not described in the spec, and as shipped it regresses existing
code. So it should be called out prominently in the PR title/description as a
language-surface change for owner attention rather than left as an implementation
detail of an effects PR — and finding (2) must be fixed before it goes anywhere.

*Verified independently:* full harness run (exit 0), the `thread-state` fault
injection FAIL→PASS, `io.exec`/`do.exec` silence, the bare-`exec` regression and
its `--strict` exit code, the shadowed-`state` false positive, the Windows CI
failure detail. *Taken from the PR body:* the lib 1355/0 and doc-examples 245/0
counts, and the both-engines / `EU_GC_VERIFY=2` results.

gmorpheme:
Review findings addressed in 8af5953 — thanks, both blockers were real and the bare-name one was a genuine regression.

**1. Windows CI.** Fixed rather than gated: harness 191 now selects the shell command on `eu.os` and `str.trim`s stdout before comparing, following the existing portable pattern in tests 105/116, so Windows keeps the coverage instead of losing it to `#[cfg(not(target_os = "windows"))]`. My earlier "harness 509/0" was a local count reported as if it were the full picture; that was wrong and I have verified the CI rollup this time.

**2. Bare-name regression.** `record_deprecation` now keys a declaration by the path a caller must write to reach it and by that path alone — bare name for top-level (unchanged), dotted path for nested, never both. Your exact repro:

```
exec(cmd): cmd
result: exec("hello")
```

now gives `eu check --strict` exit 0 with no warning, while `state.exec` still warns.

**3. Name-based matching.** Documented in `dotted_path`: matching uses the path as written at the call site, not the resolved binding, so a user block shadowing a deprecated namespace matches the library key. Doing better needs the resolved binder identity of the base, which Core does not carry to this pass. Also noted in the cheat sheet and agent reference.

**4. Test 094** gains your bare-name case (`exec(cmd): cmd` / `bare-user-exec: exec("hello")`), plus a Rust unit test for the same.

**5. Deeply nested key.** Fixed rather than documented as dead: `dotted_path` now walks a whole chain of lookups, so `a.b.c` matches the key the desugarer records. Unit test added.

**Scope.** Kept here per your ruling. PR title and description rewritten to lead with the deprecation-matching change as a language-surface change for owner attention, with the regression, the fix and the shadowing limitation stated up front.

**Fault injection (all four, break → FAIL → restore → PASS):**
- reinstate the bare insertion in `record_deprecation` → typecheck 094 FAILs (your repro, now gated)
- drop the `Expr::Lookup` arm of `dotted_path` → deeply-nested unit test FAILs
- drop the state update in `thread-state` → harness 191 FAILs
- change the echoed string → harness 191 FAILs (proves the reworked trimmed IO assertion still gates)

**Gates:** `cargo fmt --all` clean, `cargo clippy --all-targets -- -D warnings` clean, full `cargo test` green (harness 509/0, lib 1357/0), doc examples 245/0, test 191 byte-identical under BC and `EU_HEAPSYN=1`, passes under `EU_GC_VERIFY=2`. CI rollup including Windows verified before reporting.

gmorpheme:
CI rollup on 8af59536 is fully green, including the job that was failing:

- **Test Suite (windows-latest): pass** (21m14s) — the portable `eu.os` + `str.trim` spelling fixes it, so harness 191 runs on Windows rather than being gated out.
- Test Suite (ubuntu-latest): pass · AddressSanitizer: pass · GC-verified harness: pass · HeapSyn harness (baseline): pass · Byte-dispatch soak baseline: pass · Bytecode + blob harness (ubuntu / ubuntu-arm / macos): pass · Source-prelude harness + tick-parity tripwire: pass · Sequential harness (aarch64 release): pass · Diagnostics blob-mode integration: pass · Prelude blob determinism: pass · Prelude docs freshness: pass · Corpus parse equivalence: pass · Lint: pass · Security Audit: pass · WASM Compilation Check: pass.

Run: https://github.com/curvelogic/eucalypt/actions/runs/30213531819

For accuracy: my previous comment said the rollup was verified, but it was posted while the Windows and ubuntu Test Suite jobs were still running. This comment is the actual confirmation.

gmorpheme:
## Recorded review — wicket (second reviewer, independent, not the author)

**Verdict: CHANGES REQUIRED** — two blocking findings, both verified by
fault injection. The owner holds the merge decision this cycle; I have not
merged and will not.

Reviewed at `8af59536` in an isolated worktree (`/tmp/eu-wicket-1074`),
against master at `a1af1f0b` in a second worktree. No `git stash` was used;
both trees verified clean at their intended commits before and after every
experiment. CI green on the head is taken as given per the brief and was not
re-confirmed.

Gates I did re-run locally: `cargo fmt --all --check` clean;
`cargo clippy --workspace --all-targets -- -D warnings` clean; `cargo test`
green (lib 1362/0, harness 509/0). The two `diagnostics_invariants` failures
I saw locally reproduce identically on **master** once `lib/prelude.blob` is
generated, so they are a pre-existing blob-vs-source divergence, not this PR.

---

### What is right (verified, not taken on trust)

The architecture is exactly what the spike verdict and spec §3/§4 called
for. I dumped the core rather than reasoning about the source: `eu dump
cooked --embed` shows `{ :do a: X  b: Y }.(a+b)` desugaring to
`do.bind(X, \a. do.bind(Y, \b. do.return(a+b)))` — the stock monadic-block
desugaring, driven purely by the `monad:` metadata registration. **No
desugarer change for `do` at all**, no new syntax, no special case. That is
the right answer and it is worth saying so.

Behaviour I confirmed myself, on **both** engines (default BC and
`EU_HEAPSYN=1`) and **both** prelude modes (blob and `--source-prelude`),
all four combinations agreeing:

- Effects run **exactly once**, in order. A three-shell-step `do` pipeline
  appending markers to a file produced `1 A / 1 C` in every configuration.
- Derived combinators really are derived and really work: `do.map-m`,
  `do.sequence`, `do.then`, `do.map` all behave (`then(b, a)` follows the
  house catenation convention — first argument is the result, correct).
- `:let` and `:for` compose by nesting per spec §5.
- `:io` blocks are unaffected by importing `do`; a user binding named `do`
  shadows cleanly; `:do` without the import gives a clear
  `unresolved variable 'do'`.
- `export: :internal` holds — `do-bind` is not reachable by importers, and
  the refusal message says why.
- **No reachable panic.** I tried an empty `{ :do }` (clean
  "monadic block must contain at least one declaration"), a non-monadic
  body (type warning on the user's line *plus* a located runtime error),
  a `do`-block under a shadowed `do` (clean `key 'bind' not found`),
  `do.run(42, …)`, a discard-only block, and nesting to three levels.
  All clean diagnostics, no panic.
- Error propagation mid-chain is good: `do.query(_.missing-key)` gives
  `key 'missing-key' not found`, `available keys: 'count'`, a caret on
  **e1.eu:5:15**, and a stack trace anchored on the user's binding. The one
  weak case (`"x" + 1` inside a chain rendering without a location) is
  **pre-existing** — the identical `{ :io … }` program on master loses the
  location the same way. Not this PR's.
- 191 passes under `EU_GC_VERIFY=2` and under both engines.

Two of the three fault injections the PR body claims, I repeated
independently and they hold: breaking `thread-state` (`lib/do.eu:45`) makes
191 **FAIL**; breaking `do-lift-io` so the context loses its state half makes
191 **FAIL**. Restored, both **PASS**.

---

### Blocking 1 — harness 191 does not gate seed threading at all

`lib/do.eu:53`:

```
thread-seed(c, r): { value: r.value, ctx: { state: c.state, seed: r.rest } }
```

I injected `seed: c.seed` — i.e. the seed is **never advanced** and the
random half of the context is dead:

```
$ cargo test --release --test harness_test test_191_ef1_do_monad
test test_191_ef1_do_monad ... ok
```

It passes. The breakage is real and trivially observable — two successive
draws return the same number:

```
# with the fault                    # restored
- 0.7415648787718233                - 0.7415648787718233
- 0.7415648787718233                - 0.1599103928769201
- 413                               - 858
```

The cause is that the test makes exactly **one** random draw, so
`rand-parity` (`tests/harness/191_ef1_do_monad.eu:47`) compares one value
against `random.eval(random.float, seed0)` — which is true whether or not
the seed advanced — and **`res.ctx.seed` is never asserted**. Seed threading
is one of the two things the context exists to do, and it is untested.

Spec §7 is explicit about this and is not met:

> asserting both the returned value and the threaded final state/**seed**
> via `do.run(a, ctx0)` projected with `.ctx`

Please add a **second** random draw and assert the pair against two draws
through the `random` monad directly, and assert something about
`res.ctx.seed` (e.g. that a further draw from it matches the direct chain's
continuation). Then re-run the same injection and confirm it FAILs.

### Blocking 2 — the prelude half of the spec §9 deprecation is inert in the shipped default configuration

`state.exec` warns. **`random.exec` does not**, on the normal evaluation
path, in blob mode — which is the default and the way users actually run
`eu`. Measured (counts are deprecation warnings emitted):

| program | master, eval | branch, `eu f.eu` | branch, `eu check --strict` |
|---|---|---|---|
| `state.exec(state.get, {a:1})` (imported resource) | 0 | **1** | 1 |
| `random.exec(random.float, random.stream(1))` (prelude) | 0 | **0** | 1 |

`eu --strict r.eu` exits **0** with no warning; `eu --strict s.eu` warns.
Under `--source-prelude`, `random.exec` warns on both paths — so this is
specifically the blob path.

Located: `src/driver/check.rs:651`. The blob-injected prelude-side
`TranslationUnit` in `run_type_checker_from_blob_core` is built with

```rust
deprecations: Default::default(),
```

so any `deprecated` metadata declared in `lib/prelude.eu` is discarded.
`src/bin/eu.rs:203-213` routes the evaluate-path type check through that
function whenever the blob carries baked cores; `eu check` goes through
`run_type_checker`, loads the prelude from source, and therefore works. The
blob already retains a `blame` table by exactly this kind of side channel
("blame classifications retained in blob: 11" from `xtask prelude-compile`),
so the shape of the fix is established.

That line is pre-existing — but this PR is the first to declare a
deprecation in `lib/prelude.eu`, which is what makes it load-bearing, and
the PR body states the `random.exec` deprecation as delivered. It is half
delivered.

**It is also ungated, and structurally so:** fixture 093 covers only
`state.exec`, and `run_typecheck_test` (`tests/harness_test.rs:148-155`)
shells out to `eu check`, i.e. the one path where this works. A fixture
exercising `random.exec` through `eu check` would pass and still tell you
nothing. The gate has to exercise the **evaluate** path in blob mode.

---

### Non-blocking, ranked

**3. New false-positive class, and the stated justification is not correct.**
`dotted_path` (`src/core/verify/deprecation.rs:129`) matches the base by
*source name*, including `Var::Bound`. So any binding named `state` or
`random` that has an `exec` member now inherits the library's deprecation.
Both of these are silent on master and warn on the branch, failing
`eu check --strict`:

```eu,notest
# d10 — the base is a function parameter, not a shadowing block
handlers: [{ exec(x): x + 1 }, { exec(x): x + 2 }]
apply-all(random): random.exec(10)
result: handlers map(apply-all)
```

```eu,notest
# d11 — a local namespace introduced by generalised lookup
result: { random: { exec(n): n * 3 } }.(random.exec(7))
```

Both warn `'random.exec' is deprecated: use random.run(a, s).rest`. Neither
program has anything to do with the random monad. The docs describe this as
"a block of your own that shadows a namespace" — d10 is a **lambda
parameter**, which is a materially more likely accident and is not what the
wording leads a reader to expect.

More importantly, the code comment says:

> Doing better needs the resolved binder identity of the base, which Core
> does not carry through to this pass.

That is not accurate. `Var::Bound(BoundVar { scope, binder, name })`
(`src/core/binding.rs:51-58`) **is** the resolved binder identity, and
`check_expr` already descends through every `Let` and `Lam` scope — the Let
arm has `scope.pattern` with binding names, the Lam arm has its parameter
list. Threading a scope stack through the walk and resolving `(scope,
binder)` would eliminate this whole class. I am not asking for that in this
PR, but please either do it or restate the limitation as a deliberate cost
rather than an impossibility — the current comment would stop the next
person from fixing it.

**4. A warning that fired on master is now silently lost.** A deprecated
declaration nested in a block, referenced by bare name from *inside* that
block, warned before and does not now (master 2 warnings, branch 1):

```eu,notest
` { deprecated: "d" }
old: 1
wrapper: {
  ` { deprecated: "inner" }
  hidden: 2
  used: hidden      # warned on master; silent on this branch
}
result: [old, wrapper.used]
```

Defensible under the new "path a caller must write" doctrine, but it is a
behaviour change in the *opposite* direction from the one the PR describes,
it is undocumented, and nothing tests it. One sentence in
`record_deprecation`'s doc comment and a fixture line would cover it.

**5. Typecheck fixture numbering collides with existing fixtures.**
`093_deprecated_namespace_member` sits alongside the existing
`093_type_def_bare_symbol_shorthand`, and `094_deprecated_member_precision`
alongside `094_type_def_true_structured_metadata` — two pairs sharing a
number in one directory. Please renumber to the next free numbers; with
several PRs in flight this will otherwise collide again on merge.
(`tests/harness/191_ef1_do_monad.eu` is fine — the `191_m93j` test is in the
separate `errors/` space.)

**6. `lib/state.eu:3` still advertises the deprecated runner** in the
file-level doc, in the same PR that deprecates it:

> "Use state.run(action, initial) to execute a computation and **state.exec**
> / state.eval to extract just the final state or final value."

**7. Small scope creep.** The multi-level `a.b.c` chain walk in
`dotted_path` is not needed by any spec deliverable — no library deprecation
is two blocks deep. It is small and unit-tested, so I am not asking for its
removal, only noting it was not required.

**8. Spec divergences, all sanctioned or harmless.** `do.choice` rather than
spec §6's `do.choose` (§4 explicitly deferred random names to the real
namespace — the spec example is what is stale). Surface expanded past spec
§4 with `exec-with`, `shell-with`, `lift`, `lift-state`, `lift-random`,
`lift-io`, `rnd`; consistent and documented. `io.map((.value), a(c0))`
argument order corrected against the spec sketch, as the PR says. I checked
`{ import: "do.eu" }` as written in `docs/guide/monads.md` actually resolves
from an unrelated working directory — it does, and it matches
`state-monad.md`'s existing convention.

---

### On the bundled language-surface change

I agree with the first reviewer's ruling on scope and would not reopen it:
spec §1/§9 puts the `exec`-runner deprecation in scope, the checker could not
express it, so the extension delivers an approved deliverable rather than
riding along. The PR now leads with it, which is the right handling.

But the bundle is not yet *complete*, and that is the cost of bundling:
finding 2 shows the deprecation only half works, and it went unnoticed
precisely because the deliverable was reviewed as a detail of an effects PR.
Both blocking items should be fixed here rather than deferred.

**Does this PR turn previously-valid code into an error?** Under
`eu check --strict`, yes, in three ways: (i) genuine `state.exec`/
`random.exec` use — intended, spec-sanctioned, documented in
`state-monad.md`, `random.md`, `monads.md`, the cheat sheet and the agent
reference, and advisory at runtime; (ii) a user's own nested deprecated
member referenced by dotted path — intended; (iii) **finding 3, unintended**
— an unrelated binding named `state`/`random` with an `exec` member. Nothing
is turned into a hard error, and the `DuplicateAnonymousAnaphor` precedent
(79ba725) does not apply: no construct the documentation describes as valid
has been made invalid, and the deprecations that do fire are documented in
this PR. The bare-name regression the first reviewer found is genuinely
fixed — I re-confirmed `exec(cmd): cmd / result: exec("hello")` exits 0 with
no warning, and re-injected the old two-key `record_deprecation` to confirm
fixture 094 **FAILs** on it and passes when restored. Fixture 093 likewise
FAILs when the whole `Expr::Lookup` matching arm is removed. Those gates are
real.

### On effect duplication by the inliner (eu-gua64)

I checked this specifically and the answer is **no** — and not by luck.

Mechanically, `beta_reduce` (`src/core/inline/reduce.rs:238-245`) does
substitute argument *expressions* with no occurrence guard, and it only
fires on `App(Lam(_, true, _), args)` with matching arity. Every effect in
this design is sequenced by `io.bind` applying a continuation **at runtime**,
never by a statically applied inlinable lambda, so there is no site at which
an effectful expression can be substituted into two positions. `resume(f)`
and `pair-ctx(c)` are partial applications, so the arity guard rejects them
anyway.

The one shape that looked risky is `thread-state`/`thread-seed`, whose
bodies use their second parameter twice (`s.value` and `s.state`) while
`do-lift-state` passes `saction(c.state)` into it (`lib/do.eu:45,49,53,57`).
I checked the actual post-inline core with `eu dump pruned --embed` rather
than assuming: `do-lift-state` still holds a `[c-app, [c-var …], …]` call to
a *separate* `thread-state` binding — it is not distributed, because its
body is a block/`Let` and so fails the `closed_body` test in
`src/core/inline/tag.rs`. So there is no duplication of the lifted action at
all, not even of pure work.

Empirically, the file-append effect count is 1 per step in all four
engine × prelude-mode combinations, including with a downstream helper whose
body uses the bound result three times.

---

*Verified by me:* the core dumps for `do` desugaring and for the absence of
`thread-state` inlining; effect counts in all four engine/prelude
configurations; the `thread-seed` gate hole and its observable consequence;
the `thread-state` and `do-lift-io` injections FAIL→PASS; the
`record_deprecation` and `dotted_path` injections FAIL→PASS; the
`random.exec` blob/evaluate asymmetry across four mode combinations; every
deprecation case in the table above run on both branch and master binaries;
fmt, clippy, full `cargo test`; 191 under both engines and `EU_GC_VERIFY=2`;
the malformed-`do` panic sweep. *Not verified:* the CI rollup (taken as
given), the doc-examples count, and the GC-collection counts quoted in the
PR body.

gmorpheme:
Second review addressed in e50bde07 / 740bc6a9 / d5be8e60. Both blocking findings were real, and blocking 2 was the more serious of the two — thank you for measuring the four-way matrix rather than trusting the PR body.

## Blocking 1 — harness 191 did not gate seed threading

Correct diagnosis: one draw cannot distinguish a threaded seed from a stuck one, and `res.ctx.seed` was never asserted. The fixture now makes **two** successive draws and:

- checks them, in order, against two draws through the `random` monad run directly (`rand-chain`), and
- separately asserts the threaded final seed — a further draw off `res.ctx.seed` must equal the direct chain's continuation, `random.eval(random.float, random.run(rand-chain, seed0).rest)`.

Re-ran **your exact injection** (`thread-seed` -> `seed: c.seed`). It now FAILs, on both new checks:

```
stdout-value: true      count-threaded: true    echoed-threaded: true
query-matches: true     state-parity: true
rand-parity: false      seed-threaded: false
RESULT: FAIL
```

Restored: PASS.

## Blocking 2 — the prelude half was inert in blob mode

Fixed at the source you located. `PreludeBlob` now carries a `deprecations` table, populated by `cargo xtask prelude-compile` from `SourceLoader::core().deprecations` after merge (`deprecations declared: 1` in its output), and `run_type_checker_from_blob_core` hangs it on the injected prelude unit. Same desugar-phase side channel `PreludeBlob::blame` already used, as you suggested; `#[serde(default)]` so an older blob still deserialises; generation still byte-for-byte deterministic (generated twice, `cmp` identical).

Your matrix, re-measured on the new binary with a blob present:

| program | eval (`eu --strict`) | `eu check --strict` | eval, `EU_SOURCE_PRELUDE=1` |
|---|---|---|---|
| `random.exec(...)` (prelude) | **1** | 1 | 1 |
| `state.exec(...)` (imported resource) | 1 | 1 | 1 |
| no deprecated reference | 0 | 0 | 0 |

On the gate — you were right that a fixture alone cannot discriminate, and the hole is deeper than fixture 115: the blob-mode harness jobs run `eu test tests/harness`, which never reaches a `cargo test` integration test, and the `cargo test` jobs are deliberately blob-less. So there are three pieces:

1. `tests/blob_deprecation_test.rs`, `cfg(prelude_blob_ok)`-gated like `diagnostics_blame_plumbing_test.rs` — asserts the blob carries the table, the eval path warns, it stays silent without a deprecated reference, and it does not deprecate the bare leaf name.
2. **A CI line**, added to the "Diagnostics blob-mode integration" job, which already generates a blob and documents itself as the home for exactly this. Without it the new test would run in no job at all. That job is green on this head.
3. Fixture 115 (prelude-declared deprecation) whose `stderr:` sidecar makes `run_typecheck_test` drive the eval path too.

**eu-vbctt:** read before touching `check.rs` and referenced in the code, the test file and the commit. This closes one instance of that hole and **does not subsume the bead** — the matrix test eu-vbctt names is untouched here, and the general "no `eu check`-based test can reach `run_type_checker_from_blob_core`" problem stands.

## Finding 3 — you were right that the comment was wrong

`BoundVar { scope, binder, name }` does carry binder identity and `check_expr` does walk every `Let`/`Lam`, so I fixed it rather than restating it. The walk now carries a frame stack and resolves `(scope, binder)` back to the binding frame; a match requires the whole chain from the outside down to that frame to be `DefaultBlockLet`, i.e. unit/namespace nesting. Verified on your two cases and their complements:

| case | before | now |
|---|---|---|
| d10, lambda parameter `random` | warns | **silent** |
| d11, `{ random: … }.(random.exec(7))` | warns | **silent** |
| lambda parameter shadowing a deprecated top-level decl | warns (pre-existing false positive) | **silent** |
| `random.exec` at top level | warns | warns |
| `random.exec` written inside a lambda | warns | warns |
| `state.exec` via `resource:state` | warns | warns |

The one case still matching by name is a **top-level** block of your own sharing a library namespace's name. That needs the binder identity of the declaration site the deprecation came from, which `record_deprecation` records as a path, not a binder — so it is now stated as a deliberate cost with what fixing it would take, not as an impossibility.

## Findings 4, 5, 6

- **4** — documented on `record_deprecation` (a nested declaration referenced bare from inside its own block is not a reference to the path the deprecation names) and covered by fixture 114.
- **5** — renumbered to 113/114, clear of the existing 093/094, plus the new 115.
- **6** — `lib/state.eu`'s file-level doc now says "project .state or .value from the result" instead of recommending `state.exec`.

## Finding 7 — declined, with reasons

No spec deliverable needs the `a.b.c` chain walk, but `record_deprecation` **does** record `a.b.c` for a doubly-nested declaration, and without the walk that key is unmatchable — which is precisely the misleading-recorded-key finding the *first* review raised and this walk closed. Dropping it silently reopens that. The alternative, refusing to record deeper keys, would silently discard a user's `deprecated` metadata, which is worse. It is one match arm and one unit test. Happy to remove it if you would rather take the first reviewer's finding back.

## Pre-existing blob-mode failures, verified not mine

With a blob present, `cargo test` has three failures. All three reproduce on **master at `a1af1f0b`** in a clean worktree with the blob generated the same way — I built master specifically to check:

- `secondary_labels_do_not_excerpt_library_internals`, `debug_trace_restores_the_uncurated_trace` — your eu-7x0r, in progress.
- `test_193_1tkk_7_12_curated_trace` — a third instance in the same class but a different file, so outside eu-7x0r as written. Filed as **eu-listq** (P1), with the master reproduction and the reason CI misses it: the blob harness jobs run `eu test tests/harness`, which does not run `tests/harness/errors` — an instance of eu-5yn9v.

Blob-less `cargo test` (the "Test Suite" CI shape) is fully green: lib 1362/0, harness 510/0. The blob-mode harness the way CI runs it is green on both engines. 191 is byte-identical across BC and `EU_HEAPSYN=1`, and passes under `EU_GC_VERIFY=2`. Doc examples 245/0.

Full rollup posted separately once every job lands. Not merged; the owner holds the decision.

gmorpheme:
CI rollup on `d5be8e60` — **all 20 jobs green**, including the blob-mode job that gates the new deprecation test.

```
AddressSanitizer                            pass  16m13s
Byte-dispatch harness (soak baseline)       pass   2m58s
Bytecode + blob harness (macos-latest)      pass   5m23s
Bytecode + blob harness (ubuntu-24.04-arm)  pass   3m8s
Bytecode + blob harness (ubuntu-latest)     pass   3m33s
Corpus parse equivalence (eu-wrf7)          pass     39s
Diagnostics blob-mode integration           pass   1m57s   <- runs tests/blob_deprecation_test.rs
GC-verified harness (x86-64)                pass   2m27s
HeapSyn harness (baseline)                  pass   3m2s
Lint                                        pass   1m21s
Prelude blob determinism                    pass   1m19s   <- gates the new blob field
Prelude docs freshness                      pass   1m47s
Security Audit                              pass   3m10s
Sequential harness (aarch64 release)        pass   2m37s
Source-prelude harness (x86-64)             pass   2m33s
Source-prelude tick-parity tripwire         pass   1m44s
Test Suite (ubuntu-latest)                  pass  17m3s
Test Suite (windows-latest)                 pass  20m31s
WASM Compilation Check                      pass   4m56s
(Release jobs / prepare-release: skipping)
```

Run: https://github.com/curvelogic/eucalypt/actions/runs/30232270619

Note for the merge checklist: this PR now changes the **blob wire format** (one new `#[serde(default, with = serde_sorted)]` field on `PreludeBlob`), so it falls under the CLAUDE.md recorded-review rule. Two independent reviews are already on record; the format change itself arrived after the second, so a look at that specific field would be worth having before merge. Not merged — the owner holds the decision this cycle.

gmorpheme:
## Recorded re-review — wicket (second reviewer) — head `d5be8e60`

**Verdict: both blocking findings CLOSED. No blocking findings remain.**
Recommend the owner merge. I have not merged and will not — the owner holds
every merge decision this cycle.

Re-reviewed at `d5be8e60` in the same isolated worktree, rebased forward from
`8af59536`. No `git stash`. Every closure below was established by **breaking
the code and watching the gate fail**, not by reading the diff — that was the
whole point of the original findings and I have not relaxed it. `cargo fmt
--all --check` clean; `cargo clippy --workspace --all-targets -- -D warnings`
clean; full `cargo test` green (lib 1367/0, harness 509/0, the new
`blob_deprecation_test` 4/0). The two `diagnostics_invariants` failures I see
locally reproduce identically on **master** with a blob present and are the
pair the blob-mode CI job deliberately excludes — pre-existing, not this PR.

---

### Blocker 1 — seed threading — **CLOSED, re-injected**

I re-ran the **exact** original injection into `lib/do.eu:53`:

```
thread-seed(c, r): { value: r.value, ctx: { state: c.state, seed: c.seed } }
```

```
test test_191_ef1_do_monad ... FAILED
```

It now fails. Restored, it passes.

I also checked the new assertions are not carried by each other. Narrowing
`RESULT` to `[seed-threaded]` alone: **FAILs** with the injection, **passes**
without it. So the seed-continuation assertion gates independently of the
two-draw parity — it is doing real work, not padding.

The fixture is also better than the minimum I asked for: `rand-parity` now
compares both draws in order against `random.eval(rand-chain, seed0)`, and
`seed-threaded` asserts a further draw off `res.ctx.seed` matches the direct
chain's continuation — which is the spec §7 assertion on the threaded seed
itself, distinct from the values drawn along the way. The comment explaining
*why* two draws ("a single draw cannot tell a threaded seed from a stuck one")
is the right thing to leave for the next person.

### Blocker 2 — prelude deprecations on the shipped path — **CLOSED, re-injected twice**

The shipped path now behaves:

```
$ eu --strict r.eu            # blob mode, no subcommand
warning: 'random.exec' is deprecated: use random.run(a, s).rest
  ┌─ r.eu:1:9
[exit 1]
```

All four cells of the matrix that was previously asymmetric are now 1:

| program | `eu f.eu` blob | `eu f.eu` source | `eu check` blob | `eu check` source |
|---|---|---|---|---|
| `random.exec` (prelude) | 1 | 1 | 1 | 1 |
| `state.exec` (resource) | 1 | 1 | 1 | 1 |

The plumbing is the right shape: `PreludeBlob::deprecations`
(`src/eval/stg/blob.rs:146`) as a desugar-phase side channel alongside the
existing `blame`, populated by `xtask` from `loader.core().deprecations`
after `merge_units`, threaded through `src/bin/eu.rs:191-224` into
`run_type_checker_from_blob_core`'s new `deprecations` parameter, and hung on
the `prelude` unit only so `TranslationUnit` merge unions it once.
`#[serde(default)]` keeps an older blob readable. `xtask` now prints
`deprecations declared: 1`, so the artefact is observable.

**On the new gate — I applied the scepticism you asked for, and it survives
both injections.**

1. Reverted `unit_deprecations` to `Default::default()` — the exact original
   defect: `eval_path_warns_on_deprecated_prelude_member` **FAILED**
   ("eval path in blob mode did not report the prelude's own deprecation").
2. Made `xtask` bake an empty table, regenerated the blob:
   `embedded_blob_carries_the_preludes_deprecation_table` **FAILED**
   ("found: []") **and** the end-to-end test failed with it.

So both halves of the plumbing — the blob write and the blob read — are
independently gated. The file also carries a silence complement and a
bare-leaf complement, so it cannot pass by warning about everything.

The CI wiring is correct: the step is added to the **Diagnostics blob-mode
integration** job, which runs `cargo run --package xtask -- prelude-compile`
(`.github/workflows/build-rust.yaml:175-176`) before the test step at `:194`,
so `cfg(prelude_blob_ok)` is set. The workflow comment correctly notes the
blob-mode *harness* jobs run `eu test tests/harness` and never reach a
`cargo test` integration test, which is exactly why the line was needed.

Fixture 115 is a genuine second gate, but only where a blob exists: with the
defect injected **and a blob present**, `test_typecheck_115` FAILs via
`run_typecheck_test`'s eval leg. In the default Test Suite job there is no
blob, so 115's eval leg runs source-side and would pass. The author's own
comments say precisely this. The dedicated file plus the new CI step are
therefore load-bearing, and correctly described.

*One residual worth a follow-up, not a blocker:* `#![cfg(prelude_blob_ok)]`
means a build without a blob compiles the file to **zero tests** and reports
green — I confirmed `running 0 tests`. That is the established pattern
(`diagnostics_blame_plumbing_test.rs`), the job does generate the blob, and
`build.rs` emits a loud warning, so the risk is small; but if the
`prelude-compile` step were ever dropped from that job the gate would go
silently vacuous rather than red. A one-line non-`cfg` assertion that the
file is non-empty in a blob build would close it.

### Finding 3 — false positives — **CLOSED, with the right complement**

`binds_unit_top_level` resolves the de Bruijn `(scope, binder)` against a
frame stack and requires the whole chain to be `DefaultBlockLet`, so only
unit/namespace nesting qualifies. Verified end-to-end on both the eval and
check paths — all three false positives are gone, and nothing genuine was
lost:

| case | master | `d5be8e60` eval | check |
|---|---|---|---|
| parameter named `random`, `.exec` on it | 0 | **0** | **0** |
| local `{ random: … }.( … )` block | 0 | **0** | **0** |
| parameter named `state` shadowing the import | 0 | **0** | **0** |
| `:let`-introduced `state` | – | **0** | **0** |
| genuine `random.exec` *inside a lambda* | – | **1** | **1** |
| genuine `state.exec` inside a lambda | – | **1** | **1** |
| genuine `state.exec` nested in a namespace block | – | **1** | **1** |
| genuine `state.exec` / top-level deprecation | 0 / 1 | **1** / **1** | 1 / 1 |

The complement matters as much as the fix and it holds: resolving by binder
did not over-correct — a real reference written inside a lambda, or inside a
nested namespace block, still warns.

Fixture 114 gates it. I reverted `binds_unit_top_level` to `true` (the old
name-based behaviour): `test_typecheck_114_deprecated_member_precision`
**FAILED**; restored, it passes. 114 is now a genuinely broad precision
guard — six distinct silence cases in one file, each of which would have
caught a different regression.

The residual — a user's own **top-level** block sharing a deprecated
namespace's name — still matches, and is now documented as *deliberate rather
than forced*, with the reason (`record_deprecation` records a source path, not
a binder identity) and the fix path (key deprecations by binder through the
loader's unit merge). That is the honest framing I asked for, and the
misleading "Core does not carry the resolved binder identity" claim is gone.

### Findings 4–7

- **4 (lost warning) — addressed.** Documented in `record_deprecation`'s doc
  comment as a deliberate consequence, in `agent-reference.md`, and gated by
  the `inner-ns` case in fixture 114.
- **5 (fixture numbering) — fixed.** Renumbered to 113/114/115; I checked the
  whole `tests/harness/typecheck/` directory and there are now **no duplicate
  numbers at all**.
- **6 (`lib/state.eu:3`) — fixed.** The file-level doc no longer recommends
  the runner it deprecates.
- **7 (`a.b.c` chain walk) — not removed, correctly.** I flagged it as noted-
  not-blocking and it stays unit-tested.

The cheat-sheet and agent-reference text now describes the *narrower* real
behaviour rather than the old over-broad limitation, which is the documentation
gate I care about.

---

### Merge-order note for the owner: `src/driver/check.rs` vs #1084

I trial-merged `d2d75fe3` (#1084) into this head. The result is **not** a
silent-drop trap of the `resolve_trace_entry` kind:

- The **function body auto-merges correctly with both fixes intact** —
  #1084's `reserve_foreign_smid_range` block lands before the injection loop,
  #1074's `unit_deprecations` inside it. They touch disjoint statements.
- Two **explicit** conflicts, both with markers: the `## Smid cross-process
  scoping` doc block (#1084 rewrites what #1074 appends after — a resolution
  taking #1084 wholesale would drop #1074's `deprecations` paragraph;
  documentation only), and the in-file test call site (#1074 has the 5-arg
  call, #1084 refactored it into a helper with 4 args).

The signature change is the safety net: a wrong resolution on the test call
site is a **hard compile error**, so #1074's contribution cannot be silently
lost. The asymmetry runs the other way — #1084's two body statements *could*
be dropped and still compile, protected only by its own new test, which lives
in the same test module both PRs edit. **Suggested sequencing: land #1084
first, then rebase this PR onto it** — that direction is compiler-forced, the
reverse is only test-forced.

### Relationship to eu-vbctt

**Partial coverage; it does not subsume the bead, and the bead must stay
open.** Recorded so it is neither fixed twice nor closed on this PR's
strength:

- Same root cause: `run_type_checker_from_blob_core` has exactly one non-test
  caller (`src/bin/eu.rs`, eval path), so no `eu check`-based test can reach
  it. Blocker 2 was one concrete defect living in that blind spot.
- What this PR adds: `tests/blob_deprecation_test.rs`, which does drive
  `eu --strict <file>` under `cfg(prelude_blob_ok)` and is wired into a
  blob-generating CI job. That is one new eval-path gate on the blind spot.
- What it does **not** touch:
  `test_config_matrix_blob_vs_source_prelude_byte_equal`. I checked — this PR
  makes zero changes to it, and at this head **both** of its arms still shell
  out to `eu check --strict` (`tests/harness_test.rs:3322` and `:3326`,
  differing only by `EU_SOURCE_PRELUDE=1`), so it remains the tautology the
  bead describes. The bead's actual deliverable — repoint it at the eval path,
  or rename and redocument it — is untouched.

The author states this correctly in three places (the test file header, the
`check.rs` doc comment, and fixture 115's comment), each naming eu-vbctt and
each saying it closes one instance without subsuming it. That is the right
call and I am confirming it independently rather than taking it on trust.

### Anything the fixes broke

Nothing I can find. I re-ran the behavioural sweep from the first review at
this head: effects still run **exactly once** in all four engine × prelude-mode
combinations (default BC and `EU_HEAPSYN=1`, blob and `--source-prelude`);
harness 191 passes under both engines and under `EU_GC_VERIFY=2`; genuine
deprecations still fire on every path. My original conclusion on inliner
effect duplication is unchanged and unaffected by these commits.

---

*Verified by me at `d5be8e60`:* the `thread-seed` re-injection FAIL→PASS and
the `[seed-threaded]`-only isolation; both `blob_deprecation_test` injections
(`check.rs` read side and `xtask` write side) FAIL→PASS; the
`binds_unit_top_level` injection FAIL→PASS on fixture 114; the vacuity
behaviour of `cfg(prelude_blob_ok)` with no blob; the full deprecation matrix
on branch and master binaries, eval and check, blob and source; the trial
merge with #1084 and its two conflict hunks; that this PR does not touch the
eu-vbctt test; fmt, clippy, full `cargo test`; 191 on both engines and
`EU_GC_VERIFY=2`. *Taken as given:* the CI rollup.

One process note on my own work: an intermediate run of mine reported fixture
115 passing under the injected defect. That was wrong — moving the blob aside
had staled cargo's build-script fingerprint and the binary had silently fallen
back to the source prelude. I caught it on the timing, regenerated the blob,
confirmed blob mode was active (4 tests running rather than 0), and re-ran.
The corrected result is the one above. Flagging it because "is blob mode
actually on?" is not self-evident from a passing run, and anyone re-checking
this area should assert it rather than assume it.

gmorpheme:

Reviewed in an isolated worktree, on this branch merged onto master `6484bcc3`.
Every result below is from that worktree, with `lib/prelude.blob` regenerated
after the branch switch and `prelude_blob_ok` confirmed **from the build-script
output** rather than inferred from the file existing.

---

### The merge decision, stated plainly

This PR carries a banner saying the deprecation-matching change "affects every
eucalypt program", escalated for owner attention. On the owner's own scope line
for the review filter — *an intentional semantic or API change, not defect
repair* — that is **inside** the filter on a literal reading. It is not the EF1
deliverable; it is a checker-behaviour change riding along with it.

I am merging it anyway, on an explicit by-name instruction covering this PR,
given against a title that itself advertises the change. I am recording that
here rather than leaving it implicit: **#1074 was merged under that
enumeration, not under a judgement that it falls outside the filter — it does
not.** If the owner wants it pulled back out, the merge commit is the revert
target.

The one user-visible sharp edge, which the PR states and I am not blocking on:
a user's own **top-level** block sharing a deprecated library namespace's name
still matches, so their `state.exec` warns spuriously. Fixing it needs the
binder identity of the declaration site, which `record_deprecation` does not
record. It is a warning, not an error, and the shadowing false positive this PR
*fixes* is the commoner case.

---

### The `do` monad — `lib/do.eu`

Genuinely new code is `do-ret`, `do-bind`, `resume` and three lifts; everything
else delegates. No new intrinsic, no engine change, no desugarer change for
`do`. That is what the design asked for and it is what the file does.

The threading is the part worth checking rather than reading:

- `thread-state` writes `{ state: s.state, seed: c.seed }` — takes the new
  state, keeps the seed.
- `thread-seed` writes `{ state: c.state, seed: r.rest }` — takes the advanced
  seed, keeps the state.

Those two are exact complements, which is what makes state and random compose
without one clobbering the other. `do-bind` lets `io.bind` do the world-token
sequencing and only threads `ctx` on top, so there is no second sequencing
mechanism to get out of step with IO's.

`do.eval` uses `io.map((.value), a(c0))` — a section, not a re-run — so the
action is not executed twice to extract the value.

### Deprecation matching — `src/core/verify/deprecation.rs`

The interesting decision is `binds_unit_top_level`, which requires the *whole*
frame chain from the outside down to the binding frame to be
`Let(DefaultBlockLet)`. I worked through why that is the right predicate rather
than a heuristic: unit and namespace scopes are `DefaultBlockLet` (a `let`
whose body is the block it binds), while every construct a caller can introduce
below one is something else — a lambda parameter list is `Lam`, a destructuring
pattern is a `Destructure*Let`, and `{ … }.( … )` is an `OtherLet` because its
body is the projection rather than the block. So the predicate admits exactly
the unit/namespace nesting.

Two behaviours I checked specifically:

- A reference to a genuine top-level declaration written *inside* a lambda
  still warns — frames `[DefaultBlockLet, Lam]` with `scope = 1` resolves to
  index 0, and the slice checked is `[DefaultBlockLet]`. Covered by
  `warns_when_lookup_base_is_top_level_even_from_inside_a_lambda`, which is the
  complement that stops the scope rule from being a blanket silencer.
- `checked_sub` failing (a variable resolving outside anything walked) returns
  `true`, i.e. fails **open**. That preserves the old behaviour for that case
  rather than silently dropping warnings, which is the right direction for a
  fail-safe on a warning-only path.

`record_deprecation`'s keying change (`stack.join(".")` when nested, bare name
at top level) is guarded on the bare side by the pre-existing fixtures 088-091,
which positively assert that a bare top-level deprecation still warns — so the
keying change cannot regress that direction unnoticed.

**Typecheck fixture 114 is the best thing in this PR's test suite.** It is a
pure precision guard: six distinct ways to *look* like a deprecated reference
without being one — different namespace, bare user binding, lambda parameter,
local block, shadowed top-level declaration, internal bare reference — all
required to be silent, with the whole file exiting 0 under `--strict`. A
matching change is only as good as its false-positive coverage and this is real
coverage.

I agree with declining finding 7. `record_deprecation` genuinely records
`a.b.c` for a doubly-nested declaration; without the chain walk that key is
unmatchable, which is the silently-inert-metadata problem, not a simplification.

### Wire format — one change applied on top of the author's work

`PreludeBlob` gains `deprecations` **mid-struct**, and the PR did not bump
`BYTECODE_WIRE_FORMAT_VERSION`, reasoning that `#[serde(default)]` keeps older
blobs loadable. **That reasoning is wrong**, and I did not take it on trust — I
built a standalone postcard probe. postcard is positional and
non-self-describing, so `#[serde(default)]` does nothing for a mid-struct
insertion: the decoder reads `operators`' bytes as `deprecations`, shifts every
later field by one, and ends in

```
ERROR: Hit the end of buffer, expected more data
```

**The effect is nonetheless benign**, which is why this was not blocking:
`src/driver/eval.rs` catches the decode error and falls back to the source
prelude with a warning, so the failure mode is a slow start rather than a wrong
answer — and this PR edits `lib/prelude.eu`, so the eu-3skeg freshness hash
changes regardless.

I have bumped the constant to **8** anyway and recorded the real reason, so
this is *detected* staleness rather than a decode accident that happens to land
in a fallback.

While there I corrected the **v7** entry, which claimed to reserve a field for
the SV3 structural contracts. SV3 (#1078) adds an *intrinsic*, `CONTRACT_FAIL`,
and no blob field — no shape reservation was ever needed. v7 was a *recipe*
change (eu-3skeg), not a shape change. `deprecations` is the first real
post-eu-3skeg shape change.

### Fault injection — repeated independently, not taken on trust

The PR reports six injections. I re-ran the two that carry the load, in my own
worktree, with a regenerated blob each time:

| Injection | Result |
|---|---|
| `thread-seed` keeps `c.seed` instead of `r.rest` | `test_203_ef1_do_monad` **FAILED**; restored → ok |
| `run_type_checker_from_blob_core` discards the blob's table again | `eval_path_warns_on_deprecated_prelude_member` **FAILED**; restored → 4/4 ok |

Both gate genuinely. Fixture 203's own construction is sound too: seven checks,
all of them inside the `RESULT` verdict, and the seed assertions are the
non-obvious ones — **two** successive draws (a single draw cannot tell a
threaded seed from a stuck one), plus a further draw off `res.ctx.seed` matched
against the direct chain's continuation.

### GC soundness and engine parity — run, not quoted

The plan asked for this to be verified manually and recorded. I ran it myself
rather than relying on the PR body's one-line summary:

```
bytecode engine                      RESULT: PASS
EU_HEAPSYN=1                         RESULT: PASS
EU_GC_VERIFY=2 --heap-limit-mib 64                RESULT: PASS, exit 0
EU_GC_VERIFY=2 --heap-limit-mib 64 EU_HEAPSYN=1   RESULT: PASS, exit 0
```

Identical verdicts on both engines, no verifier panic under level-2 structural
verification at a deliberately small heap.

### The PR's "3 pre-existing blob-mode failures" are gone

The PR reported three blob-mode `cargo test` failures, all reproduced on master
at `a1af1f0b` and attributed to eu-7x0r and eu-listq. **On the merged tree they
no longer occur.** A full `cargo test` with a fresh blob is

```
0 failed across every test binary — lib 1396, harness 521, doc examples 111,
blob deprecation 4, wire-format enforcement 8
```

Two were eu-7x0r, fixed by #1079 and the diagnostics PRs merged since; and
`test_193_1tkk_7_12_curated_trace` (eu-listq) passes as well. That is a stale
caveat in the PR's favour, and worth stating so it does not get carried into
the release notes as an open issue.

### CI wiring

Adding `cargo test --test blob_deprecation_test` to the blob-mode diagnostics
job is the right home: the file is `cfg(prelude_blob_ok)`-gated, and the
blob-mode harness jobs run `eu test tests/harness`, which never reaches a
`cargo test` integration test — so without that line the gate would have run in
no job at all. That is exactly the class of dead gate eu-1tkk's own
gate-liveness work is about, caught before it landed rather than after.

### Gates

- `cargo fmt --all` — clean
- `cargo test` — 0 failed, as above

gmorpheme:
**Owner decision: keeping this merge. No revert.**

Recording the disposition for the file, since the merging agent explicitly noted this might need pulling back out.

This PR was merged at `5dbc44ec` while the coordinator was relaying an owner ruling that it should stop for personal review — the two crossed in transit, and the cause was an ambiguous "carry on" the coordinator had sent *before* the merging agent raised its banner. The merging agent flagged the tension itself, recorded that it was merging under the owner's by-name enumeration rather than under a judgement that the deprecation-matching change falls outside the owner-review filter, and named the revert target. That transparency is why this was a clean choice rather than a discovery.

The owner has since reviewed and elected to keep it. `git revert -m 1 5dbc44ec` was verified clean at the time of the decision — the v8 wire-format bump rides inside this merge, so it would have reverted to 7 along with the `deprecations` field, and nothing merged afterwards touches `src/driver/check.rs`, `wire_format.rs`, the deprecation code or `lib/do.eu`. The choice was unconstrained.

**Carried forward to the 0.14 release notes**, since both are upgrade-visible:
- Prelude deprecations reach blob mode for the first time (the table was empty in `run_type_checker_from_blob_core`, making every prelude deprecation inert in the shipped default configuration). With `--strict` promoting warnings to errors, a user writing `state.exec` or `random.exec` goes from passing to failing on upgrade.
- The known false positive: a top-level block of your own sharing a deprecated library namespace's name still matches, because `record_deprecation` does not record the binder identity of the declaration site.

PR #1075, opened 2026-07-26
PP — process parallelism: par-map + safe reductions (eu-u9xj.6)

gmorpheme:
> **Do not merge — owner review.** GC/memory change, `unsafe` (fork/mmap), and
> a new intrinsic + observable behaviour change — all three owner-review
> categories from CLAUDE.md. The owner reviews this personally.

## What this is

`par-map`/`par-sum`/`par-max`/`par-min`/`par-concat` (eu-u9xj.6): a new,
**highly experimental**, **macOS/Linux-only** vocabulary for mapping over
independent elements in parallel worker processes. Every combinator is
semantically identical to its sequential form (`xs par-map(f)` = `xs
map(f)`, etc.) — a pure performance advisory. Semantics may change without
notice ahead of stabilisation. Bundled: an independent complexity fix to
`max-of`/`min-of` (eu-pazes), found while sizing the PP fixtures — a
duplicating non-tail recursion (2^N) rewritten to a strict `foldl` (linear).

## Design

A POSIX copy-on-write `fork()`: above a size threshold, with W>1 workers,
inside a host declared fork-safe (the `eu` CLI; not the LSP/WASM/test
harness), `__PARMAP`'s own STG wrapper forces `xs`'s spine (a `SeqSpine`
combinator, compile-time — memoises a single-pass source correctly, unlike
forcing from inside the intrinsic at runtime), forks W workers who each
serialise their index chunk into a shared mmap arena, and reassembles in
index order — deterministic. Everywhere else (below threshold, one core,
**Windows**, any fork-path fault) it falls back transparently to the
identical sequential map, sharing the same codec so both paths give the
identical value and boundary error. `par-*` is strict; `f` must be pure and
return serialisable data. Design docs under
`docs/superpowers/{specs,plans,reports}/2026-07-25-pp-*`.

## Tested / verified

- Value-equivalence with `map`/`sum`/`max-of`/`min-of`/`concat`, incl. every
  serialisable type (numbers/strings/symbols/booleans/null/zoned-datetimes,
  standalone and nested), metadata-carrying results compared as rendered
  output; a streaming-import that is registered/partially-consumed does not
  disable `par-*`. Fault-injection verified per boundary-codec tag.
- Out-of-process: real COW-fork byte-identical to the sequential oracle on
  both engines under `EU_GC_VERIFY=2`; GC-stress mid-map is fault-injection
  verified (reverting the root-set fix reproduces SIGSEGV/panic 3/3).
- `fmt`/`clippy -D warnings`/full `cargo test` green; CI job
  `test-pp-parallelism` drives the fork path on Linux under `EU_GC_VERIFY=2`.
- Rebased onto master; fixtures renumbered into shared 222–227 range;
  prelude docs regenerated; diagnostic snapshots blessed (2 of 213, both
  fixtures new to this PR — a first-time snapshot, not a changed one).

## Known limitations

- Windows/spawn-based parallelism out of scope (eu-9udwg); worker-local
  reduction fusion deferred as a perf optimisation (pipes `par-map` into the
  exact sequential reducer for now).
- eu-gua64 (general inliner argument-duplication defect) is **not** fixed —
  only the two `max-of`/`min-of` victims are addressed here.
- eu-pkgoo (P1, filed from this PR, confirmed contained to the par-map
  boundary): an unused function calling `cal.zdt(...)` anywhere in a unit
  breaks an unrelated `par-map` returning `null`. The 222 fixture's zdt
  check uses a `t"…"` literal deliberately — don't "tidy" it back.
- Round-1 review (wicket) raised GC-handle-rooting and metadata-loss
  findings, both fixed; `force-spine` was later replaced by a wrapper-level
  `SeqSpine` STG combinator per owner feedback — see PR comments for both.


gmorpheme:
## Recorded review — wicket (independent, not the author)

**Verdict: REQUEST CHANGES.** This satisfies the CLAUDE.md "recorded review by
someone other than the author" prerequisite for the GC/memory/`unsafe`/engine
surface, but it is **not** an approval to merge, and the owner retains the merge
decision regardless.

Reviewed in an isolated worktree off `origin/master` at
`feat/furnace-pp-parallelism`. Everything below marked *verified* I reproduced
myself; I have separated that from what I took from the PR body.

### The good news first (verified, not taken on trust)

The COW-fork mechanism itself is sound. I built a 1500-element fixture whose `f`
returns blocks containing numbers, arithmetic and strings, ran it through the
real fork path on the **default bytecode engine and `EU_HEAPSYN=1`**, each under
`EU_GC_VERIFY=2`, and diffed against the sequential `map` oracle: **byte-identical
on both engines, zero GC panics, clean exit**. `par-sum` over N=2000 gives
`1999000` matching the oracle, faster than it. The arena's disjoint-segment
design, the `_exit`-only child discipline, the SIGINT reset and the
kill-and-reap-survivors logic all read correctly. The spike's R5 engine-equivalence
claim holds up.

I also independently repeated the declared fault injection: reversing the
`build_list` fold order (`serialise.rs:446`) makes `194_pp_par_equivalence` FAIL;
restoring it makes it PASS. The gate does gate.

### Blocking findings

**1. CI is red, and it is this PR's own gate test that fails.**
`194_pp_par_equivalence...FAIL` in eight jobs — ASAN, GC-verified, byte-dispatch,
HeapSyn, source-prelude, aarch64-sequential, and bytecode+blob on all three
platforms. This alone stops the PR. I reproduced it locally on macOS. Minimal
repro:

```
./target/release/eu test --allow-io tests/harness/079_streams.eu tests/harness/194_pp_par_equivalence.eu   # FAIL
./target/release/eu test --allow-io tests/harness/190_odkp_xml_entity_refs.eu tests/harness/194_pp_par_equivalence.eu   # PASS
```

**2. The R2 impure-producer boundary check is the root cause, and it is wrong in
three separate ways.** This is the substantive finding — fixing CI without fixing
the design would just hide it.

- *It fires before the fork decision.* The check sits at `driver.rs:77`, ahead of
  `decide_workers` at `:93`. On the sequential fallback there is no `fork()` and
  therefore no shared fd offset — the hazard R2 describes cannot arise — yet the
  error is still raised, and it pre-empts the fallback entirely. The PR body's
  "below threshold, W≤1, non-unix … falls back to a transparent sequential map"
  is not true while a producer is registered.
- *It keys on registration, not on being in flight.* Producers are registered at
  **import** time (`src/import/mod.rs:94-113`), during desugaring, whether or not
  the stream is ever touched. Verified — this four-line file fails, with no fork
  possible on a 2-element list:

  ```
  { import: ["jl=jsonl-stream@testdata/aux_import.jsonl"] }
  result: [1, 2] par-map(_ * 2)
  # error: par-map: cannot serialise a value with a live streaming import
  #        across the parallel boundary
  ```

- *The state is process-lifetime and leaks across evaluations.* `PRODUCER_TABLE`
  is a `thread_local!`, and `live_impure` is cleared only when a producer is
  *observed exhausted*. A partially-consumed stream — the normal case for a lazy
  stream — poisons every later `par-*` in that process. That is exactly the
  cross-file contamination in (1), and it will equally affect the LSP server and
  the WASM API, which evaluate many units per process. Note also `is_pure`
  defaults to `false` and nothing in the tree overrides it, so *every* streaming
  import is a permanent hazard.

**3. `par-max` / `par-min` cannot do useful work at any input size.** The PR body
frames the deferred spec §4 worker-local fusion as "a pure performance
optimisation … the reduce of N small values is cheap", and files eu-pazes
separately. Those two facts interact and the conclusion does not hold. I measured
`max-of` myself on literal lists:

| N | 16 | 18 | 20 | 22 |
|---|----|----|----|----|
| time | 0.13s | 0.38s | 1.35s | >90s (timeout) |

That is a clean 2^N (`max(l head, max-of(l tail))` re-evaluates the recursive
thunk in both the comparison and the result). `DEFAULT_THRESHOLD` is 1024, so:
below 1024 elements `par-max` never forks and gives no parallelism at all; at or
above 1024 the parent then runs `max-of` over N results, needing ~2^1024 steps.
Verified: a 30-element `par-max` with `EU_PP_THRESHOLD=8` does not terminate in
60s. There is no N at which `par-max` is both parallel and terminating.

Worker-local reduction is therefore not a performance nicety here — it is the
thing that keeps the reducer's input at W rather than N, and it is what makes the
day09-p1 `par-max` motivation real. Either land the §4 fusion, or drop
`par-max`/`par-min` from this PR until eu-pazes is fixed. Shipping combinators
that are "correct" only because both forms hang is worse than not shipping them.

**4. The WASM build is broken.** Five dead-code items on `wasm32` —
`DEFAULT_THRESHOLD`, `env_usize`, `decide_workers`, `chunk_bounds`, and the unused
`n` binding at `driver.rs:89`. The fork machinery is `#[cfg(unix)]` but its
helpers are not. Reproduced locally with `cargo build --target
wasm32-unknown-unknown --lib`.

**5. Prelude docs are stale** — the five new `par-*` entries. Run
`eu doc --prelude --output-dir docs/reference/prelude/`.

### Non-blocking, but please address

**6. Every combinator reports itself as `par-map`.** `parallel/intrinsic.rs:34`
hard-codes `"par-map"`, and the four reductions are prelude wrappers over
`__PARMAP`, so the `combinator` parameter threaded through the driver and
serialiser is effectively a constant. Verified: `[1,2,3] par-concat(mkfn)` reports
`par-map: cannot serialise a function …`. The PR body's "naming the value kind and
combinator" holds only for `par-map` itself.

**7. Harden `SegmentReader::next` (`arena.rs:192`).** It reads a record length out
of `MAP_SHARED` memory written by *another process* and hands it straight to
`slice::from_raw_parts`; the only bound is a `debug_assert!` in `read_slice`,
compiled out in release. The driver already treats worker output as potentially
corrupt ("corrupt read → sequential re-run"), so the parent should validate
`cursor + LEN_PREFIX + len <= segment_end` before dereferencing rather than after.
Cheap insurance on the one genuinely cross-process trust boundary in the PR.
`Arena::new`'s `segment_size * n_workers` is also unchecked, though the driver's
`ARENA_MAX` clamp keeps it unreachable today.

**8. The fork point's single-threadedness is assumed, not asserted.**
`fork.rs:71` documents it as a precondition, but nothing enforces it — and the
harness test had to be written around the fact that libtest violates it. This repo
ships an LSP server (`src/driver/lsp/`); a document evaluating `par-map` over
≥1024 elements inside a multi-threaded host would fork a multi-threaded process.
Please add a runtime guard that falls back to sequential unless the process is
known single-threaded, rather than relying on the threshold to make it unlikely.

**9. Inherited crash handler in children.** CLAUDE.md says the SIGSEGV/SIGBUS
handler installs unconditionally in `main()`; workers inherit it. A faulting
worker would run it in a forked child, where it may not be async-signal-safe.
Worth resetting alongside SIGINT.

`Smid::default()` at `fork.rs:156` is test-only, and `apply1_thunk`'s use matches
the existing `apply2_thunk`/`meta_value` precedent — neither is a concern.

### Summary

The hard part — COW fork, the shared arena, worker hygiene, engine neutrality —
is done well and I verified it works. What is not ready is the boundary check
(wrong placement, wrong lifetime, wrong trigger), the two reduction combinators
that cannot terminate at any parallel size, and three failing CI gates. I have not
merged and will not; per the coordinator the owner holds that decision.

*Verified independently:* CI failure reproduction + minimal repro, the unconsumed-
import repro, `max-of` scaling, `par-max` non-termination, `par-sum` correctness,
fork-path equivalence on both engines under `EU_GC_VERIFY=2`, the `build_list`
fault injection, the WASM dead-code errors, `par-concat` error naming.
*Taken from the PR body:* the `cargo test` / clippy / fmt results, and the
`test_pp_fork_path_equivalence_both_engines` fault injection.

gmorpheme:
## Review findings addressed — `c0bc36f1` (rebased onto current master)

Thanks — the review was right on every count, and two of the findings led
somewhere further than they looked. Point by point.

### 1 + 2. The R2 boundary check, redesigned (CI red, root cause)

You identified the wrong placement, wrong trigger and wrong lifetime
independently; they turned out to have a single answer. The check no longer
exists as a pre-emptive test at all.

A forked worker now **marks itself**, and while marked, `producer_next` /
`producer_drain` refuse to advance an *impure* producer — whose fd offset it
shares with the parent — and fail the worker. The driver already answers any
worker failure with a sequential re-run in the parent, where the producer is
consumed correctly, so the user simply gets `map` semantics. That fixes all
three of your objections at once:

- **placement** — the check exists only inside a `fork()`ed child, so the
  sequential fallback is untouched and genuinely transparent;
- **trigger** — it fires on a producer actually being *advanced*, not
  registered; an imported-but-untouched stream is no obstacle;
- **lifetime** — the flag only ever exists in a child that leaves via `_exit`,
  so nothing survives into a later evaluation. The LSP server and the WASM API
  are unaffected because they never fork in the first place (see 8).

The refusal happens *before* the producer advances, so the parent's offset is
untouched and the re-run is exact.

`is_pure` has lost its default: it is now a required trait method, and the
three import producers state `false` explicitly with the reason.

### 2a. A bug the regression test found: `par-*` consumed a stream destructively

Writing the fixture for the above turned up something worse, which was present
in the version you reviewed:

```
{ import: ["jl=jsonl-stream@testdata/aux_import.jsonl"] }
a: jl par-map(_.name)   # [Alice, Bob, Charlie]  — correct
c: jl count             # 1  — should be 3
```

The machine memoises a thunk by writing the result back into the **environment
slot** naming it (`vm.rs` `Atom`/`Ref::L(i)`: blackhole the slot, push
`Update`). `IntrinsicMachine::force` takes an already-resolved closure and has
no slot, so the spine `collect_spine` forced was never memoised: the next
reader re-forced the tail, met the exhausted producer, and got a one-element
list.

Fixed by forcing the spine in the **prelude**, through ordinary evaluation,
before `__PARMAP` sees it (`force-spine`, suppressed from export). The general
ABI hazard is filed as **eu-14ya9** with a design note — it is GC-subtle
(an updating force must root the owner across a sub-run that can evacuate), so
it wants designing, not patching.

### 3. `par-max` / `par-min`

Your diagnosis was correct and I can sharpen it: it is `2^N`, and the slow
direction is **ascending** (max at the tail — the comparison branch returns the
recursive thunk and re-enters it). Measured on the branch before the fix:
N=18 0.37s, N=20 1.29s, N=22 5.02s, N=24 >60s. Descending is the cheap
direction.

A second effect explains why measurements of this disagreed so violently across
reports: **the blob prelude and the source prelude differ in complexity class.**
Master's release binary does `range(0,20000) max-of` in 0.14s; the *same
binary* with `--source-prelude` needs 7s for N=22. Any branch that edits
`lib/prelude.eu` staleness-invalidates the blob, so it measures the slow path
while master measures the fast one. A user-level copy of the definition is
exponential in both modes, so the blob was masking the bug, not fixing it.
Filed as **eu-9wq0s**.

Rather than land the §4 fusion or drop the combinators, I fixed the reducer
(eu-pazes): `max-of`/`min-of` are now a strict left fold over the same
`max`/`min`, same semantics, same empty-list panic. `range(0,100000) max-of` is
0.34s. So `par-max` is now both parallel and terminating at every size:

```
$ EU_PP_TRACE=1 EU_GC_VERIFY=2 eu parmax.eu     # xs = range(0, 20000)
par-max: 20000 elements — forked 13 workers
par: 399960002   seq: 399960002   agree: true   # identical on EU_HEAPSYN=1 too
```

Indicative timing (single run, not a PROTOCOL.md measurement): a 4000-element
`par-max` over a non-trivial `f` takes 0.19s wall / 1.16s user against 1.42s
sequential.

I have corrected the **eu-pazes** title and description back to "exponential"
and recorded the measurements — the "quadratic and order-dependent" reading
relayed to me does not reproduce.

### 4. WASM

`cargo build --target wasm32-unknown-unknown --lib` is clean. `DEFAULT_THRESHOLD`,
`env_usize`, `decide_workers` and `chunk_bounds` (and its test module) are
`#[cfg(unix)]`; the stray `n` moved inside the `cfg(unix)` block.

### 5. Prelude docs

Regenerated.

### 6. Combinator naming

`__PARMAP` now takes the combinator as a leading **symbol** argument
(`__PARMAP(:par-concat, f, xs)`), strict, so each wrapper names itself.
`[1,2,3] par-concat(mkfn)` reports `par-concat: cannot serialise a function
across the parallel boundary`. Gated by
`tests/harness/errors/196_pp_concat_non_serialisable.eu`.

### 7. Arena hardening

`SegmentReader` now validates, in release as well as debug, that the length
prefix fits the segment and that the payload it claims fits too — before
forming a slice — and stops at the first inconsistency. The record count (also
cross-process data) is clamped to what the segment could physically hold.
`Arena::new` uses `checked_mul`. Three unit tests scribble corrupt counts and
lengths into a live segment; removing the bound check makes two of them fail.

### 8. Fork-point single-threadedness — now asserted, and it mattered

I implemented the runtime probe you asked for and it immediately revealed that
**the `eu` binary is not single-threaded**: `pti_threadnum` is 3 (the
large-stack evaluation worker `main` spawns, the original thread parked in
`join`, and the `ctrlc` crate's signal watcher). A strict "count == 1" guard
would therefore have silently disabled the entire feature — and the fork test
would still have passed, because the design makes the fork unobservable in the
output.

So the guard is an explicit **host declaration** instead, defaulting to *do not
fork*: `declare_fork_safe_host()` is called from `run()` in `src/bin/eu.rs`
past the LSP branch, where the process owner can vouch that the other threads
are parked. The LSP server, the WASM API, embedders and libtest never call it
and never fork; `par-*` is simply sequential there. The thread count *is*
recorded at declaration and re-checked at the fork point, so a host that
declares and then spawns threads loses the privilege.

Two consequences worth calling out:

- `EU_PP_TRACE=1` now reports the path each `par-*` took
  (`forked 4 workers` / `sequential (below threshold)` / `sequential (process
  is not a declared fork-safe host)`).
- `test_pp_fork_path_equivalence_both_engines` now **asserts it forked**.
  Without that it was a tautology — a never-forking build passes by matching
  the oracle it never left. This is what caught the above.

### 9. Inherited crash handler

Children reset `SIGSEGV`/`SIGBUS` to `SIG_DFL` alongside `SIGINT`.

---

## Validation

- `cargo fmt --all` ✅ · `cargo clippy --all-targets -- -D warnings` ✅ ·
  `cargo test` (full, blob-less) **all green** ✅ ·
  `cargo build --target wasm32-unknown-unknown --lib` clean ✅
- Fork path re-verified under `EU_GC_VERIFY=2` **and** `EU_GC_STRESS=1` on the
  default engine and `EU_HEAPSYN=1`: byte-identical to the sequential oracle,
  zero GC panics, clean exit — and the trace confirms it really forked.
- Blob mode (`cargo xtask prelude-compile` then `cargo test`) is green except
  `test_193_1tkk_7_12_curated_trace` and two `diagnostics_invariants` tests,
  which fail identically on `origin/master` — pre-existing (eu-7x0r, eu-rqwh),
  and the CI blob job is deliberately scoped around them.

### Fault injection (every new gate, broken → FAIL, restored → PASS)

| Gate | Injection | Result |
|---|---|---|
| `test_195_pazes_max_of_linear` (complexity) | `max-of` back to the `cond` shape | FAILED at 180s deadline → restored → ok |
| `195_pazes_max_of_linear` (semantics) | `max-of` folds `min` | FAIL → restored → PASS |
| `196_pp_stream_import_boundary` | drop `force-spine` from `par-map` | FAIL → restored → PASS |
| `196_pp_stream_import_boundary` | reinstate the registration-keyed pre-emptive check | FAIL → restored → PASS |
| `test_error_196_pp_concat_non_serialisable` | hard-code the combinator to `"par-map"` | FAILED → restored → ok |
| `test_pp_fork_path_equivalence_both_engines` | `process_is_fork_safe()` → `false` | FAILED ("expected the COW-fork path…") → restored → ok |
| `arena::tests` (2 of 3) | remove the record-length bound check | FAILED → restored → ok |

The previously declared `build_list` injection still holds.

> Still **do not merge** — owner review.

gmorpheme:
Follow-ups pushed after the review-fix commit:

- `1ff16853` — the new "it actually forked" assertion required a fork on **Windows**, which has no fork path at all (every `par-*` there is sequential by construction), so a correct build failed it. Now `cfg(unix)`-gated; the equivalence assertions still run everywhere.
- `0f045887` — documented `EU_PP_TRACE` / `EU_PP_THRESHOLD` / `EU_PP_WORKERS` / `EU_PP_ASSUME_SINGLE_THREADED` in `docs/development/debugging.md`.

**Full CI rollup was green on `1ff16853`** — all 21 required checks pass, including AddressSanitizer, GC-verified, byte-dispatch, HeapSyn, source-prelude, aarch64-sequential, bytecode+blob on all three platforms, the dedicated Process-parallelism job, WASM, prelude-docs freshness, and both Test Suites. The eight jobs the review found red are all green. (Re-running now for the docs commit.)

### Beads filed from this pass

- **eu-14ya9** (P2) — `IntrinsicMachine::force` cannot memoise, so any intrinsic that forces a lazy spine loses thunk updates. `force-spine` fixes `par-*`; the general hazard needs an updating force in the ABI, and that is GC-subtle (the owner must be rooted across a sub-run that can evacuate), so it wants designing.
- **eu-9wq0s** (P2) — the blob prelude and the source prelude differ in **complexity class** for the same definition. This is why measurements of `max-of` disagreed so violently between reports: any branch editing `lib/prelude.eu` invalidates the blob and silently switches to the slow path.
- **eu-wpswc** (P2) — separately from all of the above, `xs map(f) sum` is **quadratic**: N=5000 0.56s, N=10000 2.33s, N=20000 9.73s (×4.2 per doubling), in both blob and source mode. `xs sum` and `xs map(f) count` are both linear at the same N, so it is the fold-over-map combination. This is the commonest shape in the language and exactly the shape PP targets, so it is worth a look before 0.14.

Still **do not merge** — owner review.

gmorpheme:
## eu-pazes: complexity settled on a clean build, and the gate now gates

Three contradictory complexity claims were on record for `max-of`/`min-of`. All
three are reconciled by one variable — **blob vs source prelude** — and it is
not the interposed `map`.

Method: fresh worktree at master `a1af1f0b`, clean `cargo build --release`,
measured in **both** prelude configurations on the same machine, every `eu`
wrapped in `timeout` with `--heap-limit-mib 8192`.

### Before, master `a1af1f0b`, superseded `max-of`

`range(1, N) max-of`, source-prelude build:

| N | 16 | 18 | 20 | 22 | 24 |
|---|---|---|---|---|---|
| ascending | 124 ms | 240 ms | 699 ms | 2568 ms | >90 s |

×3.9 per +2 in N — **2^N**. Descending is linear (86 / 101 / 168 ms at
N = 1000 / 5000 / 20000). Adding `map((_0 * _0))` changes nothing
(123 / 239 / 717 / 2655 ms, >90 s at 24).

Same source, blob build: ascending 47 / 72 / **153 ms** at N = 1000 / 5000 /
20000 — **linear**.

### Mechanism

`beta_reduce` (`src/core/inline/reduce.rs:220`) substitutes arguments into an
inlinable callee with no occurrence or duplication guard. `max(l, r): if(l > r,
l, r)` uses `r` twice, so `max(l head, max-of(l tail))` reduces to two copies of
the recursive call — two thunks, no sharing, T(n) = 2·T(n−1). `eu dump inlined`
prints it:

```
max-of = λ(l).
COND([CLAUSE(NILP(l), PANIC("max of empty list")), CLAUSE(NILP(TAIL(l)), HEAD(l)),
      IF(GT(HEAD(l), max-of(TAIL(l))), HEAD(l), max-of(TAIL(l)))]);
```

Ascending input returns the second copy, so both are forced (2^n); descending
returns the head and forces the recursion only for the comparison (linear). In
blob mode the prelude combinator is an opaque baked global that is never
inlined, hence linear there.

It is **not prelude-specific**. Pure user code is 2^N in blob mode too —
61 / 122 / 361 / 1326 / 5211 ms for N = 16…24 — and the same algorithm with the
recursive result bound to a name is 118 ms at N = 20000:

```eu,notest
mymax(a, b): if(a > b, a, b)
mymax-of(l): if(l tail nil?, l head, mymax(l head, mymax-of(l tail)))
```

Filed as **eu-gua64** (P1, core inliner). This PR's `foldl` rewrite only removes
the two victims that live in the prelude: `foldl` passes the accumulator as a
plain variable, which cannot be duplicated.

### After, this branch

`range(1, N) max-of` ascending: source-prelude 84 / 93 / **136 ms** at
N = 1000 / 5000 / 20000 (before: no result at N = 1000 within 90 s); blob
48 / 63 / 131 ms (unchanged).

`par-max` motivation restored — blob mode, N = 20000:
`range(1,20000) par-max((_0 * _0))` returns 399960001 in **0.16 s** forking 13
workers, against **9.56 s** for the sequential `map((_0 * _0)) max-of`.

### Correcting my own gate

`test_195_pazes_max_of_linear` **could not fail** as first written: with the
superseded definition restored and rebuilt, the fixture still passed in 1.1 s,
because blob mode never inlines the prelude. It now runs the fixture under
`--source-prelude`, which is where the 2^N lives.

**Fault-injection verified** after the change: superseded definition restored
**and rebuilt** — the prelude source is embedded at build time, so editing
`lib/prelude.eu` without a rebuild changes nothing, which is what masked the
first attempt — the fixture does not complete within the 180 s deadline and the
test panics with the eu-pazes message; restored — PASS in 1.13 s.

### Not the cause, measured separately

* **eu-14ya9** (`IntrinsicMachine::force` cannot memoise) — intrinsic-only; none
  of these expressions goes through an intrinsic that forces a spine, and the
  duplication is visible in the core dump before the machine runs.
* **eu-wpswc / eu-pghae** — a genuinely separate **quadratic**, isolated cleanly
  here: `reverse max-of` 182 ms at N = 20000 vs `reverse map(sq) max-of`
  8953 ms; `map(sq) sum` 9517 ms vs plain `sum` 121 ms; `map` alone and
  `reverse` alone both linear. Untouched by this PR — `max-map`/`min-map` stay
  quadratic until eu-wpswc is fixed, which is why their coverage in the fixture
  is deliberately restricted to 3-element lists.

### Gates

`cargo fmt --all`, `cargo clippy --all-targets -- -D warnings` clean.
`cargo test --no-fail-fast`: harness 511 passed / 1 failed, lib 1374 passed / 0
failed; the only failures are the three known pre-existing blob-mode ones
(`test_193_1tkk_7_12_curated_trace`, `debug_trace_restores_the_uncurated_trace`,
`secondary_labels_do_not_excerpt_library_internals` — eu-7x0r / eu-rqwh), which
fail identically on the PR base and on `origin/master`. Full CI rollup on
`eebc421a`: **20 pass, 0 fail** (5 Release/prepare-release jobs skipping, as
expected for a non-tag build).

gmorpheme:
### Two amendments to the measurement comment above, and a scope note for review

**1. Confidence label on the `par-max` figure.** The 60x figure in
[the comment above](https://github.com/curvelogic/eucalypt/pull/1075#issuecomment-5086092395)
— blob mode, `range(1, 20000) par-max((_0 * _0))` = 399960001 in **0.16 s**
forking 13 workers, vs **9.56 s** for sequential `map((_0 * _0)) max-of` — is
**measured-single** per `docs/superpowers/engine-ab/PROTOCOL.md`: one run, one
developer machine, not interleaved, not a quiet-machine session. It is
evidence that the day09-p1 motivation is restored, **not** a release claim and
not a number to quote in CHANGELOG or ROADMAP. The quiet-machine session is
eu-1hcw and owner-only.

The same label applies to every timing in that comment. They are
order-of-magnitude complexity-class evidence — the gaps being distinguished are
2^N versus linear, so machine noise cannot flip the conclusion — but none of
them is protocol-grade.

**2. This PR does not close eu-gua64.** The PR body now says so explicitly.
Summary for the reviewer: the prelude rewrite here *sidesteps* the inliner
defect (a `foldl` accumulator is a plain variable, so `max`'s two uses of its
second parameter become two references to one shared thunk) but does not
repair it. `beta_reduce` at `src/core/inline/reduce.rs:242` still substitutes
arguments with no occurrence count, and pure user code with the same shape is
still 2^n **in blob mode**. eu-gua64 is filed P1 against core, quill is
dispatched on it, and it needs its own blast-radius review because fixing it
changes core shape broadly. Please do not treat merging this as closing it.

Likewise eu-wpswc / eu-pghae (the separate quadratic in `map(f)` piped into a
fold) and eu-14ya9 (intrinsic thunk-update memoisation) are untouched — three
distinct root causes, not one wearing four hats.

**Unchanged from the previous report:** `BYTECODE_WIRE_FORMAT_VERSION` is not
touched by this branch (verified against `origin/master`: the diff contains no
reference to it in `build.rs` or `xtask/src/main.rs`). Both regression tests
remain fault-injection verified as described above. No code changed in this
round — the head commit is still `eebc421a`; only the PR body and the bead
records were updated.

gmorpheme:
## Recorded review — PR #1075 (wicket)

**Verdict: changes required.** Two blocking findings, both reproduced on this
branch at `eebc421a`. This is the recorded review the CLAUDE.md
"GC/memory, `unsafe`, engine internals" rule requires; it is **not** an
approval, and I am not merging — the owner holds the merge decision.

The design is good. The spec is followed closely, the deviations are disclosed
honestly, the arena layout (count-header written *last*, so a worker dying
mid-write reads back as an empty segment) is genuinely crash-safe, the
`declare_fork_safe_host` opt-in is the right shape, and the boundary-error and
producer-purity work is careful. The problems below are not in the fork
machinery — they are in the code that runs on **both** paths.

### What I verified, and how

Isolated worktree at `eebc421a`, release build, macOS/aarch64. Independent
fault injection on both gates. I did not re-check CI (confirmed green by the
owner). Timings below were taken tonight on a contended machine and are
**indicative only** — none is a release-grade figure and none of my findings
turns on a wall-clock number; where a number matters I have used a *ratio*,
which is contention-robust.

---

# Part A — process parallelism (eu-u9xj.6)

## A1 — BLOCKING: heap handles are held across `force()`, so the GC invalidates them. Reachable panic.

This is the headline finding and it is not theoretical.

The repository already has an explicit, written rule for this. `src/eval/stg/list.rs:470-474`:

```
// Forcing only the n+1 cells we visit restores the lazy invariant ...
// No handle is held across a force — each `force` consumes the handle it is
// given — so it is GC-safe.
```

and the two places that *cannot* obey that rule use the stash root set instead
— `src/driver/io_run.rs:618-632`:

```
// SAFETY: each call to evaluate_to_whnf_for_io runs machine.run() which
// ends with collect::collect.  If evacuation occurs, any SynClosure values
// held on the Rust stack ... become dangling — the GC does not know about them.
// We convert raw_fields into an ordered Vec and stash ALL closures upfront so
// the GC can trace and update their heap pointers during each sub-evaluation.
```

`src/driver/bytecode_io_run.rs:72-79` says the same for `BcValue`. The stash is
a real root set: declared at `src/eval/machine/vm.rs:306`, scanned at `:1600`,
pointer-updated after evacuation at `:1627`.

**The PP driver obeys neither rule.** It holds large collections of heap
handles across many `force()` calls:

- `src/eval/stg/parallel/driver.rs:140-174` — `collect_spine` accumulates every
  element head into `out` (`:160`) while calling `machine.force(cur)` on each
  tail.
- `src/eval/stg/parallel/driver.rs:196-209` — `sequential_map` accumulates
  `results` (`:205`) across N forces, and holds `elements` and `f` throughout.
- `src/eval/stg/parallel/driver.rs:247-268` — the worker loop, same shape.
- `src/eval/stg/parallel/serialise.rs:272-292` — `serialise_list` holds `cur`
  across `machine.force(head)` **and** across the recursive `serialise_value`,
  then dereferences it at `:286`. `serialise_block` (`:315-343`) is the same.

`force()` runs the machine, and the machine collects mid-run:
`src/eval/machine/vm.rs:1737` (HeapSyn) and `src/eval/bytecode/machine.rs:525-526`
(bytecode). `enter_whnf_subrun` / `evaluate_to_whnf_impl` stash only
`state.current` and the continuation stack — nothing a BIF is holding on the
Rust stack.

There is also a structural reason this happened, which is worth fixing
deliberately rather than patching: **`IntrinsicMachine` exposes no root set at
all.** `stash_push` is an inherent method on `Machine` (`vm.rs:2358`), not on
the neutral ABI, so an intrinsic written against `&mut dyn IntrinsicMachine`
*cannot* protect a handle even if it wants to. Any fix needs a neutral
root-set method on the trait (implemented over `state.stash` on both engines),
not just local re-ordering in the driver.

### Reproduction

```
cat > /tmp/cpar.eu <<'EOF'
f(x): range(0, 100) map((_0 + x)) sum
xs: range(1, 250)
r: xs par-map(f)
n: r sum
EOF
EU_GC_STRESS=1 EU_PP_THRESHOLD=100000 RUST_BACKTRACE=1 \
  ./target/release/eu --heap-limit-mib 32 /tmp/cpar.eu -x yaml
```

```
thread '<unnamed>' panicked at src/eval/bytecode/machine.rs:2400:29:
index out of bounds: the len is 8585 but the index is 2250422912
stack backtrace:
   2: core::panicking::panic_bounds_check
   3: eucalypt::eval::bytecode::machine::handle_op_predecoded
   4: <BcBifContext as IntrinsicMachine>::force
   5: eucalypt::eval::stg::parallel::driver::par_map
   6: <ParMap as StgIntrinsic>::execute
...
thread 'main' panicked at src/bin/eu.rs:43:10: main thread panicked
EXIT=101
```

`2250422912` is a garbage code ordinal read out of a closure the collector
moved or reclaimed while the driver held it. **3/3 runs panic**, on both the
sequential path (`EU_PP_THRESHOLD=100000`) and the fork path
(`EU_PP_THRESHOLD=1 EU_PP_WORKERS=4`, which panics in the workers *and* again
in the parent's fallback, and separately at `src/eval/memory/array.rs:404`
"bounds error"). A 2000-element variant at `--heap-limit-mib 64` also produced
`tried to call a number as a function` — i.e. the same corruption sometimes
surfaces as a **silent wrong-shaped value** rather than a crash.

**Control:** the identical workload with `map` instead of `par-map`, same env,
same heap limit, 3/3 runs clean and correct (`n: 4345050`). The defect is in
the new code, not in `EU_GC_STRESS`.

Re-verified after the machine's disk pressure was cleared; the fault-injected
build compiled successfully in both cases, so this is not a build artefact.

### Why the PR's validation did not catch it

- `EU_GC_VERIFY=2` verifies reachability *from roots*. These handles are not
  roots and the collector does not know they exist, so verification is
  structurally incapable of detecting this. The `test-pp-parallelism` CI job
  gives no coverage here.
- The fixture (`tests/harness/testdata/pp_fork_fixture.eu`) is 16 elements with
  a trivial `f`. No collection occurs during the map, so the window never
  opens.
- The spike's phase B/D forced GC *inside* the machine, but did not hold
  intrinsic-side handles across the forcing, so it could not have found this.

**Required:** a neutral root-set on `IntrinsicMachine`; every handle held
across a `force()` in `driver.rs` and `serialise.rs` protected by it; and a
harness test that actually collects during a `par-map` (the repro above,
scaled to run in CI, under `EU_GC_STRESS=1`) so the gate exists.

## A2 — BLOCKING: `par-map` silently discards metadata, changing rendered output

`par-map` is documented as "identical to `xs map(f)`". It is not, and the
divergence is **silent** — no error, and the value compares equal.

```
$ cat /tmp/m3.eu
id(x): x
b: { ` :suppress hidden: 1
     v: 2 }
sb: [b] map(id)
pb: [b] par-map(id)

$ ./target/release/eu /tmp/m3.eu -x json
{"b":{"v":2},"pb":[{"hidden":1,"v":2}],"sb":[{"v":2}]}
```

`map` honours `:suppress`; `par-map` **emits the suppressed key**. Identical on
the fork path (`EU_PP_THRESHOLD=1 EU_PP_WORKERS=2`). In a data-templating tool
whose users mark internal or sensitive fields `:suppress`, a combinator
advertised as a drop-in performance advisory that re-exposes them is a serious
correctness problem.

Cause: the codec has no `Meta` case. `serialise_value`
(`serialise.rs:144-218`) dispatches on `data_tag`, which only matches
`HeapSyn::Cons`; `serialise_block` reads the key with
`field_native(view, &pair, 0)` (`:328-331`), taking the raw native and
discarding any wrapper. `deserialise_value` rebuilds a bare `BlockPair`.

Why the tests cannot see it:

- `t-map-block` in `194_pp_par_equivalence.eu` uses `mkrec(x): { n: x sq: x * x }`
  — no metadata.
- Every equivalence assertion is written with `=`, and **`=` ignores
  metadata**: in my repro `bsame: sb = pb` reports `true` while the rendered
  output differs. So the whole equivalence suite is blind to this class by
  construction.
- `test_pp_fork_path_equivalence_both_engines` compares the fork path against
  the *sequential par-map* path, not against `map`. Since both paths share the
  codec, no codec defect can ever fail that test. The doc comment calls the
  no-env run "the sequential oracle"; it is not an oracle for the codec.

**Required:** either preserve metadata across the boundary, or refuse it with
`NotSerialisable` (loud, and consistent with the §6 philosophy) — silently
dropping it is the one option that should be off the table. Plus: at least one
equivalence assertion that compares **rendered output** rather than `=`, and a
metadata case in `194`.

## A3 — Major: worker panics leak raw Rust panic text to the user's stderr

`run_workers` (`fork.rs:184-191`) wraps the worker in `catch_unwind`, but the
default panic hook has already written `thread '<unnamed>' panicked at …` to
the inherited stderr by the time `catch_unwind` returns. The parent then falls
back to sequential and may well succeed — so a *successful* run can print Rust
panic spew. My A1 fork-path repro shows exactly this. Set
`panic::set_hook(Box::new(|_| {}))` in the child immediately after
`reset_signals_to_default()`.

## A4 — Major: `waitpid` is not `EINTR`-safe; a signal leaves a live, unreaped worker

`fork.rs:206`: `libc::waitpid(pid, &mut status, 0)` with no `EINTR` retry. The
`eu` CLI has a Ctrl-C watcher, so signal delivery during a long `par-map` is
expected, not exotic. On `EINTR`, `r != pid`, so the parent records
`WaitFailed` and calls `kill_and_reap(&pids[idx..])` (`:211`) — which
**excludes the worker that just EINTR'd** (`idx` was already incremented at
`:203`). That worker is neither killed nor reaped: it keeps burning a core to
the end of its chunk and then becomes a zombie, while the parent redundantly
re-runs the whole map sequentially. Retry on `EINTR`, and include the current
pid in the kill set on a genuine wait failure.

## A5 — Moderate: only the *total* record count is validated, not per worker

`driver.rs:292` checks `results.len() != n`. The parent knows each worker's
exact chunk length (`chunk_bounds`), so it should check each segment yields
exactly that many records. As written, a short read in one segment compensated
by a long read in another passes the check and silently reassembles results in
the **wrong order** — the one thing the design promises cannot happen. The
check is free; the arena's own unit test (`arena.rs:313-330`) already notes
that a corrupt count can yield spurious empty records, which is precisely the
input shape that would exercise this.

## A6 — Moderate: the user-facing docs overclaim

`lib/prelude.eu` and `docs/reference/prelude/io.md` both say "identical to
`xs map(f)`" with no qualification. `par-*` is spine-strict and
element-strict, which is a real observable difference:

```
g(x): if(x = 3, panic("boom"), x)
[1,2,3] map(g) head        # => 1
[1,2,3] par-map(g) head    # => error: panic: boom
```

The spec (§3) is explicit that `par-*` is strict; the docs are not. A new
user-facing language feature with a purity contract, a strictness contract, a
serialisable-subset restriction and a boundary error gets one table row and no
guide page. Please qualify the doc strings and add a short guide section.

## A7 — Minor

- **The gating spike report is not in the repo.** The module header
  (`parallel/mod.rs:6-7`), the CI job comment and the PR body all cite
  `docs/superpowers/reports/2026-07-25-pp-fork-safety-spike.md`. It exists only
  on the unmerged `spike/pp-fork-safety` branch, so on master these are dangling
  references to the document that authorises the whole design. Please include it.
- `fork.rs:145` — `EU_PP_ASSUME_SINGLE_THREADED=1` bypasses the fork-safety
  gate in *any* host, including the LSP server. Documented as diagnostics-only;
  consider also requiring a debug build, or at least a stderr warning.
- `driver.rs:239` — `clamp(w * (16 + 64), ARENA_MAX)` panics if `min > max`.
  Only reachable with absurd `EU_PP_WORKERS`, but `clamp` on computed bounds is
  worth replacing with an explicit `min`/`max`.
- `arena.rs:227` — the `checked_add` failure path returns `None` via `?` without
  setting `remaining = 0`, unlike the two bounds checks below it. Unreachable,
  but inconsistent.
- **Orphan workers.** If the parent is killed with anything other than a
  process-group signal, workers run to completion unattended. Ctrl-C is handled
  correctly (children reset SIGINT to `SIG_DFL`); a bare `kill -TERM` to the
  parent is not.
- `PER_ELEM_CAP` is per *input* element, but `par-concat`'s results are lists
  whose size is unrelated to the input count. Overflow degrades to a silent
  sequential fallback, which is safe but means `par-concat` can quietly never
  parallelise.
- Spec §8's "≥1.5× wall on a real data-parallel case" success criterion has no
  test and no gate; the only evidence is the single measurement in the PR body.
  Correctly labelled measured-single there, but worth noting it is unenforced.

## Answers to the specific points raised

1. **Fork and the managed heap.** The fork point is sound. Fork happens from
   the evaluation thread inside a BIF, with the machine quiescent, and the
   `declare_fork_safe_host` / `process_is_fork_safe` opt-in (recording a thread
   baseline and withdrawing if threads appear) is a better answer than
   "single-threaded" and is correctly wired only in `src/bin/eu.rs:65-72`. A
   collection in the child is fine — it operates on private COW pages. **But**
   see A1: the problem is not the fork, it is that the heap handles the driver
   holds are invisible to the collector on *both* sides.
2. **The shared arena.** Layout is sound: disjoint per-worker segments, no
   atomics, count-header written last in `finish()` so a partial write reads
   back as an empty segment, and the reader bounds every untrusted field
   (`arena.rs:222-242`, with good unit tests). No pointers cross. Two caveats:
   the parent's reads are ordinary non-atomic reads of memory another process
   wrote, relying on `waitpid` for ordering — true in practice on every real
   kernel, but it is an assumption outside the Rust memory model and deserves a
   comment; and see A5 on per-worker validation.
3. **Failure modes.** fork fails → clean sequential fallback. Worker non-zero
   exit / signal → fallback. Arena overflow → fallback. Corrupt record →
   fallback. Non-serialisable → proper `ExecutionError::NotSerialisable`, exit
   1, good message and location. All correct. The gaps are A3 (panic spew), A4
   (EINTR zombie) and A7 (orphans).
4. **Determinism.** The *mechanism* is deterministic (index-ordered
   reassembly). The *guarantee* is not currently tested against `map`: the
   fork test's "oracle" is the sequential `par-map` path sharing the same
   codec, and every assertion uses `=`, which ignores metadata. A2 is the
   concrete failure that slipped through. `194` does compare against `map` and
   does gate correctly (I confirmed every `t-` binding feeds `RESULT` in 194,
   195 and 196), but only over metadata-free values.
5. **Non-Unix.** Confirmed a genuine fallback, not a parallel implementation.
   `driver.rs:97-125` is one `#[cfg(unix)]` block that either returns early or
   falls through to the same `sequential_map` every other path uses. There is
   exactly one map implementation, so Windows cannot diverge. Good.
6. **`unsafe`.** I found **no unsound `unsafe`**. All of it is in `arena.rs`
   and `fork.rs`; the mmap/munmap, the bounds-checked raw accessors, the
   post-fork `signal` calls and `_exit` are all correctly reasoned and the
   safety comments are accurate. A1 is a *safe-code* memory bug — stale
   GC-managed pointers behind a safe API — which is arguably worse, because no
   `unsafe` block flags it for review.
7. **Engine parity.** Structurally good: the driver is written entirely against
   the neutral ABI and the new `apply1_thunk` has both a HeapSyn default and a
   bytecode override, and the fork test exercises both engines. A1 reproduces
   on the bytecode engine; the equivalent code path exists on HeapSyn
   (`vm.rs:1737`) and I see no reason it is exempt.

---

# Part B — `max-of`/`min-of` over `foldl` (eu-pazes)

**This part is good.** Reviewed on its own merits; I would take it as-is apart
from one stale comment.

## B1 — The gate genuinely gates (independently fault-injection verified)

I did not take the author's word for it. In my worktree, with no `git stash`
involved:

| Step | Result |
|---|---|
| Baseline (as submitted) | `test_195_pazes_max_of_linear ... ok` (11.8 s / 15.2 s, debug, 180 s deadline) |
| Old `cond`-recursive `max-of`/`min-of` restored, rebuilt | **FAILED** — deadline expiry with the intended message: *"max-of/min-of did not reduce a 20000-element list within 180s — they are no longer linear in list length (eu-pazes)"* |
| Restored, rebuilt | `ok` again |

The fault-injected build **compiled successfully** (`Finished test profile in
3.25s`) before the test ran, so this was a test failure, not a disk-pressure
build artefact. Tree left clean; `lib/prelude.blob` is untracked and I removed
the one I generated.

**I also verified the load-bearing `--source-prelude` claim**, which is the
unusual part of this gate. With the old definition restored and a blob
generated:

- blob mode, N = 20000: completes, ~3.4 s — **linear**, would pass the gate
- source-prelude mode: N = 14/16/18/20 → 0.64 / 1.21 / 3.40 / 12.09 s

The author is right: without `--source-prelude` the test could not fail. The
flag is necessary, and the coverage constraint being recorded on eu-5yn9v is
the correct follow-up.

## B2 — Complexity class: the PR is right; the eu-pazes bead is not

`eu-pazes` carries a coordinator comment stating *"the complexity class in this
bead and in the PR #1075 review is WRONG. max-of/min-of are NOT exponential —
they are QUADRATIC and ORDER-DEPENDENT."* On the source-prelude path that
correction is itself wrong, and the PR's text is accurate. Ratios are immune to
tonight's machine contention: 0.64 → 1.21 → 3.40 → 12.09 s for N = 14 → 20 is
≈1.9× **per element** (~19× over six elements); quadratic would be ≈2×
over that range. That is exponential.

The most likely explanation is that the correction measured blob mode, where
(as I confirmed above) the same definition is linear, or measured the separate
`map`-into-fold quadratic. The PR body already distinguishes all three defects
carefully. **Please reconcile the eu-pazes bead comment** so the record does not
contradict the merged code's rationale.

## B3 — Framing relative to eu-gua64: correct as written

The PR explicitly states eu-gua64 is not fixed, names `beta_reduce`
(`src/core/inline/reduce.rs`), notes it still hits pure user code in blob mode,
says quill is dispatched, and says in terms "Merging this PR must not close
eu-gua64." The 195 fixture header and the test doc comment repeat it. The bead
is OPEN, P1, assigned to quill, and linked from eu-pazes as DISCOVERED. I see
nothing that could lead a reader to conclude the inliner defect is resolved.

One fix required: `tests/harness/testdata/pp_fork_fixture.eu:11-14` still says
*"the prelude's `max-of` is (pre-existing, orthogonal to PP) exponential in list
length, so a larger N would make the debug `cargo test` binary … intractably
slow"*. That is now stale — this PR fixes it — and it is the justification given
for the fixture's size, so it should be updated or the fixture resized.

---

## Summary

| # | Severity | Finding |
|---|---|---|
| A1 | **Blocking** | Heap handles held across `force()`; reachable panic (exit 101) and silent corruption. Needs a root set on `IntrinsicMachine`. |
| A2 | **Blocking** | `par-map` silently drops metadata; `:suppress` keys reappear in output. Tests structurally cannot detect it. |
| A3 | Major | Worker panic text leaks to user stderr on otherwise-successful runs. |
| A4 | Major | `waitpid` not `EINTR`-safe; leaves a running, unreaped worker. |
| A5 | Moderate | Per-worker record counts unvalidated; wrong-order reassembly possible. |
| A6 | Moderate | Docs claim unqualified equivalence with `map`; `par-*` is strict. |
| A7 | Minor | Spike report absent from the repo; several small robustness items. |
| B1–B3 | — | eu-pazes: gate verified to gate, framing correct, one stale comment. |

Part B could ship on its own today. Part A needs A1 and A2 resolved before it
is safe to merge; A3–A6 should go in the same round.

Happy to re-review promptly.

gmorpheme:
## Review round: A1–A7 addressed (head `f1983941`)

Thank you — this review found two real defects and, more usefully, named the
reason my own validation could not have found either. The point about oracles
that share an implementation with the thing under test is the one I have tried
hardest to act on rather than acknowledge; it changed the tests below and it
also turned up a *third* structural blind spot, described under A1(b).

---

### A1 — heap handles held across `force()` — **fixed structurally**

**Decision: extend the neutral ABI with a root set.** Not restructuring the
call sites.

Restructuring cannot work here, and it is worth being precise about why: the
driver's whole job is to accumulate N results while forcing N thunks. There is
no ordering of `sequential_map` in which no handle is live across a `force()` —
the accumulator *is* the live handle. `serialise_list`/`serialise_block` are
the same: the spine cursor must survive the recursive serialisation of the
element, which forces. Any local reordering would have been a narrowing of the
window, not a fix, and would have left the next intrinsic to rediscover it.

So `IntrinsicMachine` now carries the root set:

```rust
fn gc_root_len(&self) -> usize;
fn gc_root_push(&mut self, closure: AbiClosure) -> usize;
fn gc_root_get(&self, idx: usize) -> AbiClosure;
fn gc_root_set(&mut self, idx: usize, closure: AbiClosure);
fn gc_root_truncate(&mut self, len: usize);
```

Implemented over each engine's existing stash — `MachineState`,
`MachineBifContext` and `BcBifContext`, i.e. the same `Vec` that `io_run.rs`
and `bytecode_io_run.rs` already rely on and that is already scanned at
`vm.rs:1600` and pointer-updated at `:1627`. No new root machinery, no new GC
surface: the gap was purely that `stash_push` was **inherent** on the machine
types, so an intrinsic holding `&mut dyn IntrinsicMachine` could not reach it,
exactly as you said.

The protocol is stated once, in the trait: push a handle, then **read it back**
after every force — the collector updates the entry in the root set, not the
copy on the Rust stack. `parallel::roots::with_roots` scopes each frame so it
is released on the error path too, and nesting is safe because a `force()`
pushes and pops its own entries.

Every site you listed now uses it: `collect_spine` (`out` and the cursor),
`sequential_map` (`results`, `elements`, `f`), the worker loop, `serialise_list`
and `serialise_block` (cursor, pair, and the value across the recursive call).
`deserialise_value` and `build_list` deliberately do **not** root anything, and
that is load-bearing rather than an oversight: allocation never collects in
this heap — `policy_requires_collection` is consulted only in the run loops —
so `force()` is the only hazard, and saying so in one place is more useful than
rooting defensively everywhere.

Your repro on this branch, 3/3 each, plus the `map` control:

| | before | after |
|---|---|---|
| sequential path (`EU_PP_THRESHOLD=100000`) | **SIGSEGV** 3/3 | 3/3 correct |
| fork path (`EU_PP_THRESHOLD=1 EU_PP_WORKERS=4`) | **SIGSEGV** 3/3 | 3/3 correct |
| `map` control | clean | clean |

(I saw SIGSEGV where you saw the bounds panic; same corruption, different
landing site.)

#### A1(b) — the gap your review implies but does not name

While fault-injecting the fix I found that **the fork path cannot be gated by
comparing output at all**, for a reason that generalises your oracle argument:
a worker fault is caught by the parent, which re-runs the whole map
sequentially. A worker that dies of a stale handle therefore produces the
*correct* answer and exit 0. I verified this the hard way — with the root set
removed from the worker loop, my new GC test still **passed**.

So `EU_PP_STRICT=1` is new: it makes a fork-path fault an error instead of a
fallback. Diagnostics only; production wants the fallback. With it, the same
injection fails the test immediately. Documented in `debugging.md`, and the
test comment records that the flag is load-bearing rather than decorative.

### A2 — metadata silently discarded — **fixed by preserving it**

Preserving, not refusing. `par-map` promises to be identical to `map`, `map`
keeps the annotation, and refusing would break `par-map` over any block with a
`:suppress` key — a normal thing to write.

The cause was upstream of the codec: `force()` *strips* metadata that nothing
consumes (`return_meta`, both engines, the empty-stack branch), so by the time
`serialise_value` saw the value the annotation was already gone. Both engines
now record it at that one point, `take_stripped_meta()` recovers it
immediately after the force, and it crosses as wire tag `0x09` (metadata value
then body), rebuilt via the existing `meta_value` ABI. Nested and computed
metadata are covered because the hook is at the strip site, not a syntactic
peek at the closure.

Your repro, all three configurations (bytecode sequential, bytecode fork,
HeapSyn) now render `pb` identically to `sb`, `:suppress` honoured.

One consequence, deliberate and documented: metadata must be data too, so
metadata holding a function is now a boundary error where `map` would have
carried it unevaluated. That is a real divergence and it is the honest one —
the alternative is dropping the annotation silently, which is precisely the
defect. New error fixture `197_pp_non_serialisable_metadata`, guide section,
and the doc string says so.

### A3–A6 — fixed

- **A3** — the child installs a silent panic hook immediately after
  `reset_signals_to_default()`, as you suggested. A run that succeeds via the
  fallback no longer prints Rust panic text.
- **A4** — `waitpid` retries on `EINTR` (a shared helper, also used by
  `kill_and_reap`). On a genuine wait failure the kill set is now
  `&pids[worker..]`, including the pid whose wait failed — your point that
  `idx` had already advanced.
- **A5** — each worker's segment is checked against the chunk length the
  parent assigned it, not just the total, with the trace line naming the
  mismatch. It is defensive: I have no way to fault-inject a corrupt count
  through the public surface, so it is not separately gated, and I would
  rather say that than imply coverage I do not have.
- **A6** — `par-*` doc strings now state the strictness, purity and
  serialisable-data contracts; `docs/reference/prelude/io.md` regenerated (the
  freshness job covers it); new guide page `docs/guide/parallelism.md`, in
  `SUMMARY.md`, with the `[1,2,3] map(g) head` versus `par-map(g) head` example
  from your review.

### A7 — fixed, with one exception

- The spike report is now in the tree at
  `docs/superpowers/reports/2026-07-25-pp-fork-safety-spike.md`, so the module
  header, CI comment and PR body no longer dangle. (No ```eu fences, so it does
  not touch the doctest job.)
- `EU_PP_ASSUME_SINGLE_THREADED=1` warns once to stderr when it takes effect,
  and `debugging.md` says it applies in hosts that deliberately never declared
  themselves.
- `clamp` on computed bounds replaced with an explicit floor/ceiling.
- `arena.rs` `checked_add` failure now sets `remaining = 0` like its
  neighbours.
- The arena's cross-process visibility assumption — non-atomic reads ordered by
  `waitpid` rather than by the Rust memory model — is written into the module
  header, with why it holds (disjoint, single-writer segments; no worker reads
  another's bytes, the parent reads none before reaping).
- `PER_ELEM_CAP` now documents that it is a per-*input*-element budget and
  therefore only a heuristic for `par-concat`, which can quietly never
  parallelise; `EU_PP_TRACE` at least makes that visible.
- Orphan workers under a non-process-group kill are documented in the module
  header rather than fixed — the fix is a process group or a pdeathsig, both of
  which change signal semantics for the whole CLI, and that is not a change to
  make inside a review round.
- **Not addressed:** spec §8's "≥1.5× wall on a real data-parallel case" has no
  gate, and I have not added one. A wall-clock threshold in CI would be a test
  that fails for reasons unrelated to the code — and no wall-clock figure taken
  in the last day is defensible anyway. That belongs to the quiet-machine
  session (eu-1hcw), and it stays unenforced until then.

---

### Tests — built to be capable of failing

Taking the point about gates that cannot fail seriously, each new check is
paired with the injection that breaks it.

**`test_pp_gc_collects_during_par_map`** — a map with collection happening
*during* it: a heap limit crossed partway through (this heap collects only at
its limit) plus `EU_GC_STRESS=1`, so every collection **evacuates** rather than
merely marks and a stale handle is guaranteed stale rather than merely at risk.
Compared against a `map` oracle, not another driver run. Sequential path plus
the fork path on both engines under `EU_PP_STRICT=1`.

The sequential configuration runs on the default engine only: the driver and
serialiser are one implementation against the neutral ABI, HeapSyn is covered
by the fork configuration, and running the sequential shape under HeapSyn as
well costs about two minutes of debug-build time for no additional code
coverage. I would rather state that than have it look like an oversight.

**Oracles changed from `=` to rendered output.** `194_pp_par_equivalence.eu`
and `pp_fork_fixture.eu` now compare `render-as(:json)` against `map`. Your
diagnosis was exact: `=` ignores metadata, so `bsame: sb = pb` was `true` while
the output differed, and the fork test's "oracle" shared the codec with the
thing under test. `pp_fork_fixture.eu` now carries its own `map-ok` and
`meta-ok` bindings and the Rust test asserts them, so the fork test finally
gates the codec rather than only the fork machinery.

**Fault injection, rebuilding each time** (the rebuild matters — I learned that
the hard way on eu-pazes):

| injection | result | restored |
|---|---|---|
| sequential-path root set removed | `test_pp_gc_collects_during_par_map` **FAILS**, exit 101, `array.rs:347` bounds panic | passes |
| worker-loop root set removed | **FAILS** under `EU_PP_STRICT=1` — and *passes* without it, which is why the flag exists | passes |
| metadata capture removed | `test_194_pp_par_equivalence` **FAILS**; fork test **FAILS** with `meta-ok: false` | both pass |

### Gates

`cargo fmt --all`, `cargo clippy --workspace --all-targets -- -D warnings`, and
`cargo test --no-fail-fast` — 1369 lib, 514 harness, every integration target,
**0 failures**. `BYTECODE_WIRE_FORMAT_VERSION` untouched (the new `0x09` tag is
in the PP arena's private same-process format, not the blob).

Full CI rollup on `f1983941` reported separately once it completes. Not merged.

gmorpheme:
### Follow-up on `3fe0b44e`: the GC gate's CI cost

The GC-safety gate in `f1983941` was green but expensive — `Process-parallelism (Linux, both engines)` went from **1m49s to 16m16s**, all of it in the test step. That is a cost I introduced and it should not stand, so it is fixed rather than left for someone to notice later.

The fixture is deliberately collection-heavy: once this heap reaches its limit *every* poll collects, so cost grows steeply with the work done after the crossing, and the job's inherited `EU_GC_VERIFY=2` multiplied it again.

Three changes, each **re-verified by fault injection** rather than assumed safe:

1. **The sequential shape shrinks** from 60 elements × 100 to 30 × 60 at `--heap-limit-mib 8`. The collection still lands mid-map; there is simply less work afterwards. Removing the sequential root set is still detected 3/3.
2. **The fork path keeps the larger shape, in its own fixture.** It splits the work W ways, so each worker needs enough allocation *within its own chunk* to reach the limit. On the smaller shape no worker collects and removing the worker-loop root set goes **undetected** — I checked, which is why there are two fixtures rather than one. This is the same failure mode as the `--source-prelude` correction on part B: a gate that runs and proves nothing.
3. **`EU_GC_VERIFY` is removed from this test's spawns.** It checks reachability from roots, and the entire bug class here is handles that were *never* roots, so it cannot contribute coverage — it only re-traverses the heap after every one of a deliberately large number of marks. The fork path still runs under `EU_GC_VERIFY=2` in that job via `test_pp_fork_path_equivalence_both_engines`, so nothing is lost. The reasoning is in the test, not just here.

Local wall time for the test: **53s → 11.8s**. Both injections still fail it; both restores still pass.

For the record, `f1983941` was fully green before this — 20 pass, 0 fail, 5 release jobs skipping — so this is a cost fix, not a correctness one. Rollup for `3fe0b44e` follows.

gmorpheme:
### CI rollup on `3fe0b44e` — 20 pass, 0 fail

```
AddressSanitizer                           pass  16m30s
Byte-dispatch harness (soak baseline)      pass   2m55s
Bytecode + blob harness (macos-latest)     pass   4m38s
Bytecode + blob harness (ubuntu-24.04-arm) pass   3m8s
Bytecode + blob harness (ubuntu-latest)    pass   3m46s
Corpus parse equivalence (eu-wrf7)         pass     33s
Diagnostics blob-mode integration          pass   1m46s
GC-verified harness (x86-64)               pass   2m25s
HeapSyn harness (baseline)                 pass   3m8s
Lint                                       pass   1m19s
Prelude blob determinism                   pass   1m21s
Prelude docs freshness                     pass   1m53s
Process-parallelism (Linux, both engines)  pass   4m52s
Security Audit                             pass   3m9s
Sequential harness (aarch64 release)       pass   2m33s
Source-prelude harness (x86-64)            pass   2m57s
Source-prelude tick-parity tripwire        pass   1m49s
Test Suite (ubuntu-latest)                 pass  17m44s
Test Suite (windows-latest)                pass  28m13s
WASM Compilation Check                     pass   4m37s
Release Windows / MacOS / Linux x86_64 / Linux aarch64 / prepare-release   skipping
```

`Process-parallelism` **16m16s → 4m52s**, against 1m49s before this PR added the GC gate. Roughly three minutes of CI for a gate on a class of memory corruption that `EU_GC_VERIFY` cannot see, which seems a fair trade; if you would rather it were cheaper still, the sequential shape can shrink further and I will re-verify the injection at whatever it lands on.

The intermediate head `f1983941` was also fully green (20 pass, 0 fail) — the follow-up commit is a cost fix, not a correctness one.

Re-review requested. Not merged, and I will not merge it.

gmorpheme:
## Recorded re-review — PR #1075 at `3fe0b44e` (wicket)

**Verdict: both blocking findings closed. No objection from me.** Merge remains
the owner's call; I am not merging.

I re-executed rather than re-read, on the same machine and the same
configurations as the original review, in an isolated worktree at `3fe0b44e`.
Everything below is an observed result.

---

## A1 — reachable memory corruption: **CLOSED**

The exact reproduction from the first review, unchanged:

```
f(x): range(0, 100) map((_0 + x)) sum
xs: range(1, 250)
r: xs par-map(f)
n: r sum
```

| Configuration | Before (`eebc421a`) | Now (`3fe0b44e`) |
|---|---|---|
| Sequential path, `EU_GC_STRESS=1`, 32 MiB, ×3 | **3/3 panic**, exit 101 | **3/3 `n: 4345050`** |
| Fork path, `EU_PP_WORKERS=4`, `EU_PP_STRICT=1`, ×3 | 3/3 panic (workers *and* parent) | **3/3 `n: 4345050`** |
| `map` control, identical workload | correct | correct — **identical value** |

And the larger shape under everything at once —
`EU_GC_VERIFY=2` + `EU_GC_POISON=1` + `EU_GC_STRESS=1`:

| | result |
|---|---|
| sequential | `n: 934230` |
| fork, bytecode | `n: 934230` |
| fork, HeapSyn | `n: 934230` |

Zero verification failures, zero poison hits, zero panics in parent or workers.

### Assessment of the rooting mechanism

I scrutinised this rather than accepting that the panic stopped, because
"registered somewhere that happens to survive" would look identical from the
outside. It holds up:

- **It is the existing root set, not a new one.** `gc_root_*` is implemented
  over each engine's `stash` — declared `vm.rs:306`, scanned `vm.rs:1600`,
  pointer-updated after evacuation `vm.rs:1627`; the same set `io_run.rs` has
  used for this exact purpose. Nothing new had to be made GC-correct.
- **The protocol is read-back, not hold.** Every site pushes, then re-reads
  with `gc_root_get` *after* each force (`driver.rs:223-226`, `290-291`,
  `356-357`; `serialise.rs:186-188`, `335-359`, `394-429`). The collector
  updates the slot, not the Rust-stack copy, so relocation is picked up. This
  is the property that matters and it is applied consistently — I checked every
  handle in `driver.rs` and `serialise.rs`.
- **Does it survive a collection during `force()` specifically** — the window
  my repro exploited? Yes: that is exactly what the `EU_GC_STRESS=1` runs
  above exercise, and `collect_spine` is careful enough to move the tail into
  the cursor slot *before* rooting the head (`driver.rs:239-243`), so the two
  are never both unrooted across a force.
- **What unregisters them:** `roots::with_roots` records `gc_root_len()` on
  entry and truncates on every exit path including `Err`. Nesting is safe
  because `force()` pushes and pops its own entry, so indices handed out by an
  outer frame stay valid. The one path that skips the truncate is a Rust
  `panic!` unwinding past it — in a worker that is caught by `catch_unwind`,
  and in the parent a panic is already fatal, so it is not a live gap.
- `stripped_meta` is a new live handle and has correctly been added to both
  `scan` and `scan_and_update` (`vm.rs:1655-1658`, `1685-1687`).

**On the verifier question you asked:** now that these are roots,
`EU_GC_VERIFY` does traverse them, and the runs above are clean. But it is
worth being precise about what that does *not* buy: verification checks
reachability from roots, so the original bug class — a handle that was never a
root — remains structurally invisible to it. If a future intrinsic forgets to
root something, the verifier will be silent again. The real safety net is the
GC-stress-plus-`map`-oracle test, not `EU_GC_VERIFY`. The author states this
plainly in the code and did not overclaim it.

### One residual assumption (not blocking)

`sequential_map` and `try_parallel` read all handles out of the root set and
then call `build_list`, relying on allocation never collecting. That is true
today — collection happens only in the machine run loop — and the comments say
so explicitly. It is an invariant a future "collect on allocation pressure"
change would break silently. Worth a note wherever that policy lives.

---

## A2 — silent metadata loss: **CLOSED**

Same repro:

```
b: { ` :suppress hidden: 1
     v: 2 }
sb: [b] map(id)
pb: [b] par-map(id)
```

| Path | Before | Now |
|---|---|---|
| sequential | `pb:[{"hidden":1,"v":2}]` | **`pb:[{"v":2}]`** |
| fork, bytecode | `pb:[{"hidden":1,"v":2}]` | **`pb:[{"v":2}]`** |
| fork, HeapSyn | — | **`pb:[{"v":2}]`** |

`pb` now matches `sb` exactly. I also probed mis-attribution, which the
"record the metadata the last force stripped" design could plausibly get wrong:
a map alternating metadata-carrying and metadata-free results across a 3-worker
fork returns the correct per-element answer on both paths, no cross-contamination.

**Do the new assertions actually compare metadata?** Yes — verified by
injection, not by reading. They compare `render-as(:json)` rather than `=`
(`194_pp_par_equivalence.eu:55-64`, `pp_fork_fixture.eu:33-35`), plus a
positive `t-meta-hidden` asserting the suppressed key is absent from the
rendered text. With the metadata capture removed from
`serialise.rs:189`, **both** `test_194_pp_par_equivalence` and
`test_pp_fork_path_equivalence_both_engines` FAIL, and the fork fixture reports
the precisely discriminating pair:

```
map-ok:  true      <- metadata-free blocks unaffected
meta-ok: false     <- the metadata assertion is what catches it
```

Restored: both pass. The assertions gate on the property, not on a coincidence.

---

## The cheapened GC gate — **still gates; both halves**

This was the right thing to be suspicious of, so I fault-injected the reduced
test rather than reasoning about the fixture sizes.

| Injection | Result |
|---|---|
| Sequential root set removed (results accumulate on the Rust stack) | **FAILED in 3.9 s** — `pp_gc_stress_fixture.eu`, UB check: `ptr::offset requires the address calculation to not overflow` |
| Worker-loop root set removed (handles hoisted out of the loop) | **FAILED in 37 s** — `pp_gc_fork_fixture.eu`, `worker 0 failed` under `EU_PP_STRICT=1` |
| Both restored | **passes, 12.1 s** |

The 16×-cheaper gate detects both halves. The cost came out of the fixture
shape and the removed `EU_GC_VERIFY`, not out of the detection.

**`EU_PP_STRICT=1` is genuinely load-bearing, and I confirmed it directly.**
With the worker-loop injection still in place and the flag removed, the fork
fixture prints `n: 469050` — *the correct answer*, matching the oracle —
because the worker dies, the parent silently re-runs sequentially, and the
output is right. Without the flag that half of the gate cannot fail. The author
identified this class themselves and built the switch to defeat it; that is the
opposite of optimising a gate into uselessness, and it also retires the
"worker-confined defects are untestable" concern from my first review.

The stated reason for dropping `EU_GC_VERIFY` from these spawns is correct on
the merits, as set out under A1: it cannot see this bug class, and the fork
path still runs under `EU_GC_VERIFY=2` in that job via
`test_pp_fork_path_equivalence_both_engines`.

---

## A3–A7: addressed

| | |
|---|---|
| **A3** | `std::panic::set_hook(Box::new(\|_\| {}))` in the child (`fork.rs:88`) — no panic spew on runs that succeed via fallback. Confirmed: my fork-path runs are clean. |
| **A4** | `waitpid_eintr_safe` retries on `EINTR` (`fork.rs:229-240`), and the wait-failure kill set is now `&pids[worker..]` — it includes the un-waited pid, which was the actual leak. |
| **A5** | Per-worker record count checked against its assigned chunk length (`driver.rs:390-417`), with the reasoning recorded. |
| **A6** | Prelude doc strings now say STRICT and state the purity and data-only contracts; new `docs/guide/parallelism.md` (102 lines). The strictness divergence itself is unchanged and correct per spec §3 — it is now documented rather than contradicted. |
| **A7** | The gating spike report is in the tree (`docs/superpowers/reports/2026-07-25-pp-fork-safety-spike.md`); `clamp` replaced with explicit floor/ceiling (`driver.rs:336-337`); `PER_ELEM_CAP`'s `par-concat` limitation documented; fork-safety override warns. |

## Part B (eu-pazes)

The stale comment is fixed — no "exponential" claim remains in
`pp_fork_fixture.eu`, and the replacement gives the real reason for the
fixture's size. My earlier verification of the eu-pazes gate stands unchanged.
The eu-gua64 framing is unchanged and still correct.

The one item still outstanding is **not in this PR**: the `eu-pazes` bead
carries a coordinator comment asserting the complexity class is quadratic. My
measurements on the source-prelude path (≈1.9× per element, N=14→20) say
exponential, matching the PR. That bead comment should be reconciled with the
record; it does not block this merge.

---

## Suite state at `3fe0b44e`

`cargo test --test harness_test pp` — 19 passed, 0 failed. `cargo fmt --all
--check` clean.

Both blockers closed by re-execution, both new gates verified to gate by my own
fault injection, all secondary findings addressed. Good work on the
`EU_PP_STRICT` insight in particular — that is a real improvement to the
testability of the whole fork path, not just a fix for what I raised.

gmorpheme:
Do not merge — owner review.

Held at the owner's instruction pending two open items:

1. `force-spine(xs): if((xs count) >= 0, xs, xs)` — under review for replacement by a wrapper-level force using the existing `SeqList`/`SeqNumList` machinery in `src/eval/stg/force.rs`, following the pattern `src/eval/stg/array.rs` already uses. Investigation in progress; not yet decided.
2. Test coverage review, requested by the owner.

Also outstanding on the branch: rewrite `max-of`/`min-of` to use `then` rather than `if` per `docs/eucalypt-style.md:17`, and correct the stale "eu-gua64, still open" comment in `tests/harness/223_pazes_max_of_linear.eu` (eu-gua64 is closed, fixed by #1086).

This hold binds every agent regardless of CI or review state, and is released only by the owner.

gmorpheme:
## Test coverage review (owner-requested, review-only — no merge; owner hold in effect)

Reviewed in isolation at `/tmp/eu-wicket-review`, head `56bb9dcc`. Assessed against forking-runtime failure modes, not line-count targets, and independently fault-injection-verified the two load-bearing regression tests rather than taking the PR description's word for it.

### Fault-injection results (independently reproduced)

- **`test_pp_gc_collects_during_par_map` genuinely gates the A1 root-set fix.** I broke the "always reread the handle from the root set after a force" discipline in both `sequential_map` and `try_parallel`'s reassembly loop (holding results in a plain `Vec<AbiClosure>` across further `force`/`deserialise_value` calls instead of via `gc_root_push`). Rebuilt: the test FAILed with a genuine UB abort (`ptr::offset` overflow, subprocess killed) rather than a soft mismatch. Reverted, rebuilt: PASSes again. This corroborates the PR's own "3/3" fault-injection claim.
- **`test_pp_fork_path_equivalence_both_engines` genuinely discriminates fork vs. sequential fallback.** I forced `decide_workers` to always return 1 (never fork). Rebuilt: the test FAILed on its own `"expected the COW-fork path"` assertion, not silently passing on the fallback answer. Reverted, rebuilt: PASSes.

Both of these are real gates, not decorative ones — `roots.rs` having zero unit tests is not, on its own, a coverage gap; the mechanism is exercised (and shown to be load-bearing) at the integration level.

### Real gaps, ranked

1. **No test — unit or integration — exercises boolean, null, symbol, or zoned-datetime values crossing the `par-map` boundary**, despite `serialise.rs`'s own module doc listing all five as the supported "serialisable-data subset" alongside numbers/strings/lists/blocks (which *are* covered). `TAG_TRUE`/`TAG_FALSE`/`TAG_NULL`/`TAG_SYM`/`TAG_ZDT` have encode and decode arms with literally no test touching them. This is the single cheapest, most concrete gap to close — e.g. extend `222_pp_par_equivalence.eu` with a predicate-returning `par-map` (booleans), one returning `null`, one returning a bare symbol, and one returning a `zdt`, compared the same way the existing checks are.
2. **Arena-overflow → sequential-fallback has no end-to-end test.** `arena.rs` unit-tests the raw `Arena` API's overflow behaviour in isolation (`overflow_is_reported_not_corrupting`), but nothing drives `try_parallel`'s own catch of that overflow (the documented `PER_ELEM_CAP`/`par-concat` weakness called out in `driver.rs`'s own comments) and confirms the driver falls back correctly rather than corrupting or hanging.
3. **The EINTR-safe `waitpid` retry loop (the A4 finding from the previous review round) has no test that actually delivers a signal mid-wait.** `worker_panic_is_contained` only exercises the caught-panic→exit(42) route; nothing proves the `EINTR` continue-loop in `waitpid_eintr_safe` actually retries rather than misreading the error. Hard to make fully deterministic, but not infeasible (e.g. `SIGALRM` fired at the test thread during a deliberately slow worker) — worth at least one attempt given this project's history of "fixed but untested" regressions.
4. Lower priority: a worker dying by raw signal (vs. a caught panic) only reaches `exited_cleanly`'s `WIFEXITED` check by inspection, not by test; the default `DEFAULT_THRESHOLD` (1024) boundary is never exercised since every test overrides `EU_PP_THRESHOLD=1`.

### Verdict

The owner's instinct is right, but narrowly: the raw per-file unit-test counts are misleading (the two counts singled out — `driver.rs`'s 1 and `roots.rs`'s 0 — are compensated for by integration tests that I confirmed actually gate), so this isn't "thin everywhere." But item 1 above is a real, currently-silent gap in a value-boundary codec, which is exactly where an unnoticed regression would land, and it's cheap to close. Recommend closing #1 before merge; #2/#3 are worth beads if not addressed now.

PR #1077, opened 2026-07-26
fix(desugar): diagnose, don't panic, on a CoreError from an imported unit (eu-mqlkv)

gmorpheme:
Fixes **eu-mqlkv** (P1).

## The bug

An invalid embedding in an **imported** `.eu` file aborted the process with a panic and **exit 101**, while the *same* file evaluated directly produced a clean codespan diagnostic and **exit 1**. That asymmetry was the tell.

```
# badlib.eu
` { type: s"any → [{{x: number}}]" }
f(v): [{ x: 1 }]

# main.eu
{ import: "badlib.eu" }
r: f(1)
```

```
$ eu badlib.eu                       # direct — fine
error: invalid embedding - invalid type annotation: ...
  ┌─ badlib.eu:1:11
exit 1

$ eu -L . main.eu                    # imported — panic
thread '<unnamed>' panicked at src/core/desugar/rowan_ast.rs:2993:22:
failure translating import: InvalidEmbedding(...)
thread 'main' panicked at src/bin/eu.rs:43:10: main thread panicked
exit 101
```

(The trigger I hit was `{{` brace-escaping, which is correct inside a plain `type:` string but wrong inside an `s"…"`, where the content is already a type expression.)

## Root cause

Three `.expect("failure translating import")` sites in `src/core/desugar/rowan_ast.rs` unwrapped the result of `translate_import`:

| Line | Context |
|---|---|
| 2642 | `rowan_declaration_to_binding` — scoped `` ` { import: … } `` on a declaration |
| 2829 | `impl Desugarable for rowan_ast::Block` |
| 2993 | `impl Desugarable for rowan_ast::Unit` |

So **any** `CoreError` raised while translating an import — not just an invalid embedding — killed the process rather than flowing to the caller's codespan renderer.

All three enclosing functions already return `Result<_, CoreError>`. The `Unit` impl returns `Err(CoreError::VersionRequirementFailed(..))` **nine lines above** the panic it replaced, which makes the `.expect` clearly unintentional rather than a deliberate invariant.

## The fix

`.expect(…)` → `?`, at all three sites, with a comment recording why.

## Tests

Two error-harness cases, deliberately covering **two different `CoreError` variants** so the fix is gated as the general one it is, not as a fix for the case that surfaced it:

- `tests/harness/errors/194_mqlkv_import_bad_embedding.eu` — `CoreError::InvalidEmbedding` from an imported unit's malformed `s"…"` annotation. `.expect` sidecar: `exit: 1`, stderr matches `invalid embedding`.
- `tests/harness/errors/195_mqlkv_import_bad_requires.eu` — `CoreError::VersionRequirementFailed` from an imported unit's unsatisfiable `requires: ">=999.0"`. `.expect` sidecar: `exit: 1`, stderr matches `does not satisfy requires`.

Both assert **exit 1** (not 101) and a real diagnostic on stderr, which is exactly the property that regressed.

## Fault injection — done

Restored all three `.expect("failure translating import")` calls, rebuilt, ran the suite:

```
test test_194_mqlkv_import_bad_embedding ... FAILED
test test_195_mqlkv_import_bad_requires ... FAILED
test result: FAILED. 4 passed; 2 failed
```

Reapplied the fix:

```
test test_195_mqlkv_import_bad_requires ... ok
test test_194_mqlkv_import_bad_embedding ... ok
test result: ok. 6 passed; 0 failed
```

Both tests genuinely gate the change.

## Gates

- `cargo fmt --all` — clean
- `cargo test` — full suite green (1353 lib + 508 harness + 111 typecheck + the rest, 0 failures)

## Notes for review

- Found while implementing SV3 (eu-u9xj.1). Deliberately kept **out** of the SV3 PR, which touches the `t-*` projection / blob wire surface and needs focused review.
- No behaviour change for programs that were working: this converts an abort into the diagnostic the same error already produced on the non-import path.


gmorpheme:
## Review — Wicket (gatekeeper), independent verification

**Verdict: APPROVE-WITH-COMMENTS.** The diagnosis is right, the fix is minimal and correct, the
diagnostics it unblocks are of good quality, and nothing regresses. One substantive finding below
(test coverage gates only one of the three fixed sites) that I'd like addressed before merge, plus
two minor points. I did **not** merge — the owner retains the merge decision this session.

Reviewed in an isolated worktree at `origin/fix/quill-import-embedding-panic` (`00d2ad02`, current
with `master` at `a1af1f0b`).

---

### CI

`gh pr checks 1077`: **20 pass, 4 `skipping`** (the four `Release *`/`prepare-release` jobs). No
failures, nothing pending. Confirmed independently.

### Gates re-run locally on the branch

| Gate | Result |
|---|---|
| `cargo fmt --all --check` | clean |
| `cargo clippy --all-targets -- -D warnings` | clean |
| `cargo test` (full) | green — 1353 lib + 508 harness + 111 typecheck + the rest, **0 failures** |

Matches the PR body's claims.

### Fault injection — independently repeated, not taken on trust

I restored all three `.expect("failure translating import")` calls myself, rebuilt, and ran the
suite:

```
test test_194_mqlkv_import_bad_embedding ... FAIL: exit code mismatch: expected 1, got 101
test test_195_mqlkv_import_bad_requires  ... FAIL: exit code mismatch: expected 1, got 101
test result: FAILED. 0 passed; 2 failed
```

Restored the fix → both `ok`. Both tests genuinely gate, and they gate on **exit 1 vs 101**
specifically, which is the property that regressed. I also confirmed the harness mechanism has real
teeth: `ErrorExpectation::validate` (`src/core/analyse/testplan.rs:59`) compares the exit code and
regex-matches stderr, and `run_error_test` asserts on the result — so neither sidecar can pass
vacuously. Both `.expect` sidecars are present in the diff (a *missing* sidecar returns `Ok(None)`
→ `Ok(0)` → silent pass, so this mattered to check).

### All three sites are genuinely user-reachable — verified individually

The lead asked whether each of the three is on a live path. I built one reproducer per site and ran
each against the **reverted** binary, confirming a distinct panic line each time, then against the
fixed binary:

| Site | Form exercised | Reverted binary | Fixed binary |
|---|---|---|---|
| `2993` `Unit` impl | `{ import: "bad.eu" }` at unit level | panic at `rowan_ast.rs:2993:22`, **exit 101** | codespan diagnostic, **exit 1** |
| `2642` `rowan_declaration_to_binding` | `` ` { import: "bad.eu" } `` on a declaration (documented in `docs/guide/imports-and-modules.md` §Scoped Imports) | panic at `rowan_ast.rs:2642:18`, **exit 101** | codespan diagnostic, **exit 1** |
| `2829` `Block` impl | leading block metadata inside a nested block — `inner: { { import: "bad.eu" } y: f(1) }` | panic at `rowan_ast.rs:2829:22`, **exit 101** | codespan diagnostic, **exit 1** |

So all three `.expect`s were live user-facing panics, not defensive invariants, and the fix repairs
all three. `grep` confirms these are the *only* three call sites of `translate_import` in the tree,
and no other `.expect`/`unwrap` in `rowan_ast.rs` converts a `CoreError` into a panic on this path.

### Blast radius — non-annotation paths verified

The change widens from one `CoreError` variant to all of them, so I exercised the neighbours. All
produce a diagnostic naming the *imported* file at the right line — not the importing file's
`import:` line, which would have been the confusing outcome:

| Case | Result |
|---|---|
| `VersionRequirementFailed` via import | exit 1, points at the imported file, with `= note` hints |
| `InvalidVersionConstraint` via import (a **third** variant, untested by the PR) | exit 1, correct location |
| Transitive import, fault two levels down | exit 1, correct location (`deep.eu:1:11`) |
| Named import (`lib=deep.eu`) | exit 1, correct location |
| Missing import file | exit 1, `could not read 'nope.eu'` — unchanged |
| Circular import | exit 1, `detected cycle involving …` — unchanged |
| Healthy import | exit 0, correct output — no regression |

---

## Finding 1 (please address before merge) — the tests gate only one of the three fixed sites

Both new fixtures use a **unit-level** import, so they exercise site `2993` only. Sites `2642` and
`2829` have no regression coverage.

Demonstrated, not inferred: I reverted *only* `2642` and `2829` to `.expect`, left `2993` fixed, and
ran the suite —

```
test test_194_mqlkv_import_bad_embedding ... ok
test test_195_mqlkv_import_bad_requires  ... ok
test result: ok. 2 passed; 0 failed
```

— while `declscope.eu` and the nested-block reproducer above still aborted with **exit 101**. Two
thirds of this fix can be reverted with CI fully green and users still hitting the panic. For a P1
panic fix that is the coverage that matters most.

Two more fixtures reusing the existing malformed-annotation helper would close it cheaply — one with
`` ` { import: … } `` scoped to a declaration, one with leading block metadata in a nested block. I'd
also suggest the scoped-import case is worth having on its own merits: it is the form the guide
documents most prominently.

## Finding 2 (minor, convention) — import helpers belong in `testdata/`

`194_mqlkv_bad_annotation.eu` and `195_mqlkv_impossible_requires.eu` sit directly in
`tests/harness/errors/`. Every existing import fixture in that directory puts its helper under
`tests/harness/errors/testdata/` (`154_internal_import_error.eu` → `testdata/154_lib.eu`, plus the
YAML/JSON/JSONL fixtures). Following that keeps non-test `.eu` files out of the directory that error
tests are enumerated from, and out of reach of any future sidecar-completeness sweep. In-file
comments naming the importer are a nice touch either way.

## Finding 3 (observation, no change requested) — desugarer state is not unwound on the new error path

`translate_import` pushes `self.file` and increments `self.named_import_depth` before
`source.content().desugar(self)?`; the new `?` skips the matching pop/decrement. Harmless today — I
checked every `Desugarer::new` call site (`src/driver/source.rs:763`, `src/import/yaml.rs`,
`src/wasm_pipeline.rs`) and each builds a fresh local `Desugarer` that is dropped when the error
propagates, so no corrupted state can survive into a later translation. Worth knowing if anything
ever moves to multi-error recovery that keeps desugaring after a `CoreError`.

## Follow-up worth a bead — `eu check` still can't gate this (pre-existing, not caused by this PR)

The direct/imported asymmetry that motivated this PR survives under `eu check`:

```
eu check deep.eu     # fault present directly  → exit 1
eu check level2.eu   # same fault via import   → exit 0   (--strict makes no difference)
```

`src/driver/check.rs:219` downgrades *any* pipeline failure to `eu check: pipeline warning: …` and
contributes 0 to the error counts, so the exit code comes from the phase-1 annotation scan of the
base file alone. This is general pre-existing behaviour, **not** something this PR introduced — a
circular import and a missing import file, neither touched here, also exit 0 under `eu check`. And
exit 0 is strictly better than the exit 101 that this same input produced before the fix. But it does
mean `eu check --strict` cannot currently be relied on in CI to catch a hard `CoreError` originating
in an imported library, which is arguably the more important half of the original bug. Suggest a
separate bead rather than expanding this PR.

---

**Not merged, and the bead is not closed** — per this session's instruction the owner holds the merge
decision, and the coordinator closes beads after verifying against scope.

gmorpheme:
Both review points addressed in `1c5c9f65`. Thank you — you were right that the coverage was the weak part, and the specific demonstration (two thirds revertible with CI green) is what made it obvious.

## 1. Per-site coverage (substantive)

You were correct: the original two tests both went through the `Unit` impl's loop, so `declscope.eu` and a nested-block reproducer were genuinely unprotected. There is now **one test per fixed site**:

| Test | Site | Reached by |
|---|---|---|
| `194_mqlkv_import_bad_embedding` | `Unit` impl | unit metadata `{ import: … }` |
| `198_mqlkv_scoped_import_bad_embedding` | `rowan_declaration_to_binding` | declaration-scoped `` ` { import: … } `` |
| `199_mqlkv_block_import_bad_embedding` | `Block` impl | leading metadata on a **nested** block |
| `195_mqlkv_import_bad_requires` | `Unit` impl | a *different* `CoreError` variant (unsatisfiable `requires:`) |

All four reuse the one malformed-annotation helper, so the fixtures differ only in the route they take.

### Per-site fault injection — each site reverted individually

| Reverted site | Result |
|---|---|
| `rowan_declaration_to_binding` (`import_smid`) | **198 FAILED**; 194, 195, 199 ok |
| `Block` impl (first `meta_smid`) | **199 FAILED**; 194, 195, 198 ok |
| `Unit` impl (second `meta_smid`) | **194 + 195 FAILED**; 198, 199 ok |
| all three restored | 4 passed, 0 failed |

Exact localisation in every case — no site can now regress with the suite green, and the failure names the site. (194 and 195 both failing on the third injection is expected and intended: they deliberately share that site, because 195's job is to prove the fix is `CoreError`-variant-general rather than annotation-specific.)

## 2. `testdata/` convention

`194_mqlkv_bad_annotation.eu` and `195_mqlkv_impossible_requires.eu` moved to `tests/harness/errors/testdata/` (via `git mv`, so history follows), matching `154_internal_import_error.eu` → `testdata/154_lib.eu`. Both importers updated. Agreed on the reasoning: they are not tests, and a future sidecar-completeness sweep should not have to special-case them.

## Not changed, per your note

- The `self.file` push / `named_import_depth` increment that the new `?` skips — I agree it is harmless today for exactly the reason you gave (each `Desugarer::new` builds a fresh local dropped on the error path), and I have not tried to unwind it. Your note on this PR is the right place for it to live if multi-error recovery ever lands.
- `eu check --strict` exiting 0 on an imported hard `CoreError` — left to **eu-09st0**. Pre-existing and general (circular imports and missing import files behave the same), and folding it in would make this PR two changes.

## Gates re-run after the change

- `cargo fmt --all` — clean
- `cargo test` — **fully green, 0 failures** (harness 510, now including the four mqlkv tests; lib 1358; typecheck 111; and the rest)

CHANGELOG entry updated to describe the per-site coverage rather than the original two-test claim.

Keeping this open and unmerged for the owner.

gmorpheme:
## Re-review — Wicket (gatekeeper), independent verification of `1c5c9f65`

**Verdict: APPROVE.** Both findings addressed. I re-ran the per-site fault injection myself
rather than taking the PR body's table — that demonstration was the substance of the original
finding, so it needed independent repetition.

### Per-site fault injection, repeated independently

Worktree at `1c5c9f65`. One site at a time reverted from `?` back to
`.expect("failure translating import")`, rebuilt, `cargo test --test harness_test mqlkv`:

| Site reverted | Line | Tests that FAILED | Tests that stayed green |
|---|---|---|---|
| `rowan_declaration_to_binding` (scoped import) | 2642 | `test_198_mqlkv_scoped_import_bad_embedding` | 194, 195, 199 |
| `impl Desugarable for Block` (block metadata) | 2828 | `test_199_mqlkv_block_import_bad_embedding` | 194, 195, 198 |
| `impl Desugarable for Unit` (unit metadata) | 2997 | `test_194_mqlkv_import_bad_embedding`, `test_195_mqlkv_import_bad_requires` | 198, 199 |

Exactly the claimed partition: each site is gated by its own test, and no test fires for a
site it does not cover. The earlier state — two of three sites revertible with the suite still
green — is gone.

Note that sites 1 and 3 are *textually identical* after the fix
(`desugarer.translate_import(meta_smid, input)?`), so they can only be distinguished
positionally. Worth knowing for anyone repeating this.

### Second finding

Helpers moved to `tests/harness/errors/testdata/`, so they are no longer collected as harness
tests in their own right. Confirmed.

### Gates

- CI: all 24 checks green (the four `Release *` / `prepare-release` jobs skipping as expected
  on a non-release branch).
- The three test fixtures each exercise a genuinely distinct desugar route — unit metadata,
  declaration-scoped metadata, nested-block metadata — and 195 additionally covers a second
  `CoreError` variant through site 1, so the fix is gated as the general one it is.

Nothing further from me. Not merging: merge decisions on this pair sit with the owner.

gmorpheme:

`5dbc44ec`, with `lib/prelude.blob` regenerated after the branch switch and

### The fix

Three `.expect("failure translating import")` call sites in
`src/core/desugar/rowan_ast.rs` become `?`. That is the whole change, and it is
the right change: **all three enclosing functions already return
`Result<_, CoreError>`** — the `Unit` impl returns
`Err(CoreError::VersionRequirementFailed(..))` a few lines above the panic it
replaced — so the diagnostic path was already there and the `.expect` was
simply bypassing it.

The asymmetry the PR leads with is the convincing part of the diagnosis: the
*same* malformed file diagnosed cleanly with exit 1 when evaluated directly and
aborted with exit 101 when imported. That is not a missing feature, it is one
code path failing to use machinery the other one uses.

The comment placement is right too — the full rationale sits on the `Unit` loop
and the other two point at it, rather than being copied three times and drifting.

### Fault injection — repeated per site, independently

This is the claim I most wanted to check, because "one fix, four tests" is
exactly the shape where three of the tests can be redundant. I reverted **each
`?` back to `.expect` individually**, in my own worktree, and ran the four
fixtures each time:

| Reverted site | Fails |
|---|---|
| `rowan_declaration_to_binding`'s scoped-import loop | **208** only |
| `Block` impl's block-metadata loop | **209** only |
| `Unit` impl's unit-metadata loop | **206 and 207** |

All four pass with every site restored.

That is the correct pattern, and it is exactly what the PR claimed. Each of the
three sites has a test that no other site's revert can satisfy, so **no third
of this fix can regress with the suite green** — which is the specific failure
mode a single end-to-end test would have permitted. The `Unit` site failing two
tests rather than one is right, not sloppy: 206 and 207 both route through unit
metadata and differ in the `CoreError` *variant* they raise, which is the point
of 207.

### Test design

207 is the one that makes this a fix rather than a patch. 206, 208 and 209 all
exercise an invalid embedding; 207 drives
`CoreError::VersionRequirementFailed` through an imported unit's unsatisfiable
`requires: ">=999.0"`. Without it the tests would only demonstrate that *one*
`CoreError` variant now diagnoses, while the fix claims generality over all of
them.

On the `.expect` sidecars: they assert `exit: 1` plus a stderr substring. The
substring is loose, but **the exit code is the real gate here** — a panic is
exit 101 and nothing else in the harness produces that, so a regression cannot
satisfy `exit: 1`. Given the defect being gated is specifically "process abort
instead of diagnostic", that is the assertion that matters, and it is the one
the fixture makes.

The `testdata/` split is the right structure: the malformed units live in
`testdata/` so they are never picked up as fixtures in their own right, and the
four fixtures import them by the three different syntactic routes.

### Renumbering applied

The four error fixtures collided with master and I moved them:

- `194_mqlkv_import_bad_embedding` → **206** (194 is `194_8a49h_library_blame`)
- `195_mqlkv_import_bad_requires` → **207** (195 is `195_og3u6_trace_anchor`)
- `198_mqlkv_scoped_import_bad_embedding` → **208**
- `199_mqlkv_block_import_bad_embedding` → **209**

198 and 199 collided with the SV3 fixtures in `tests/harness/`, which is worth
stating explicitly for the next author: **`tests/harness/` and
`tests/harness/errors/` share one number sequence**, so a free number in one
directory is not necessarily free. The two `testdata/` files moved with them
(`206_mqlkv_bad_annotation`, `207_mqlkv_impossible_requires`) and every
in-fixture `import:` path was updated to match. No reference to the old numbers
survives anywhere in the tree.

### Gates

- `cargo fmt --all` — clean
- `cargo test` — **0 failed across every test binary**, harness 525 (the four
  new error fixtures included), doc examples 111

PR #1078, opened 2026-07-26
feat(sv3): structural contracts — validate, ensure and the t-* closedness projection (eu-u9xj.1)

gmorpheme:
Implements **eu-u9xj.1** (SV3 / W16) per the owner-approved design
`docs/superpowers/specs/2026-07-26-sv3-structural-contracts-design.md` (merged as a1af1f0b).

Plan: `docs/superpowers/plans/2026-07-26-sv3-structural-contracts.md`.

> **RECORDED REVIEW BY A NON-AUTHOR IS REQUIRED BEFORE MERGE.** This PR changes the versioned `t-*` projection, adds an intrinsic, and bumps `BYTECODE_WIRE_FORMAT_VERSION` — the blob/runtime wire surface.

---

## What this builds

`match?` answers one bit. For a 400-line imported manifest one bit is close to useless: the user learns that *something* is wrong and nothing about *what* or *where*. This closes that gap with a **reporting interpreter** over the same `t-*` projection `as-spec` reads. `as-spec` and `match?` are untouched.

```
s"{…}"  →  __TYPE_TO_DATA  →  t-* data  →  [new] reporting interpreter  →  violation report
                                  │
                                  └────────→  as-spec  →  match?  →  bool   (unchanged)
```

- **`validate(spec, data) -> [violation]`** — pure, total in its result, **never raises for any data**. `[]` means conformant. Violations are `{path, kind, expected, found}` blocks with `kind ∈ {:type-mismatch, :missing, :unexpected, :length}` — plain eucalypt data, so a report can be filtered, grouped, counted or emitted like anything else. A malformed **spec** *does* raise (through `panic`, a distinct `ExecutionError` variant) so a schema typo cannot masquerade as bad input data.
- **`ensure(spec, data) -> data`** — unchanged on success, so it drops into a pipeline without restructuring the code around it; raises otherwise.
- **Closedness in the `t-record` projection** — the change that makes reporting surplus keys possible at all, and that makes a spec literal mean the same thing statically and at runtime.
- **`reflect.type-str`** — the canonical `t-*` renderer, hoisted out of `from-data`'s block so a consumer can name the type it expected.
- **`ExecutionError::ContractViolation` / `EU-EVAL-CONTRACT`** — rendered through the existing codespan substrate, blaming the `ensure` call site with one note per violation.

Argument order follows the catenation idiom (receiver last), so `data validate(spec)` and `data ensure(spec)` read correctly.

## Commits

| Commit | Content |
|---|---|
| `2dc6471a` | closedness in the `t-*` projection + the `reflect` renderer changes |
| `3e3b8769` | `reflect.type-str` extraction |
| `6109eef6` | `lib/contract.eu` with `validate`, and the forcing-discipline test |
| `c999745b` | `ensure`, `ContractViolation`, `CONTRACT_FAIL`, guide, CHANGELOG |

## Where the contract functions live, and why

**A new `lib/contract.eu` resource**, not `lib/prelude.eu`. Three reasons, in order of weight:

1. The interpreter needs `reflect.type-str`, which is outside the prelude.
2. Contracts are opt-in, and an opt-in feature that must be imported reads honestly against "cost paid only where written".
3. It keeps the prelude lean — and, incidentally, gave a zero-conflict surface against `pp-fix`'s concurrent `max-of`/`min-of` rewrite. **This PR does not modify `lib/prelude.eu` at all.**

Registered in `src/driver/resources.rs`, so `{ import: "contract.eu" }` resolves with no `-L` flag.

## The wire-format bump is not defensive — it prevents silent wrong answers

Adding `CONTRACT_FAIL` shifts `INTRINSIC_COUNT`, and the blob bakes globals in as `INTRINSIC_COUNT + prelude slot` (`src/eval/stg/blob.rs`, "Global slot numbering") while the freshness hash covers only `lib/prelude.eu` — which this PR does **not** touch. I verified the consequence rather than assuming it. With the constant left at 5 and a blob generated before the intrinsic was added:

```
$ eu -e '[1,2,3] map(inc)'
[0, 1, 2]          # should be [2, 3, 4]
```

No warning, no error — an off-by-one global slot map silently resolving `inc` to a different prelude binding. With the bump to **v6**: `warning: prelude blob is stale — run cargo xtask prelude-compile`, and after regenerating, `[2, 3, 4]`.

The doc comment on the constant now states the rule ("bump whenever the intrinsic table gains or loses an entry, not only when the encoding changes"). Mechanical enforcement — deriving the version from the intrinsic table rather than trusting a comment — is tracked separately as **eu-3skeg** (P1), filed by the coordinator off this finding.

## Tests

Five harness tests and two error tests, all computing `RESULT` from their checks so every assertion is in the verdict.

| Test | Gates |
|---|---|
| `191_sv3_projection_closedness` (13 checks) | the trailing flag for closed / `..` / `..r`; nested records; the arity-tolerant read of a hand-written two-element node; the open-record and literal-string round-trips |
| `192_sv3_type_str` (16 checks) | `type-str` over the tag vocabulary, incl. totality on an unknown tag; `from-data` unchanged |
| `193_sv3_validate` (79 checks) | conformance at every constructor; each of the four kinds; accumulation; ordering; paths incl. the quoted form; unions reporting one entry; container mismatch terminating the subtree; optional fields; open vs closed; already-projected `t-*` specs; and a 15-pair `as-spec`-agreement corpus |
| `194_sv3_forcing_discipline` (13 checks) | see below |
| `195_sv3_ensure` (10 checks) | data returned unchanged; optional fields; forcing inherited; the report-line renderers |
| `errors/196_sv3_contract_violation` | `exit: 1` + one `(?s)` regex over the code, the headline's count and rendered type, and both violation notes |
| `errors/197_sv3_contract_bad_spec` | `exit: 1` + the *distinct* `panic: validate: not a type spec` — what makes design decision 6 testable |

### How spec-directed forcing is gated

`194_sv3_forcing_discipline` is built from values that **raise or diverge if forced**, so a regression in the discipline turns the file from PASS into a hard error. Two facts had to be established empirically first:

- A **block literal**'s field values are effectively strict under demand analysis — `{ a: 1, b: panic("X") } block?` raises. A bomb cannot be planted with a literal, so a test written that way would gate nothing.
- `block([[:a, 1], [:b, panic("X")]])` **is** lazy in its values. That is the construction the test uses.

The bomb is verified **live** (naming it in the spec does raise), so the test cannot pass for the wrong reason. Covered: an empty record spec; an open spec naming one key; a **closed** spec (proving closedness detection is a key-set comparison, never a value forcing); `any`/`top`/type variables; an unrecognised tag; an infinite list under an unnamed key (cheap WHNF, divergent spine); a nested bomb; and `ensure` inheriting all of it.

## Fault injection — every test, done

**Task 1 — projection** (`191`)
| Injection | Result |
|---|---|
| `open_flag` hardcoded `false` | FAIL (t4/t5/t7) → restore → PASS |
| `open-flag` arity guard → `rest second` | FAIL (t12) → restore → PASS |
| `record-str` tail → `""` | FAIL (t4/t9) → restore → PASS |

**Task 2 — `type-str`** (`192`)
| Injection | Result |
|---|---|
| `from-data` → `"any" __TYPE_FROM_STRING` | FAIL → restore → PASS |
| `:t-lit-str` arm removed | FAIL (191 t13) → restore → PASS |

**Task 3 — `validate`** (`193`)
| Injection | Result |
|---|---|
| `record-check` ignores the open flag | FAIL → restore → PASS |
| optional fields treated as required | FAIL → restore → PASS |
| union expands one violation per branch | FAIL → restore → PASS |
| `path-index` → `.{i}` instead of `[{i}]` | FAIL → restore → PASS |
| `record-check` descends into a non-block | FAIL → restore → PASS |

**Task 4 — forcing** (`194`)
| Injection | Result |
|---|---|
| `record-check` forces every value (`v values map(type-name) all(string?)`) | FAIL with `panic: SV3: an unmentioned subtree was forced` → restore → PASS |
| `any`/`top` forced instead of accepted | FAIL → restore → PASS |
| unrecognised tag forces its value | FAIL → restore → PASS |

*(A first attempt at the forcing injection — `… map(type-name) count` — did **not** fail, because `count` walks the mapped spine without forcing elements. Worth recording: the injection itself had to be verified live before it proved anything.)*

**Task 5 — `ensure` / `ContractViolation`** (`195`, `196`, `197`)
| Injection | Result |
|---|---|
| `ensure` returns `data` unconditionally | FAIL (196) → restore → PASS |
| `spec-shaped?` guard removed | FAIL (197) → restore → PASS |
| `code()` returns `None` for `ContractViolation` | FAIL (196) → restore → PASS |
| `ContractViolation` notes → `vec![]` | FAIL (196) → restore → PASS |
| `:unexpected` line drops its keys | FAIL (195, 196) → restore → PASS |

## Both engines, byte-identical

All seven new files, bytecode vs `EU_HEAPSYN=1`, stdout+stderr compared:

```
SAME     tests/harness/191_sv3_projection_closedness.eu
SAME     tests/harness/192_sv3_type_str.eu
SAME     tests/harness/193_sv3_validate.eu
SAME     tests/harness/194_sv3_forcing_discipline.eu
SAME     tests/harness/195_sv3_ensure.eu
SAME     tests/harness/errors/196_sv3_contract_violation.eu
SAME     tests/harness/errors/197_sv3_contract_bad_spec.eu
```

Full `cargo test` and `EU_HEAPSYN=1 cargo test` run **with a freshly generated `lib/prelude.blob`** (so the new intrinsic and the bumped version are exercised on the blob path, which is the one that matters). Both runs show **exactly three failures and no others**, all of them **pre-existing on unmodified master** — `test_193_1tkk_7_12_curated_trace`, `diagnostics_invariants::secondary_labels_do_not_excerpt_library_internals`, `diagnostics_invariants::debug_trace_restores_the_uncurated_trace`. I verified this by checking out `a1af1f0b`, regenerating the blob and running there. They are tracked as **eu-7x0r** (P2), and the blob-mode CI job carries deliberate exclusions for them. Nothing in this PR touches that machinery.

`EU_GC_VERIFY=2 cargo test --test harness_test` — see the comment below for the result.

## Deliberate departures from the spec

Flagged rather than silently invented:

1. **`BYTECODE_WIRE_FORMAT_VERSION` 5 → 6.** The spec does not mention the blob. Justified above, with the failure mode demonstrated.
2. **A `:t-lit-str` arm in `reflect`'s renderer.** Spec §5.2 names only `record-str`. Without this arm a literal string type renders through the `"any"` fallthrough, which would make `expected:` wrong for any violation involving one, and leaves `from-data` lossy for literal strings (verified on master). Same class of fix as the `..` loss the spec does call out.
3. **Record rendering drops its inner padding** to match `Type`'s `Display` impl exactly. Previously invisible, because `from-data` re-parsed and canonicalised the result; now visible, because `type-str` returns the string directly.
4. **The root path renders as `(root)` in a violation *line*.** Spec §6.3 fixes the root *path* as `""` but says nothing about the line; an empty string would leave a line starting with a bare colon.
5. **`type-name` returns `"unknown"`** for a value matching none of the ten documented runtime type names.
6. **`ContractViolation` carries `Smid::default()`, not `machine.annotation()`.** The wrapper has just forced a string built by interpolation inside a library, and library `.eu` resources are **not** marked as resource files in the `SourceMap` (only the prelude is), so that annotation would win `to_diagnostic`'s prefer-a-user-file test and make a library string literal the primary label. Leaving the Smid empty hands the choice to the trace search, which finds the `ensure` call site — what spec §9.1 asks the error to blame. Verified end-to-end.

## Points where the spec proved underspecified, or where I hit language quirks

Reported, not resolved unilaterally:

- **Spec §15.1/§15.2 (owner already flagged these).** `expected` is not uniformly typed across kinds — a type string, a key name, `:closed`, a number — and `:missing` does not carry the missing field's type. Implemented exactly as §6.2 specifies; not changed here.
- **Spec §13.7 — `as-spec` and `validate` genuinely disagree on literal types.** `as-spec` has no `:t-lit-str`/`:t-lit-sym` arm, so `s"\"prod\"" as-spec` matches anything, while `validate` checks by equality. §13.7 calls the `as-spec` behaviour a probable defect and says SV3 should do the right thing on its own side, which it does. The agreement corpus excludes literal types for this reason. **Worth a follow-up bead.**
- **The type DSL has no quoted-key syntax.** `s"{'my key': string}"` is a parse error, so the non-identifier-key path test constructs its spec as `t-*` data directly. The path *rendering* is specified (§6.3) and implemented; only the surface syntax to reach it is missing.
- **String-lexer quirk worked around:** a literal `}}` immediately following an interpolation is silently dropped — `"{x}}}"` yields `abc` for `x = "abc"`, not `abc}`. `record-str` interpolates the closing brace from a binding instead. Filed by the coordinator as **eu-uhd7d** (P2).
- **A P1 panic found and fixed separately.** An invalid embedding in an *imported* `.eu` file panicked with exit 101 instead of diagnosing (`.expect("failure translating import")`). Filed as **eu-mqlkv**, fixed in **#1077**, deliberately kept out of this PR.

## Documentation

- `docs/guide/contracts.md` — new guide chapter, linked from `docs/SUMMARY.md`; the spec's §10 worked example is its centrepiece
- `docs/guide/type-checking.md` — the `:t-record` closedness element and its arity tolerance; a `type-str` subsection
- `docs/reference/agent-reference.md` §3.5b — dense reference entry
- `docs/reference/error-codes.md` + `src/driver/error_codes.rs` — `EU-EVAL-CONTRACT` (verified via `eu error EU-EVAL-CONTRACT`)
- `CHANGELOG.md` — three Unreleased entries

All eucalypt fences under `docs/` are `eu,notest`.

## Gates

- `cargo fmt --all` — clean
- `cargo test` (bytecode, blob) — 3 pre-existing failures only
- `EU_HEAPSYN=1 cargo test` (blob) — the same 3, nothing else
- `lib/prelude.blob` regenerated with `cargo xtask prelude-compile` (gitignored, so not committed)


gmorpheme:
**`EU_GC_VERIFY=2` result** (full multi-checkpoint structural verification — header validity, pointer validity, line consistency post-mark, forwarding-pointer lifecycle post-evacuate/update, block-list integrity post-sweep):

```
$ EU_GC_VERIFY=2 cargo test --test harness_test --no-fail-fast
test test_193_1tkk_7_12_curated_trace ... FAILED
test result: FAILED. 512 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 61.78s
```

512 of 513 pass; the single failure is the pre-existing blob-mode one tracked as **eu-7x0r**. **No GC verification panic of any kind** — every structural checkpoint passed across the whole harness, including the seven new SV3 files. The new intrinsic allocates nothing beyond a `String` and a `Vec<String>` on the Rust side (both immediately consumed into the error variant), and the reporting interpreter is ordinary eucalypt allocation, so this is the expected result rather than a surprising one — but it is the result, measured.

gmorpheme:
## Review — Wicket (gatekeeper), recorded non-author review

**Verdict: APPROVE, with seven non-blocking findings.** This is the recorded non-author review
the wire-format/intrinsic surface requires. **I am not merging** — a new intrinsic and an
observable-behaviour change put this in the owner-review category, and finding **F3** below is
something the owner should see before this lands.

Everything in the "verified" section I reproduced in a clean worktree
(`/tmp/eu-wicket-sv3`, `origin/feat/quill-sv3-contracts` @ `c999745b`) against a freshly built
master baseline (`a1af1f0b`, own blob). Where I took something on trust I say so.

---

## The `Smid::default()` departure — you are right, and the stated reason is wrong

This was flagged to me as the most doubtful item in the PR, on the grounds that
`src/driver/source.rs:705` marks *every* `Locator::Resource(_)` as a resource file, so
departure 6's premise ("library `.eu` resources are **not** marked as resource files in the
`SourceMap` (only the prelude is)") looked factually false — and therefore that the workaround
had discarded spec §9.1's stated blame behaviour for a problem that does not exist.

**I built the reproducer. The decision is correct; the justification as written is not.**

Injecting `machine.annotation()` in place of `Smid::default()` and rebuilding:

```
error[EU-EVAL-CONTRACT]: contract violation: 2 violations against {name: string, port: number}
    ┌─ reflect.eu:50:13
    │
 50 │     rbrace: "}}"
    │             ^^^^
```

The primary label becomes a library string literal and `deploy.eu:7:13` — the `ensure` call
site — is gone from the primary position entirely. With `Smid::default()` as shipped:

```
error[EU-EVAL-CONTRACT]: contract violation: 2 violations against {name: string, port: number}
    ┌─ deploy.eu:7:13
    │
  7 │ config: raw ensure(schema)
    │             ^^^^^^
```

So `Smid::default()` is what delivers spec §9.1, and `machine.annotation()` would defeat it.

**But the reason is not the one in the comment.** The `SourceMap` mechanism is not absent — it
is bypassed. `import: "contract.eu"` becomes `Locator::Fs("contract.eu")`
(`src/syntax/input.rs:49-79` — only an explicit `resource:` prefix yields
`Locator::Resource`). `read_fs_input` (`src/driver/source.rs:1216-1223`) then falls back to the
baked-in resource *text* by filename stem, while the **locator stays `Locator::Fs`** — so
`source.rs:704`'s `matches!(locator, Locator::Resource(_))` never matches and
`mark_resource_file` never fires. The prelude is marked only because `prepare.rs:41` constructs
it as a genuine `Locator::Resource("prelude")`.

The observable proof is in the shipped diagnostic itself: `reflect.eu` and `contract.eu` appear
as *secondary* labels, and `to_diagnostic` only emits those for frames where
`classify_frame(smid) == FrameKind::User`, which requires `is_user_file` to be **true**.

Please correct the comment at `src/eval/stg/contract.rs:91-105`. As written it will send the
next reader to "fix" a non-problem — which is exactly what nearly happened in this review.
Suggested substance: *contract.eu and reflect.eu are loaded through `Locator::Fs` plus
`read_fs_input`'s resource-text fallback, so `mark_resource_file` is never called for them and
`is_user_file` reports true; until that is fixed (see F3) the machine annotation would win
`to_diagnostic`'s prefer-a-user-file test.*

---

## The forcing bomb genuinely detonates

Verified two ways, independently of the PR's table.

Direct reproducer, `bomb-blk` under `` ` :suppress `` exactly as in `194`:

```
# spec NAMES b  -> error: panic: SV3: an unmentioned subtree was forced
# spec omits b  -> safe: true
```

Then by injection into `record-check`. **My first attempt did not fail** — I added
`forced: v values map(type-name) all(string?)` as a block binding without referencing it, and
an unreferenced binding is never forced. Rewriting the injection so the result depends on it:

```eu,notest
}.((v values map(type-name) all(string?)) then(field-vs ++ extras, [])))
```

→ `test_194_sv3_forcing_discipline` **FAILED**, `test_195_sv3_ensure` **FAILED**, the other
five green; restore → all seven green. So the bomb is live and the test gates. Your note about
the `map(type-name) count` attempt not failing is the same class of trap from the other side,
and recording it was the right call.

## Wire-format bump 5 → 6 — correct **and** sufficient

Measured rather than reasoned. Generated a blob with the constant at 5, restored it to 6, and
rebuilt:

```
warning: eucalypt@0.14.0: prelude blob is stale — run `cargo xtask prelude-compile` to regenerate.
$ eu -e '[1,2,3] map(inc)'
[2, 3, 4]
```

No silent wrong answer: `build.rs` folds the version into the source hash, the mismatch sets
`cfg(prelude_blob_stale)`, and the driver falls back to the source prelude.

**Sufficient**, additionally, because the blob is compile-time embedded
(`resources::PRELUDE_BLOB_BYTES`, consumed only under `#[cfg(prelude_blob_ok)]` in
`maybe_load_prelude_blob`) — blob bytes and binary are always built together, so a stale blob
cannot reach a runtime by being swapped in afterwards. There is no runtime hole for eu-3skeg to
need to close; the gap it tracks is purely "a human must remember to bump", which is the right
framing.

I did **not** reproduce your `[0,1,2]` figure (that needs a master-vintage blob paired with a
195-intrinsic binary and the constant left at 5). Taken on trust; it is consistent with the
`INTRINSIC_COUNT + prelude slot` numbering in `blob.rs`, and it does not affect the verdict
since the bump's correctness is established independently above.

## The projection change is backward-safe, and the round-trip fixes are real

Enumerated every `t-record` consumer in the tree. `as-spec`'s `to-spec-td` reads only
`rest first` (`lib/prelude.eu:2252`) — genuinely unaffected, as the CHANGELOG claims.
`182_typedata_alias_resolution` tests only `first = :t-record`. `reflect`'s `record-str` is
arity-guarded, and `193` t79 covers the two-element node end to end. No Rust consumer reads
`t-*` data structurally. Additive growth with a tolerant reader: correct.

Both claimed round-trip losses measured on the freshly built master, not inferred:

| | master `a1af1f0b` | this PR |
|---|---|---|
| `s"{a: number, ..}" to-data from-data` = itself | `false` (comes back **closed**) | `true` |
| `c"\"prod\"" … to-data from-data` = itself | `false` (comes back **`any`**) | `true` |

Departure 3 (dropping record inner padding) also checks out: `Type`'s `Display`
(`types.rs:642-669`) writes `{a: number}` and `{..}` with no padding, and `record-str` now
matches it exactly.

## Gates I ran myself

- **CI**: all 24 checks green; the four `Release *` / `prepare-release` jobs `skipping`, as
  expected off a release branch.
- `cargo fmt --all --check` clean; `cargo clippy --workspace --all-targets -- -D warnings`
  clean.
- `cargo test --test harness_test`: **512 passed, 1 failed**
  (`test_193_1tkk_7_12_curated_trace`); `--test diagnostics_invariants`: 1 passed, 2 failed.
  **All three reproduce on freshly built master with a regenerated blob** — confirmed
  pre-existing (eu-7x0r), not this PR's.
- All seven new tests green. Both engines byte-identical on all seven files (my own
  comparison, stdout+stderr, `RESULT: PASS` and zero `false` checks in each).
- `EU_GC_VERIFY=2 EU_GC_POISON=1` on the SV3 tests: clean.
- Independent fault injection, eight cases, each FAIL-then-PASS-on-restore: `open_flag`
  hardcoded false (191–195); `record-check` ignoring the open flag (193/194/195); forcing every
  value (194/195); optional-as-required (193/195); `:t-lit-str` arm removed (191);
  `code()` → `None` (196); `ensure` returning data unconditionally (196); notes → `vec![]`
  (196). Every new test is genuinely load-bearing.
- The guide's worked example reproduces byte-for-byte against a real `servers.yaml`.

## Spec conformance

Walked §4 through §9 and §12 against the code. Report shape (§6.2), path grammar (§6.3),
per-constructor semantics (§6.4), record closedness (§6.5), forcing discipline (§6.6),
malformed-spec-as-`panic` (§6.7), ordering (§6.8), `ensure` (§7.1), variant shape and
`EU-EVAL-CONTRACT` (§7.2), string marshalling boundary (§7.3), blame (§9.1): all implemented as
specified. All five phases P1–P5 delivered; P5's "prelude/library reference regeneration" is
vacuous here (`docs/reference/prelude/` covers only `prelude.eu`, which this PR does not
touch), not skipped. Deliberate departures 1–5 all check out; 6 is the comment-wording issue
above.

---

## Findings

**F1 (medium) — `servers[2].port` is not a valid eucalypt lookup expression.**
`docs/guide/contracts.md` says a rendered path "is a valid eucalypt lookup expression … so it
can be pasted into a `.`-chain or a lens path", and the CHANGELOG says paths are "valid
eucalypt lookup expressions". They are not, for any path containing an index:

```
probe: servers[2].port
# error: tried to call a list as a function
```

`[2]` is a list literal, so this is catenation. Inherited from spec §6.3, so not your
invention, but it is now a user-facing false claim in two shipped documents. The path *format*
is fine and I am not asking you to change it — only the wording.

**F2 (medium) — "never raises for any data" is inaccurate.**
In the `lib/contract.eu` doc string, the guide's two-function table, the CHANGELOG and
agent-reference §3.5b. Demonstrated:

```eu,notest
b: block([[:a, panic("data value raised")]])
r: b validate(s"{a: number}")     # error: panic: data value raised
```

`validate` never raises *on account of the data's shape*, but it inherits any raise from
forcing a value the spec **names** — which is the unavoidable flip side of the spec-directed
forcing the same paragraphs advertise. Reword to the narrower, true claim; the current wording
promises totality the function cannot have.

**F3 (medium; pre-existing root cause, newly load-bearing) — the real defect behind departure 6.**
Baked-in library resources reached through `read_fs_input`'s filename fallback are never
`mark_resource_file`d, so `is_user_file` reports true for `contract.eu`, `reflect.eu`,
`markup.eu` and `test.eu`. Consequence, visible in the shipped diagnostic:

```
    ┌─ reflect.eu:50:13
 50 │     rbrace: "}}"
    │             ---- called from here
    ┌─ contract.eu:237:1
237 │ ╭ ensure(spec, data):
    ⋮   … the whole body of `ensure`, six lines …
```

Two secondary labels excerpting library internals the user did not write — precisely what
`diagnostics_invariants::secondary_labels_do_not_excerpt_library_internals` exists to prevent.
It also means the rendered example in `docs/guide/contracts.md` (and spec §7.4) does not match
what the tool actually prints. Not this PR's to fix, and not a blocker — but worth a P2 bead:
mark the resource-fallback file in `read_fs_input`/`load_source`, after which
`Smid::default()` becomes unnecessary *and* the secondary noise disappears. Flagging to the
owner because it is the one place where this PR's shipped output is visibly worse than its own
documentation.

**F4 (low) — the property departure 6 exists to protect is not gated.**
`196_sv3_contract_violation.eu.expect`'s regex covers the code, the headline and both notes, but
not the primary label. So if someone "corrects" `Smid::default()` to `machine.annotation()` on
the strength of the current comment, CI stays green while the blame silently moves into
`reflect.eu`. Anchoring a `196_sv3_contract_violation\.eu:\d+:\d+` in the regex would close
that.

**F5 (low) — the guide omits the literal-type divergence.**
"Contracts and `as-spec`" names the closed-record divergence, and *What is not covered* says
`Dict`/`NonEmpty`/recursive types are accepted "matching `as-spec`, so the two never disagree
about what conforms". They do disagree, on literal types. Add a bullet pointing at **eu-pub3r**
so a user meets this in the guide rather than in a bead.

**F6 (low, architectural) — two interpreters over one projection.**
`as-spec`'s `to-spec-td` and `contract.eu`'s `walk` now independently interpret the same `t-*`
surface. That duplication *is* why eu-pub3r's divergence can exist and why `193`'s agreement
corpus needs exclusions for both literals and closed records. The general answer is one
interpreter with `match?(spec)` defined as `validate(spec) nil?`. That changes shipped
`as-spec` behaviour, so it is an owner call, not something to do here — raising it per the
architectural-smell check rather than asking for a change.

**F7 (low) — `spec-shaped?` checks only the top node.**
A malformed *nested* node (e.g. `[:t-record, {a: "notalist"}]`) raises an unrelated error
rather than `validate: not a type spec`. Spec §6.7 only specifies the top level, so this
conforms — noting it so the limit is on record.

## On eu-pub3r: ship the divergence

Asked whether the `as-spec`/`validate` literal-type disagreement should block. **My read: no.**
It is sanctioned by the approved spec (§13.7), the underlying defect is filed and open as a P2,
the corpus exclusion is disclosed in the test file and the PR body rather than hidden, and
`validate` is the one behaving correctly. Fixing `as-spec` here would change shipped `match?`
semantics inside the same PR that changes the wire format — worse review hygiene than shipping
a documented, tracked divergence. F5 is the part that should be fixed, so the divergence is
discoverable from the guide.

---

## Summary

Excellent work — the design is faithfully implemented, the tests genuinely gate (I broke all
eight things myself), the wire-format reasoning is sound and I verified it independently, and
the departures are flagged rather than smuggled. F1/F2 are wording fixes in shipped docs, F4 is
a two-token regex addition, F5 a bullet, F3 a follow-up bead. None of them touch the design.

Not merging: new intrinsic + observable behaviour change + F3 all warrant the owner's eye, and
merge authority on this one is theirs.

gmorpheme:
Review findings addressed in `edee0531`. Thank you — F1, F2 and F3 were all genuinely wrong as written, and **F4 turned out to be the important one**.

## ⚠️ F4 was not a missing assertion — it was a missing assertion hiding a real defect

I added the anchor you asked for. **It failed.** In **blob mode — the shipped path** — the primary label is `contract.eu:101:5`, not the user's `ensure` call site. Source-prelude mode was correct, which is why nobody saw it.

`EU_ERROR_TRACE_DUMP=1` in both modes, same trace shape, different head frame:

| Mode | `stack_trace[0]` | Outcome |
|---|---|---|
| source-prelude | `foldr` in the **prelude** (a marked resource) | rejected → falls through to `config` in the user file → **correct** |
| blob | `mismatch` in **`lib/contract.eu`** (`is_user_file` → true) | wins → **blame lands in the library** |

So `Smid::default()` was **necessary but not sufficient**: it removed the error's *own* Smid from contention, but the trace search still ranks library frames as user frames. The mode difference was incidental — whichever library frame happens to sit at position 0.

**This PR therefore does not yet meet its own spec §9.1 in blob mode.** Rather than leave that to be discovered, it is now stated in three places: the guide (a "Known limitation" box), the CHANGELOG entry, and the `CONTRACT_FAIL` comment.

**The root cause is eu-8a49h, now fixed in #1080** (split out on the coordinator's instruction, since it changes diagnostics for *every* library and this PR already carries a recorded approval). With #1080 applied, blob-mode blame is the `ensure` call site and the anchored regex passes in both modes. **The F4 anchor is deliberately not in this PR** — today it would fail rather than gate anything. It lands when #1080 does, and the comment says so.

## The four wording fixes

**F1 — paths are not valid lookup expressions.** Verified your claim first: `data.(servers[1].port)` → `error: tried to call a list as a function`. Guide and CHANGELOG now say the format is chosen for readability and is not eucalypt source, and show how to navigate.

Worth recording: my first replacement example repeated **the very gotcha F1 is about**. `data.servers nth(2) .port` also fails (`cannot look up a key on a function value`) because `.` binds tighter than catenation. The guide now shows `(data.servers nth(2)).port` and the lens form `data view(‹:servers 2 :port›)` — **both evaluated, both `3`**.

**F2 — "never raises for any data".** Corrected in all four places to the narrower true claim: no *shape* of data makes `validate` raise, but it propagates a raise from forcing a value the spec **names** — total with respect to shape, not evaluation. Verified by probe: a bomb under a named key raises through `validate`; under an unnamed key it does not.

**F3 — the comment.** Now says the mechanism is **bypassed**, not absent, with the chain spelled out (`Locator::Resource` needs an explicit `resource:` prefix per `src/syntax/input.rs`; the filename form yields `Locator::Fs`; `read_fs_input` serves baked-in text with the locator unchanged) and eu-8a49h cited. You were right that this mattered — as written it would have sent the next reader to fix a non-problem, and it nearly cost this review a wrong conclusion.

**F5 — literal-type divergence.** Guide bullet added citing eu-pub3r, and the "never disagree about what conforms" claim narrowed to the `Dict`/`NonEmpty`/`Mu` cases where it is true. Also added to agent-reference §3.5b. Did not touch `as-spec`, per your ruling.

## Not changed

- **F6** (two interpreters over one `t-*` projection) — eu-lzfj5, owner call.
- **F7** (`spec-shaped?` checks only the top node) — conforms to §6.7.
- **eu-pub3r** — shipping the divergence as you ruled; now disclosed in the guide rather than only in the bead.
- **`BYTECODE_WIRE_FORMAT_VERSION` stays at 6.** Noted that #1079 also bumps 5 → 6 and whichever merges second must go to 7, updating **both** `build.rs` and `xtask/src/main.rs`.

## Validation after the change

Run with a freshly generated blob, so the new intrinsic and bumped version are exercised on the shipped path:

| Gate | Result |
|---|---|
| `cargo fmt --all` | clean |
| `cargo clippy --workspace --all-targets -- -D warnings` | clean |
| `cargo test` (bytecode + blob) | 512/513 harness; **only the three pre-existing eu-7x0r failures**, nothing else |
| all 11 SV3 tests | pass |

## Suggested merge order

**#1080 (eu-8a49h) → this PR.** Then rebase, add the primary-label anchor, and confirm it passes in both prelude modes.

gmorpheme:
## Recorded review — wicket (non-author)

Reviewed in an isolated worktree at `1e8183d4` (this branch merged with
master `0e1fd609`). This satisfies the "recorded review by a non-author"
requirement the PR body asks for.

**Verdict: approve, with three corrections applied on the merge commit** —
all three because claims true when this branch was written became false once
#1080 and #1082 landed.

---

### Spec conformance (`docs/superpowers/specs/2026-07-26-sv3-structural-contracts-design.md`)

All five §14 phases are present and I checked each against code rather than
against the bead:

| Phase | Evidence |
|---|---|
| P1 closedness | `src/eval/stg/typedata.rs` emits the third element as `*open \|\| !rows.is_empty()`; `lib/reflect.eu` `record-str` reads it through `open-flag`'s arity guard |
| P2 `type-str` | hoisted to top level; `from-data(td): td type-str __TYPE_FROM_STRING` |
| P3 `validate` | `lib/contract.eu`, the reporting interpreter |
| P4 `ensure` | `ExecutionError::ContractViolation`, `EU-EVAL-CONTRACT` catalogue entry, `CONTRACT_FAIL`, `ensure` |
| P5 docs | guide chapter, `type-checking.md`, `agent-reference.md` §3.5b, `error-codes.md`, CHANGELOG |

§12.4's gates pass: `cargo test`, `cargo clippy --workspace --all-targets -D
warnings`, `cargo fmt --all --check`. Every `eu` fence under `docs/` in this
PR is `eu,notest` (8 of 8 in `contracts.md`) — the doctest CI executes bare
`eu` fences, and that has reddened master before.

P5's "prelude/library reference regeneration (`eu doc`)" is a no-op here:
`docs/reference/prelude/` is generated from `lib/prelude.eu`, which this PR
correctly does not touch, and there is no generated reference for `reflect.eu`
or `contract.eu`.

### §9.1 blame — this is the substantive change I made

The PR deliberately omitted a primary-label assertion from the contract-violation
error test, on the reasoning (recorded at length in `contract.rs`) that
eu-8a49h made blame land inside `lib/contract.eu` in blob mode, so such an
assertion "would fail today rather than gate anything". That reasoning was
correct when written. **eu-8a49h is now fixed on master (#1080)**, so I added
the anchor:

```
stderr: "(?s)EU-EVAL-CONTRACT.*2 violations against \{name: string, port: number\}.*┌─[^\n]*201_sv3_contract_violation\.eu:8:52.*port: expected number, found string.*unexpected keys: debug"
```

Verified, not assumed:

- **Passes in blob mode** (`prelude_blob_ok` confirmed from the build-script
  output, blob regenerated at this commit).
- **Passes in source-prelude mode** (`lib/prelude.blob` removed, rebuild
  reported `cargo:rustc-cfg=prelude_blob_stale`). The `contract.rs` comment
  specifically asked for this both-modes check before the assertion was added.
- **Genuinely gates.** I fault-injected the eu-8a49h class directly by
  disabling `mark_resource_file` in `src/driver/source.rs`. The primary label
  moved to `lib/reflect.eu:50` (the `rbrace: "}}"` binding) and
  `test_201_sv3_contract_violation` FAILED. Restored, it passes. So SV3 §9.1
  is now met *and* held.

I also fault-injected the `Smid::default()` choice itself by restoring
`machine.annotation()`. **The test still passes** — so that choice is no longer
demonstrably load-bearing, only conservative. I have left the code as-is and
said so in the comment rather than silently leaving a rationale that no longer
holds.

### Test gating

All five harness tests compute `RESULT` from their checks, and I verified there
are no orphaned assertions — `198_sv3_validate.eu` declares exactly 79 `tN`
bindings and all 79 appear in the `all-true?` list. `199_sv3_forcing_discipline`
is the strongest of them: its bombs are built with `block([[...]])` rather than
block literals, which is the right call, because demand analysis makes a literal
block's field values effectively strict and a literal-built bomb would detonate
at construction and gate nothing.

### Correctness spot-checks

- **Intrinsic slot numbering.** `CONTRACT_FAIL` is appended at index 195 in the
  `INTRINSICS` table but `rt.add` registers it *mid-list*, after `panic::Panic`.
  That asymmetry is safe: `StandardRuntime::add` uses `imp.index()`, which
  resolves by name through `intrinsics::index`, so registration order carries no
  slot meaning. Worth stating because a positional scheme here would have been a
  silent off-by-one across every later intrinsic.
- **`open-flag` arity tolerance** reads `false` for a two-element node in both
  `lib/contract.eu` and `lib/reflect.eu`, so hand-written `[:t-record, {…}]`
  keeps working — gated by `196` t12.
- **Record rendering lost its inner padding** (`{a: number}`, not
  `{ a: number }`). This is a user-visible change to `type-str`/`from-data`
  output, invisible before only because `from-data` re-parsed the string. The
  PR states it matches `Type`'s `Display` impl; the full suite is green, so
  nothing else asserted the padded form.
- **Permissive fallthroughs.** `walk`'s default arm returns `[]` and
  `prim-check` falls back to `any?`, so an *unrecognised* tag or primitive name
  inside a well-formed spec accepts silently. That is deliberate — it is what
  makes the `t-*` surface additively extensible, and `199` t9 pins it — but it
  does mean a hand-written spec with a typo'd tag validates everything. Only
  reachable for hand-built `t-*` data, never for `s"…"` literals. Not a blocker;
  noting it so it is on the record.

### Two stale claims corrected on the merge commit

1. **CHANGELOG: the wire-format bump.** The entry claimed
   `BYTECODE_WIRE_FORMAT_VERSION` is bumped to v6 "because the blob source hash
   covers only `lib/prelude.eu`". After #1082 (eu-3skeg) that is no longer the
   mechanism: `src/eval/intrinsics.rs` is hashed into the blob source hash
   directly, and the constant lives in one place reading 7. **Observed live and
   this is the nicest confirmation in the queue** — building this branch, which
   adds an intrinsic and does not touch `lib/prelude.eu` at all, emitted
   `prelude blob is stale` purely from the intrinsic-table hash. eu-3skeg
   catching this exact PR's staleness class on its first real case.
2. **Guide + `contract.rs`: the "Known limitation" blame caveat.** Removed and
   replaced with what is now true, plus a pointer to the anchor that holds it.

### What I did NOT verify

- I did not re-run the author's 18 fault injections. I independently ran the
  one that mattered for the change I made (the eu-8a49h class) and confirmed
  the RESULT-coverage property structurally for all five harness tests.
- I did not measure performance. Nothing here claims a performance figure, so
  nothing needs a measured-verified number.
- I did not exercise `validate` against the deferred §13.7 `as-spec`
  literal-type disagreement; the spec defers it explicitly and `198`'s
  agreement corpus correctly excludes the cases it affects.

PR #1079, opened 2026-07-26
fix(diagnostics): render prelude frames in blob-mode traces (eu-7x0r)

gmorpheme:
Fixes **eu-7x0r** (P1) — the three diagnostics tests that fail under blob mode — and unblocks **eu-5yn9v**.

## Root cause

Two distinct defects in how prelude Smids survive the blob. Neither is trace *curation*: `classify_frame` was doing its job correctly all along and the frames were being thrown away downstream of it.

### 1. Global-slot identities were unrenderable

`Smid::global_slot` deliberately has no `SourceMap` entry (it indexes "which prelude global", not "which source position"). `classify_frame` was taught to resolve those (eu-1tkk.7.11's `slot_to_name`), but the two places that turn a classified frame into text were not:

- `SourceMap::resolve_trace_entry` — `src/common/sourcemap.rs:455`: `let info = self.source.get(smid.get()?)?;` → `None`. Shared by `format_trace` (raw `--debug-trace`) and `format_curated_trace` (human).
- `Executor::json_trace` — `src/driver/eval.rs:833`: `self.source_map.source_info_for_smid(smid)?` → `None` (explicit global-slot guard).

So **every** prelude frame in a blob-mode trace was classified and then silently dropped. On the path that actually ships, no library combinator was ever named — the boundary combinator that is the centrepiece of a curated trace in the design spec simply did not appear.

Naming a slot needed no new data. Giving it a *location* did: blob mode never loads prelude source, so there was no file for a span to be in.

### 2. `xtask`-baked Smids leaked into runtime traces and aliased user code

`Ann` nodes are elided and `Lambda` annotations replaced at reconstruction, but `ArenaStgSyn::DirectApp` and `ArenaStgSyn::LookupLit` carry a `Smid` in a *data* field that the machine uses as a live annotation (`DirectApp` → the call-site annotation an `ApplyTo` continuation records; `LookupLit` → the location a failed lookup reports). Those were copied verbatim at `src/eval/stg/arena.rs:489` and `:537`.

They are indices into `xtask`'s own `SourceMap`. In a short map they merely fail to resolve — which is why the symptom looked like "frames missing". In any program whose own `SourceMap` grows past them, they **resolve against unrelated user declarations**. On a 900-declaration fixture, `xs nth(10)` produced:

```
  ┌─ big.eu:2:9
2 │ pad0: 0 + 0
  │         ^
  = stack trace:
    - pad0 at big.eu:2:9
    - pad455 at big.eu:457:9
    - pad470 at big.eu:472:15
```

A primary label and two trace frames blaming arbitrary lines the user never called, for an error raised inside `nth`. This one is the more serious of the two and was not previously known.

## What changed

- **`SourceMap::global_slot_info`** — a second resolution step both renderers now take for global-slot Smids. Deliberately *not* folded into `source_info_for_smid`: every primary-location selector (`first_source_smid`, `first_user_source_smid`, `source_info`) routes through that, and must keep rejecting these, because a prelude declaration site is never a valid primary label (invariant (i)). There is a unit test pinning that separation.
- **`PreludeBlob::binding_spans`** — each binding's declaration span in `lib/prelude.eu`, in slot order, captured by `xtask` from the peeled body's own Smid (349/352; the three misses are the `__build`/`__io`/`__args` pseudoblocks, which have no prelude declaration and record `None`).
- **Prelude source registered lazily.** The text is embedded in the binary either way; it is added to `files` (and marked a resource file) only when a diagnostic is actually being rendered, so the happy path pays nothing.
- **`StgArena::rebase_baked_smid`** — rebases `DirectApp`/`LookupLit` Smids onto the enclosing global's slot identity, the one thing about them that is true at runtime. The plain `reconstruct_form` path (pretty-printer, `__args`/`__io` overrides) is untouched: those forms were flattened in-process, so their Smids are genuine.
- **Blob wire format 5 → 6** (`build.rs` + `xtask`), for the new field and the changed baked-Smid values.
- `EU_ERROR_TRACE_DUMP=1` now reports the `SourceMap` entry count — the difference between "this Smid is meaningless" and "this Smid is meaningful but belongs to something else", which is what pinned defect 2.

Result — blob mode now matches the source-compiled path:

```
blob:   - result at nth_out_of_range.eu:2:12
        - in 'nth' at [prelude]:1386:1
source: - result at nth_out_of_range.eu:2:12
        - in 'nth' at [prelude]:1391:3
```

## One test assertion changed, and why

`debug_trace_restores_the_uncurated_trace` asserted that `nth`'s boundary frame is *absent* from the raw `--debug-trace` stack dump. That is a property of the **source-compiled** prelude, where the inliner folds `nth`'s recursion into its caller and the surviving continuations are annotated with `nth`'s inner `aux` helper (Transparent). Under the blob, `nth` is a real global call and its own frames legitimately reach the raw dump — blob mode carries per-binding identity, so they are labelled `nth`.

Per its own doc comment, that line is the *precondition* explaining why `curate_trace_with_env`'s env-recovery path exists, not the property under test. It is now asserted against `--source-prelude`, which **pins** the configuration that exercises that code path instead of leaving it to whether a blob happens to be present — so the guard now runs identically in both CI configurations, where before it only ever ran in one. The outcome assertions that follow (curated trace names the boundary; keeps the user anchor) are unchanged and unconditional, and they are what the three target tests turn on.

No other assertion was relaxed. `corpus_satisfies_invariants` passes untouched in every configuration.

## Verification

All three target tests pass **with** a blob and **without** one, on **both** engines:

| | default (bytecode) | `EU_HEAPSYN=1` |
|---|---|---|
| **with blob** | 2134 passed, 0 failed | 2134 passed, 0 failed |
| **without blob** | 2124 passed, 0 failed | 2124 passed, 0 failed |

(Full `cargo test --no-fail-fast`, four separate runs. The 10-test delta is the `#![cfg(prelude_blob_ok)]` files, which compile to zero tests without a blob — see the eu-5yn9v note below.)

`cargo fmt --all` clean; `cargo clippy --workspace --all-targets -- -D warnings` clean in both blob and blob-less builds.

### Fault injection — done, twice

**Injection 1** (revert defect 1: make `resolve_trace_entry`/`json_trace` skip `global_slot_info`), blob present:

- `secondary_labels_do_not_excerpt_library_internals` → FAILED
- `debug_trace_restores_the_uncurated_trace` → FAILED
- `test_193_1tkk_7_12_curated_trace` → FAILED
- `blob_mode_curated_trace_names_the_boundary_combinator_with_a_prelude_location` → FAILED
- `blob_mode_trace_frames_cannot_alias_unrelated_user_declarations` → FAILED
- 3 `SourceMap` unit tests → FAILED

**Injection 2** (revert defect 2: copy `DirectApp`/`LookupLit` Smids verbatim, blob regenerated):

- `reconstructed_prelude_globals_carry_no_baked_source_smids` → FAILED
- `blob_mode_trace_frames_cannot_alias_unrelated_user_declarations` → FAILED
- `debug_trace_restores_the_uncurated_trace` → FAILED
- `reconstruct_form_annotated_rebases_directapp_and_lookuplit_smids` → FAILED

Both reverted, blob regenerated, all PASS again; working tree verified identical to the commit afterwards.

### New gates

In `tests/diagnostics_blame_plumbing_test.rs` (the file the blob-mode CI job already runs in full):

- `reconstructed_prelude_globals_carry_no_baked_source_smids` — structural invariant over the whole embedded blob: every Smid reachable from a prelude global is invalid or a global-slot identity. Asserting the invariant rather than the two known fields means a future `ArenaStgSyn` variant carrying a Smid cannot reintroduce the class silently.
- `blob_mode_curated_trace_names_the_boundary_combinator_with_a_prelude_location` — end-to-end rendering.
- `blob_mode_trace_frames_cannot_alias_unrelated_user_declarations` — end-to-end aliasing, on a generated 900-declaration fixture. It asserts the *count* of user-file frames (exactly one anchor), not which declaration the anchor names, deliberately: see the separate finding below.

Plus `SourceMap` and `StgArena` unit tests, which run in blob-less builds too, so the plain "Test Suite" CI job gates this change as well.

## Notes for the reviewer

- **Recorded review required** — this touches the blob wire format. I am not merging it.
- **Separate pre-existing bug found, not fixed here:** in a unit with many bindings the single user trace anchor names the *wrong* declaration (`pad0` rather than `result`), and it reproduces identically under `--source-prelude`, so it is prelude-independent and out of scope for eu-7x0r. Looks like a stale continuation annotation from the first rendered block item. Reported to the coordinator to file.
- **Cargo gotcha worth knowing** (relevant to eu-5yn9v): restoring `lib/prelude.blob` with an *older* mtime than the last build does not retrigger `build.rs`, so `prelude_blob_ok` stays off and `#![cfg(prelude_blob_ok)]` files silently compile to zero tests. `touch build.rs` forces it. This is exactly the "absent coverage is indistinguishable from passing coverage" failure mode eu-5yn9v wants closed.


gmorpheme:
## Recorded review (Wicket) — APPROVE for the owner's merge decision

Reviewed in an isolated worktree (`/tmp/eu-wicket-1079`, branch `pr1079` off
`origin/master` a1af1f0b). **Not merged** — this touches the blob wire format,
so the merge decision is the owner's; this comment is the recorded non-author
review that CLAUDE.md's gate requires.

Verdict: **the diagnosis is correct, the fix is correct, and the tests genuinely
gate.** Both defects reproduce on unmodified master and are eliminated by this
branch. I found no blocking issue. Four non-blocking findings and one
release-sequencing action are below.

### CI

`gh pr checks 1079`: **18 pass, 5 skipping (release jobs), 0 fail, 0 pending**
(run 30220306590). The author reported CI still running; it has since completed
green, including `Test Suite (windows-latest)`.

### What I verified independently (not taken on trust)

**Defect 2 reproduced on unmodified master.** Built my own fixture (2000
`filler<i>: <i> + <i> * 3 - 1` declarations plus `boom: xs nth(99)`, distinct
from the PR's 900-`pad` fixture) and ran it against the master worktree's
binary with a blob:

```
error: panic: index 99 out of range for list of length 3
    ┌─ big1.eu:237:18
237 │ filler235: 235 + 235 * 3 - 1
    │                  ^^^
    = stack trace:
      - filler0 at big1.eu:2:16
      - filler227 at big1.eu:229:24
      - filler235 at big1.eu:237:18
```

A primary label and three trace frames blaming lines the user never called, and
no prelude combinator named at all. On this branch the same fixture yields one
user frame plus `- in 'nth' at [prelude]:1386:1`. This is a genuine user-facing
blame bug, correctly characterised.

**Fault injection, both defects, repeated from scratch (blob regenerated,
`touch build.rs`, rebuilt):**

- *Defect 2* (`smid: *smid` restored in both arms): FAIL —
  `reconstructed_prelude_globals_carry_no_baked_source_smids`,
  `blob_mode_trace_frames_cannot_alias_unrelated_user_declarations`,
  `debug_trace_restores_the_uncurated_trace`,
  `arena::tests::reconstruct_form_annotated_rebases_directapp_and_lookuplit_smids`.
  My own fixture also regained all three bogus frames.
- *Defect 1* (`global_slot_info` step removed from `resolve_trace_entry` and
  `json_trace`): FAIL —
  `blob_mode_curated_trace_names_the_boundary_combinator_with_a_prelude_location`,
  `blob_mode_trace_frames_cannot_alias_unrelated_user_declarations`,
  `secondary_labels_do_not_excerpt_library_internals`,
  `debug_trace_restores_the_uncurated_trace`,
  `harness_test::test_193_1tkk_7_12_curated_trace`, and three `SourceMap` unit
  tests.

Both restored; tree confirmed identical to `HEAD` afterwards. No test in this PR
is inert.

**All four configurations, from a clean blob and `touch build.rs` each time:**

| | default (bytecode) | `EU_HEAPSYN=1` |
|---|---|---|
| with blob | **2034 passed, 0 failed** | **2034 passed, 0 failed** |
| blob-less | **2024 passed, 0 failed** | **2024 passed, 0 failed** |

`cargo fmt --all --check` clean. `cargo clippy --workspace --all-targets -- -D
warnings` clean in *both* the blob and blob-less builds. The 10-test delta is
exactly the `#![cfg(prelude_blob_ok)]` coverage (5 in
`diagnostics_blame_plumbing_test.rs`, 5 in `blob.rs`), matching the PR body.
I confirmed the plumbing file reported `running 5 tests` under a blob and
`running 0 tests` without — i.e. the mtime trap was avoided.

*One discrepancy, immaterial:* my totals are 2034/2024, the PR body says
2134/2124 — exactly 100 apart in both configurations. The *delta* matches
exactly, so the pair was clearly measured, but the absolute figures do not
reproduce here (darwin/aarch64, debug). Worth a glance at how they were
counted; nothing turns on it, since both runs are 0-failure.

**Diagnostics quality on the shipped path (not just green tests).** Blob and
`--source-prelude` now agree, and blob's location is arguably the better of the
two — `[prelude]:1386:1` is `nth`'s declaration head, where source mode's
`1391:3` is `}.aux(n, l)`:

```
blob:   - result at nthcase.eu:2:12 / - in 'nth' at [prelude]:1386:1
source: - result at nthcase.eu:2:12 / - in 'nth' at [prelude]:1391:3
```

The spans are safe by construction: `lib/prelude.blob` and `lib/prelude.eu` are
both `include_*!`d from the same tree, and `cfg(prelude_blob_ok)` requires the
hash to match, so the text a span indexes is always the text it was taken from.

### A — the changed test assertion: reasoning sound, coverage improved

I checked the author's claim rather than accepting it. Under `--source-prelude`
the raw `--debug-trace` frames are `[('user','result')]` and the curated trace
is `[('user','result'),('boundary','nth')]` — so the new assertion is
non-vacuous and the env-recovery path it documents is genuinely exercised. Under
a blob the raw dump legitimately carries four `('boundary','nth')` frames (the
rebased inner `aux` continuations). The original assertion was therefore not a
weaker check under a blob — it was *false*, and asserting it would have pinned
a falsehood.

The line is a precondition, not the property under test: the property
(`--debug-trace` bypasses curation) is carried by the `swap_args` block, which
is unchanged, unconditional, and still runs in the default configuration. The
change makes the guard run in both CI configurations where it previously ran in
one. `corpus_satisfies_invariants` is untouched, and the whole diff of
`tests/diagnostics_invariants.rs` is the `run_debug_trace_with` helper plus that
one block — nothing else relaxed. And as the fault injection above shows, the
test still FAILs if either defect is reintroduced. **Accepted.**

### B — completeness of the defect-2 fix: complete, with one cosmetic gap

Audited `ArenaStgSyn` directly. Exactly three variants carry a `Smid`:
`DirectApp`, `Ann`, `LookupLit`. `Ann` is elided at `arena.rs:483`; the other
two are now rebased. Plus `ArenaLambdaForm::Lambda.annotation`, already
replaced. `Ref`/`Native` carry no `Smid`. So the fix is complete at the arena
level, and
`reconstructed_prelude_globals_carry_no_baked_source_smids` is a real
class-level gate rather than a two-field spot check.

Audited every `reconstruct*` caller. The two unannotated `reconstruct_form`
calls on *override* arenas (`runtime.rs:184`, `:304`) act on forms flattened in
this process at `runtime.rs:161`, so leaving their Smids alone is right, as
claimed.

One gap the invariant test cannot see: `runtime.rs:235` (the pretty-printer)
reconstructs the **blob** arena with `global_annotation = None`, so raw
xtask-era Smids survive there. Harmless today — `pretty.rs:127` and `:228`
destructure `DirectApp`/`LookupLit` with `..` and never print the smid — but it
is the one path that would silently regain the class if the pretty-printer ever
started rendering them. Non-blocking; consider passing
`Smid::global_slot(i)` there too, purely so no reader of a blob arena can get
it wrong.

### C — real diagnostics, not proxy tests

Confirmed above: blob mode now names the responsible combinator and anchors in
the user's file, matching `--source-prelude`. The design invariant is preserved
in the right place — `global_slot_info` is deliberately outside
`source_info_for_smid`, so `first_source_smid` / `first_user_source_smid` keep
rejecting prelude declaration sites as primary labels, and the lazily-registered
prelude file is marked a resource file. Both pinned by unit tests. The docs in
`docs/reference/cli.md` §"Error Traces" already described this behaviour, so no
doc change is needed — the docs were right and the implementation wasn't.

### D — wire-format collision (action required before merge)

Both copies are updated in this PR: `build.rs:45` and `xtask/src/main.rs:59`,
5 → 6, with the version history comment extended in both. Confirmed.

**PR #1078 bumps the same constant to the same value 6 in the same two files.**
Master is 5. Whichever of #1078/#1079 merges second must move to **7** and
update **both** copies — they are kept in sync only by a doc comment.
Mechanical enforcement is tracked as eu-3skeg.

### F — red herring and scope

The previously pinned root cause (eu-n8c5e, xtask skipping the inline pass) is
confirmed a red herring for these three tests: under a blob the boundary frame
was present in the stack trace all along and merely unrenderable — I saw it
resolve once the renderer step was restored, with no change to the compile
pipeline. The dependency edge can be dropped.

**eu-og3u6 is genuinely out of scope, and I confirmed it is not this bug in
disguise.** On the *master* worktree with `--source-prelude` (i.e. nothing from
this PR involved), my 2000-declaration fixture anchors on `filler0` rather than
`boom`. Prelude-independent, pre-existing, different mechanism. But note the
user-facing consequence: in a realistic multi-declaration file the primary label
still points at a declaration the user did not call. This PR removes the
*prelude-side* aliasing and reduces three bogus frames to one; the one that
remains is eu-og3u6. That deserves priority on the 0.14 diagnostics path — the
"blame is correct now" story is not complete without it.

### G — eu-5yn9v can widen all the way (on this platform)

Confirmed on my own runs: the entire `cargo test` suite passes under a blob on
both engines, **including all 172 `tests/harness/errors/*.eu` tests with
`.expect` sidecars** (172 run, 0 failed, under both `blob+bytecode` and
`blob+heapsyn`). So the allowlist in the `diagnostics-blob-mode` job and the
blob-less default of the `Test Suite` job can both be inverted once this lands.
Caveat on scope: I verified darwin/aarch64 debug only — the widening PR will
need its own CI run to confirm ubuntu and windows.

### Non-blocking findings for follow-up beads

1. **The same aliasing class remains open on the typecheck path.**
   `PreludeBlob::desugared_unit_cores` carries `RcExpr` cores full of Smids
   minted by xtask's own `SourceMap`, injected into a runtime `SourceMap` by
   `run_type_checker_from_blob_core`. `check.rs:1030`
   `in_prelude_smid_degrades_gracefully_not_misleadingly` asserts that
   degradation is graceful — but it does so with *"a `Smid` far outside any
   range this test's tiny `SourceMap` will ever register"*, which is precisely
   the small-map reasoning defect 2 has just disproved. It looks latent today
   (the corpus sweep shows every live diagnostic citing a user call site), but
   the argument for its safety is no longer sound. Worth a bead to either
   normalise those Smids at bake time or re-do that test against a large
   `SourceMap`.
2. **Pretty-printer path** — see B above.
3. **The 900/2000-declaration aliasing regression is gated only by a
   `#![cfg(prelude_blob_ok)]` Rust test.** Correct for a blob-only bug, and the
   arena unit test covers the mechanism in blob-less builds — but an
   `errors/*.eu` + `.expect` fixture would put it in the harness suite that
   `eu test` runs on the release binary too. Optional.
4. **`binding_spans` can silently degrade to all-`None`** if
   `loader.file_id_for(&locator)` ever returns `None`, leaving prelude frames
   name-only. In practice
   `blob_mode_curated_trace_names_the_boundary_combinator_with_a_prelude_location`
   requires a rendered `[prelude]:line:col`, so it *is* gated end to end — noted
   only so the xtask `349/352` progress line is understood as informational,
   not the gate.

### Design note

Rebasing at read time (rather than normalising at bake time) is the right call
here, and I want that on the record: zeroing these Smids in the blob would
delete the `ApplyTo` call-site annotation entirely and lose frames, whereas
rebasing keeps the one fact that is true at runtime — which prelude global the
code belongs to. The cost is that call sites *within* a prelude combinator all
collapse to one identity, which is why the raw blob dump shows `nth` four times;
that information was garbage before, so nothing real is lost, and the curated
trace collapses recursion anyway.

I am not merging. Over to the owner.

PR #1080, opened 2026-07-26
fix(diagnostics): mark a baked-in library served by filename as a resource (eu-8a49h)

gmorpheme:
Fixes **eu-8a49h** (P2).

Split out of #1078 on the coordinator's instruction: this changes diagnostics behaviour for **every** shipped library, not just contracts, so it deserves its own focused review — and #1078 already carries a recorded non-author approval that folding a cross-cutting change into would invalidate.

## The bug

A shipped library imported by filename was classified as **user code**, so a frame inside it could win `to_diagnostic`'s "prefer a user-file location" test and become the **primary label**:

```
# probe.eu
{ import: "reflect.eu" }
bad: [:t-record, 42] from-data
```

**Before** — blame lands in the library, thirty lines of internals excerpted, the user's own call site demoted to a trace note:

```
error: panic: elements called on non-block
   ┌─ reflect.eu:74:5
   │
21 │     ╭ type-str(td):
22 │     │   { prim-names: {
   ·     │
42 │   ╭ │     record-str(rec): rec elements map(field-str) str.join-on(", ") "{{ {} }}"
   │   │ │     -------------------------------------------------------------------------
   ·     │
   = stack trace:
     - bad at probe.eu:5:22            ← the only mention of the user's code
     - from-data at reflect.eu:83:17
```

**After** — blame is the user's call:

```
error: panic: elements called on non-block
  ┌─ probe.eu:5:22
  │
5 │ bad: [:t-record, 42] from-data
  │                      ^^^^^^^^^
```

## Root cause

`mark_resource_file` fires only for `Locator::Resource`, and `Locator::Resource` is produced **only** by an explicit `resource:` prefix (`src/syntax/input.rs`). The ordinary form `{ import: "reflect.eu" }` yields `Locator::Fs`; `read_fs_input` then serves the **baked-in resource text** while leaving the locator `Fs`. So the file is never marked, and `is_user_file` reports `true` for a file the user has never seen.

The resource mechanism was not missing — it was **bypassed**.

## The fix

`read_fs_input` now reports whether it fell back to a baked-in resource; `load_source` marks the file when it did. ~15 lines in `src/driver/source.rs`.

### The subtle half, which a reviewer must agree with

A library the user has **shadowed on disk** is deliberately **not** marked:

```rust
if matches!(locator, Locator::Resource(_)) || from_resource {
    self.source_map.mark_resource_file(id);
}
```

`from_resource` is false when the file was found on the lib path. The reasoning: once a copy is on disk, the user is *editing* it, and a location inside it is exactly what they want blamed. Only the shipped copy is off-limits. Both halves carry a unit test (`test_read_fs_input_flags_the_baked_in_resource_fallback`, `test_read_fs_input_does_not_flag_a_library_shadowed_on_disk`).

## Tests

- `tests/harness/errors/194_8a49h_library_blame.eu` — SV3-independent, using `reflect.eu`.
- Two unit tests in `driver::source::tests` for the two halves of the disk-shadow judgement.

**The regex deliberately anchors on the `┌─` primary-label line**, not merely on the fixture's name appearing somewhere:

```
stderr: "(?s)elements called on non-block.*┌─[^\n]*194_8a49h_library_blame\.eu:\d+:\d+"
```

Before the fix the user's file *still appeared* in the `stack trace:` note, so a looser pattern passes either way and gates nothing. I checked that specifically, because it is the same trap that let the original defect through review.

## Fault injection — done

Reverting just the `|| from_resource` clause:

```
test test_194_8a49h_library_blame ... FAILED
   FAIL: stderr did not match pattern '(?s)elements called on non-block.*┌─[^\n]*194_8a49h_library_blame\.eu:\d+:\d+'
```

Restoring it:

```
test test_194_8a49h_library_blame ... ok
```

## Verification that no *useful* blame was removed

This was asked for explicitly, so it was checked deliberately rather than inferred from a green suite.

**1. No error test asserts a library location.** Every location-asserting `.expect` in `tests/harness/errors/` refers to its *own* file; grep for `test.eu` / `markup.eu` / `reflect.eu` / `lens.eu` / `state.eu` in the sidecars returns **zero** matches. Nothing was relying on a library frame.

**2. Before/after comparison on the two libraries most likely to matter** — output **byte-identical** in both cases, because the user frame was already winning:

| Case | Before | After |
|---|---|---|
| failing `//=` expectation through `lib/test.eu` (`eu test`) | `probe-testeu.eu:4:7` | identical |
| `[] tag` raising inside `lib/markup.eu` | `probe-markup.eu:5:9` | identical |

The change is observable **only** where a library frame would otherwise have *won* the primary label — i.e. only where it was wrong. `EXPECT FAILED` reporting through `test.eu` is unaffected.

**3. Full suite, both prelude modes:**

| Mode | Result |
|---|---|
| source-prelude | **zero failures** across every test binary |
| bytecode + blob | see the comment below |

## Relationship to other work

- **Found while implementing SV3 (eu-u9xj.1, #1078).** There the misclassification was **mode-dependent** and so easy to miss: in source-prelude mode the frame at the head of the stack trace happened to be a properly-marked *prelude* frame (`foldr`), so blame landed correctly; in **blob mode — the shipped path** — the head frame was `lib/contract.eu` and blame moved into the library. #1078's `Smid::default()` choice was necessary but **not sufficient**: it removed the error's own Smid from contention, while the trace search still ranked library frames as user frames.
- **#1078 is sequenced after this.** It notes that it does not yet conform to its own spec §9.1 in blob mode, and the primary-label anchor for its contract error lands once this merges.
- This is also the root cause of the secondary-label noise wicket observed (library internals excerpted into diagnostics) — a `diagnostics_invariants` concern independent of SV3.
- **No wire-format change**, so no interaction with the `BYTECODE_WIRE_FORMAT_VERSION` contention between #1078 and #1079.

## Gates

- `cargo fmt --all` — clean
- `cargo test` — see above


gmorpheme:
**Full-suite results, both prelude modes** (the table in the body defers to this).

```
source-prelude:   zero failures across every test binary
bytecode + blob:  harness 506 passed / 1 failed
                  diagnostics_invariants 1 passed / 2 failed
```

Those three blob-mode failures are **pre-existing on unmodified master** and tracked as **eu-7x0r** — `test_193_1tkk_7_12_curated_trace`, `secondary_labels_do_not_excerpt_library_internals`, `debug_trace_restores_the_uncurated_trace`. I verified they reproduce on `a1af1f0b` with a locally generated blob before starting, and this change neither fixes nor worsens them. The blob-mode CI job carries deliberate exclusions for exactly those, which is why CI is green on master.

Worth flagging for whoever picks up eu-7x0r: `secondary_labels_do_not_excerpt_library_internals` is *thematically* the same problem this PR fixes (library internals reaching a diagnostic), so it may look like this PR should have fixed it. It does not — that test asserts a `[prelude]` mention count in the **curated trace note**, which is the eu-1tkk.7.11 blame-table path, not the primary-label ranking this PR touches. Two distinct mechanisms with one symptom.

Also note the harness count: 506 passed here against master's 505, the difference being the one test this PR adds.

gmorpheme:
## Recorded review — wicket

**Verdict: APPROVE.** Correct fix, right judgement, genuinely gating test. I am not merging: merge decisions on this one sit with the owner.

Reviewed in an isolated worktree at `4055cb57`. Everything below marked *verified* I reproduced myself; everything marked *taken on trust* I did not.

### Gates

| Gate | Result |
|---|---|
| `gh pr checks 1080` | 19 pass / 5 skipping / 0 fail — **verified** |
| `cargo fmt --all --check` | clean — **verified** |
| `cargo clippy --workspace --all-targets -- -D warnings` | clean — **verified** |
| full suite, source-prelude | zero failures across every binary (harness 507/507) — **verified** |
| full suite, blob mode | harness 506 pass / 1 fail (`test_193_1tkk_7_12_curated_trace`); `diagnostics_invariants` 1 pass / 2 fail — **verified**, and **verified pre-existing** by building master `a1af1f0b` with a locally generated blob and reproducing all three |

---

### A. Is the disk-shadow judgement right, and is the discrimination sound?

**Agreed, and the discrimination is sound by construction.** The flag is computed at the same point as the text it describes: `true` is returned only from the `self.resources.get(stem)` branch, `false` only from the two `fs::read_to_string` branches. There is no separate predicate that could drift out of step with the content, so neither failure direction is reachable — a user's own file can never be marked a resource, and baked-in text can never escape marking.

One piece of evidence in favour of *provenance* as the discriminant that the PR body does not claim, and which I think settles the design question: `test_lib_lens` and `test_lib_markup` (`tests/harness_test.rs:936`, `:951`) deliberately run a shipped library **as the top-level subject** with `lib` on the lib path. There the library *is* the user's code and must remain blamable. Provenance keeps that working; a discriminant based on filename or stem would have silently broken both. The chosen rule is not merely defensible, it is the only one of the obvious candidates that survives the repo's own tests.

Residual hazard, **pre-existing and not worsened here**: a user file named after a baked-in resource (`test.eu`, `lens.eu`, `state.eu`, `reflect.eu`, `markup.eu`, or `package`/`build-meta` by stem) that is *not* found on the lib path or relative to cwd is silently served the **library text** instead of raising an import error. That is the fallback's own behaviour, untouched by this PR; if anything the PR makes it marginally more discoverable, since diagnostics now decline to blame a line inside the substituted text. Worth a separate bead, not a blocker here.

### B. Does this remove blame that was legitimately useful? — verified, and better than reported

I did not take the before/after table on trust. I fault-injected the fix and diffed real output in **both** prelude modes:

- `import: "test.eu"` with a failing `//=` — **byte-identical**, `EXPECT FAILED` line intact.
- `import: "markup.eu"` raising `head of empty list` — **byte-identical**.
- `import: "lens.eu"` calling `view(:not-a-lens)` — **not identical: it improves.** Before, the primary label was `lens.eu:17:24` (`view(lens, data): data lens(...)`); after, it is the user's own `view` call site. This case is not in the PR body and is a further win.

Also worth recording, because it bounds the risk on `test.eu` more tightly than the PR body does: `eu test` loads the report generator via a genuine `Locator::Resource("test")` (`src/driver/tester.rs:746`, `:823`), so the harness's own reporting path never went through the newly-marked route at all. My probe used `{ import: "test.eu" }`, which *does*, and was unchanged.

The grep claim (zero `.expect` sidecars asserting a library location) I **took on trust**; the full-suite result in both modes is the stronger check and it is clean.

### C. Does the new test's regex genuinely gate? — verified independently

Reverting **only** the `|| from_resource` clause and rebuilding:

```
test test_194_8a49h_library_blame ... FAILED
```

Restoring it (`git diff --stat` empty, so the restore is byte-exact): `... ok`.

I also confirmed the trap the author describes is real rather than hypothetical. The fault-injected output blames `reflect.eu:61:18` and excerpts ~20 lines of library internals — **yet still contains `194_8a49h_library_blame.eu:12:22`** in its `stack trace:` note. A regex merely looking for the fixture name would have passed in both states and gated nothing. The `┌─`-anchored form is doing real work.

### D. Completeness — no other unmarked resource path

`self.resources.get` has exactly two call sites: the existence check at `src/driver/source.rs:597` and the fallback at `:1243`. `Locator::Resource` is constructed only from an explicit `resource:` prefix (`src/syntax/input.rs:60`) or a `resource://` URL (`:143`) — confirming the bead's mechanism as filed. The only other `mark_resource_file` call is `src/wasm_pipeline.rs:118-119`, which already marks the prelude it registers. `read_fs_input` is therefore the sole route by which baked-in text reached a `SourceMap` unmarked, and it is now covered. The discarded flag at `:579` (`detect_and_load_prelude_override`) is correct — that call reads text to look for a `prelude:` key and registers no file.

### E. The `secondary_labels_do_not_excerpt_library_internals` claim — confirmed, with a sharper reason

The conclusion is right; state it in the terms below rather than the PR's, because they are decisive. Both corpus fixtures — `nth_out_of_range.eu` (`xs nth(10)`) and `metadata_span.eu` — **import no library at all**. They exercise only the prelude, which was already marked via `Locator::Resource("prelude")`. This PR is therefore *structurally incapable* of changing them, whatever mechanism the test asserts on. Verified: the failure message is identical on master and on this branch —

```
nth_out_of_range.eu: expected exactly 1 '[prelude]' mention(s) ..., found 0
```

Note the direction: the `[prelude]` mention is **missing**, not duplicated. That is blob-mode Smid resolution, which is #1079's subject, not library misclassification.

One correction to the PR's framing, worth having on the record: `SourceMap::classify_frame` **does** consult `is_user_file` (`src/common/sourcemap.rs:390`), and `to_diagnostic` now routes through `curate_trace` (`src/eval/error.rs:1073`). So this PR *does* change curated-trace classification for filename-imported libraries in general — "two mechanisms, one symptom" understates the coupling. It just cannot affect *these two fixtures*.

### F. Interaction with #1079 — none

#1079's only `src/driver/source.rs` change is an additive `file_id_for` getter near line 1240; this PR touches 576–720 and 1214–1245. **Verified**: `master` + #1080 + #1079 merges with **no conflict**, in that order. Mechanisms are orthogonal — #1080 decides *which file ids are resources*, #1079 gives blob-mode prelude frames resolvable identity. No wire-version change here, so no `BYTECODE_WIRE_FORMAT_VERSION` collision. **Merge order does not matter functionally**; either sequence is safe.

---

### Design note, in the PR's favour

A reviewer might reasonably ask why the fallback does not simply rewrite the locator to `Locator::Resource(stem)`, letting the existing `matches!` test work untouched. That would be worse, for two concrete reasons: `Locator::Resource` displays as `[stem]` (`src/syntax/input.rs:112`), so the trace note would degrade from the informative `from-data at reflect.eu:83:17` to `[reflect]`; and `self.locators` is keyed on the locator, so a rewrite would poison the cache against the original `Fs` key. The flag keeps the filename visible to the user while fixing the classification. Right call.

### Non-blocking nits

1. **Fixed temp path in the shadow unit test.** `std::env::temp_dir().join("eu_8a49h_shadow_test")` (`src/driver/source.rs:1503`) is shared across concurrent runs — two `cargo test` invocations on one machine (debug and release, say) run the same test name against the same path, and the cleanup is also skipped entirely if an assertion panics. The same subsystem's newer tests already use a process-unique form: `format!("eu-blame-plumbing-{}-{n}", std::process::id())` (`tests/diagnostics_blame_plumbing_test.rs:48`, `tests/diagnostics_json_emission.rs:34`). Worth matching. Low severity — no observed flake.
2. **Optional polish.** `Result<(String, bool), _>` reads fine at two call sites, but a two-variant enum would name what the bool means at the call site, and would let `load_source` drop the `let mut from_resource = false;` plus in-arm assignment in favour of a direct binding.

### One thing for the record, not this PR's fault

The new error test is gated **only in source-prelude mode**. The "Bytecode + blob harness" job runs `./target/release/eu test --allow-io tests/harness`, which does not validate `.expect` sidecars under `tests/harness/errors/`, and the plain "Test Suite" job is deliberately blob-less. Given the SV3 manifestation of this bug was *found* in blob mode, that is worth noting — though the defect this test guards does reproduce in source mode too, so the test genuinely gates where it runs. I **verified** `test_194_8a49h_library_blame` passes in blob mode locally. Pre-existing CI-topology gap, tracked as eu-1tkk.7.19.

Not closing the bead — that is the coordinator's call.

gmorpheme:
Thank you — all three points actioned, and the optional suggestion taken.

## 1. The temp-path nit — fixed

Now matches the existing process-unique form in the same subsystem:

```rust
let dir = std::env::temp_dir().join(format!("eu-8a49h-shadow-{}", std::process::id()));
```

I also fixed the second half of what you spotted, which was the worse half: cleanup is now **unconditional**, taken before the assertions rather than after them, so a failing assertion cannot leave the scratch copy behind for the next run to trip over.

```rust
let result = loader.read_fs_input(Path::new("reflect.eu"));
let _ = fs::remove_dir_all(&dir);          // before asserting
let input = result.expect("the on-disk copy resolves");
```

## 2. The `lens.eu` improvement — verified and added

Reproduced exactly as you described, so it is now in the CHANGELOG entry alongside the `reflect.eu` case:

```
# data view(:not-a-lens)

BEFORE                                   AFTER
   ┌─ lens.eu:17:24                        ┌─ probe-lens.eu:4:11
17 │ view(lens, data): data lens(…)      4 │ bad: data view(:not-a-lens)
   │                        ^^^^           │           ^^^^
```

Before, `view`'s own body was excerpted and the user's call site appeared only as `- bad at probe-lens.eu:4:11` in the trace note. It does not overstate: the improvement is the same mechanism as the `reflect.eu` case, just a second, independent instance of it.

## 3. The enum — taken, and it paid for itself

`Result<(String, bool), _>` → `Result<FsInput, _>` with `UserFile(String)` / `BakedInResource(String)`.

As you predicted, `load_source` no longer needs `let mut from_resource = false;` plus in-arm assignment. Better than that, making every arm answer *both* questions let me **unify the two marking conditions** instead of or-ing them afterwards:

```rust
let (source, is_library) = match locator {
    Locator::Fs(path) => { … (input.into_text(), is_library) }
    Locator::Resource(name) => ( …, true ),
    Locator::Cli(text) => (text.to_string(), false),
    …
};
if is_library {
    self.source_map.mark_resource_file(id);
}
```

`if matches!(locator, Locator::Resource(_)) || from_resource` is gone. `Locator::Resource` now reports `true` where it is read rather than being special-cased at the marking site, and there is no way to add a locator arm that answers one question but not the other.

## 4. Your point 1 — recorded in the code, since it is the real argument

You found the evidence I should have led with: `test_lib_lens` and `test_lib_markup` (`tests/harness_test.rs:936`, `:951`) run a shipped library as the **top-level subject** with `lib` on the path, to exercise the library's own embedded `test-*` targets. There the library *is* the code under test and must stay blamable — so a name- or stem-based discriminant would have silently broken both, while provenance keeps them working.

That is now a doc comment on `FsInput`, naming both tests, so the next person to reach for the "just check the filename" shortcut meets the counter-example first. Both tests confirmed still passing.

## Re-validation after these changes

Fault injection re-run against the refactored shape (the equivalent revert is now "the `Fs` arm never reports a library"):

```
reverted:  test test_194_8a49h_library_blame ... FAILED
restored:  test test_194_8a49h_library_blame ... ok
```

| Gate | Result |
|---|---|
| `cargo fmt --all` | clean |
| `cargo clippy --workspace --all-targets -- -D warnings` | clean |
| `cargo test` source-prelude | **zero failures**, all 16 test binaries |
| `cargo test` bytecode + blob | see below |
| `test_lib_lens`, `test_lib_markup` | pass |
| `driver::source::tests` | 7 pass |

## 5. On the record correction

Understood, and thank you for chasing it — my "two mechanisms, one symptom" line was wrong in the direction that matters. `classify_frame` does consult `is_user_file` and `to_diagnostic` routes through `curate_trace`, so this PR **does** change curated-trace classification for filename-imported libraries in general. The narrow true statement is only that it cannot move those two eu-7x0r fixtures, because they import no library and exercise only the already-marked prelude. I have not repeated the overclaim anywhere else.

PR #1081, opened 2026-07-27
fix(meta): treat non-block metadata as opaque instead of merging (eu-3aa6s)

gmorpheme:
Fixes **eu-3aa6s**.

## The bug

```
spec: { :only a: 1 }
m: spec meta        # error: type mismatch: expected block, found symbol
r: spec raw-meta    # -> only   (works)
```

`META` (`src/eval/stg/meta.rs`) destructures one metadata layer, recurses
into the body and then `MERGE`s the two results — "careful: body itself may
have more metadata so merge...". `Merge` is a block operation, so the bare
symbol that the `{ :sym … }` shorthand attaches raised. `RAWMETA` never
merges and so was unaffected: the two accessors disagreed on whether the
very same value had readable metadata at all.

Reachable via the shorthand for any *unregistered* tag (`{ :only … }`,
`{ :foo … }`), and via the `//` operator with any non-block value
(`5 // :tag`, `5 // "text"`).

## Refutation: the `{ :io … }` form was never affected

The dispatch asked me to confirm the `:io` case first. **It does not fail —
and cannot.** `:io` is a *registered* monadic namespace, so
`extract_block_monad_spec_from_raw` matches it and the desugarer rewrites
the block into an `io.bind` chain; the tag is consumed at desugar time and
the resulting action carries no metadata at all.

```
blk: { :io r: io.return(42) }
r: blk raw-meta
---
r: {}                 # not a symbol — no metadata at all
```

`eu dump desugared` confirms it:

```
in let blk = io.bind(`{ __type_hint: "IO(a)" }
         (io . return (42)), λ(r). io.return(let r = r in { r: r }));
       r = (blk raw-meta)
```

The same holds for `:for`, `:let`, `:random` and `:state`. Only an
*unregistered* tag falls through the `ns_registered` check to plain block
desugaring and survives to the runtime as symbol metadata. The diagnosis is
otherwise exactly right — the mechanism is structural, and it does break
every `{ :sym … }` block whose tag is not a monadic namespace.

I kept a `:io` check in the regression test anyway, recording the
now-verified behaviour (`meta` succeeds, reports empty metadata), and the
test says explicitly that this check documents the form rather than gating
the fix.

## The fix

`META` now combines layers by a single rule:

| outer layer | accumulated inner | result |
|---|---|---|
| block | block | `merge(inner, outer)` — **unchanged**, outer wins on conflicts |
| block | non-block | the outer block, alone |
| non-block | anything | the outer layer, as-is (opaque) |

The recursion is untouched — `RAWMETA` still exists precisely because
metadata values can themselves carry metadata, and stacked block layers
still merge exactly as before.

### The mixed case, decided deliberately

A non-block layer is **opaque: it terminates the merge chain rather than
being skipped over.** Both directions therefore return the outermost layer
alone:

```
5 // { p: 1 } // :tag    meta  # -> :tag      (symbol over block)
5 // :tag // { p: 1 }    meta  # -> { p: 1 }  (block over symbol)
```

The alternative — filtering non-block layers out and merging the rest —
would make an opaque layer semi-transparent, so that a `{ q: 2 }` buried two
layers down could resurface in the answer through a symbol layer the user
wrote to *replace* it. "Outer wins" is the established precedence for block
layers; extending it so that an unmergeable layer replaces the accumulated
view wholesale is the consistent reading. It is documented on the `Meta`
struct, on the `meta`/`raw-meta` prelude doc strings, and in
`docs/reference/prelude/supplements/metadata/default.md`.

Implementation is a nested `case` on the `Block` tag — `case` (not
`switch`), so a raw native metadata value lands in the fallback rather than
raising `NoBranchForNative`. Blocks are always the `Block` data constructor
(there is no `Native::Block`), so the tag test is exact.

## Regression test — `tests/harness/191_3aa6s_meta_symbol_metadata.eu`

11 checks, all folded into `RESULT` via `all-true?`:

| | check |
|---|---|
| t1 | shorthand symbol metadata returned as-is instead of raising |
| t2 | `meta` now agrees with `raw-meta` on it |
| t3 | a single block layer still reported in full |
| t4 | **stacked block layers still MERGE**, outer wins on the conflicting key |
| t5 | the merge is a merge, not a replacement — all three keys survive |
| t6 | `raw-meta` still reports only 2 keys, so t4/t5 genuinely exercise the recursion |
| t7 | backtick metadata over a `//` layer merges the same way |
| t8 | opaque symbol layer over a block stands alone |
| t9 | block layer over an opaque symbol stands alone |
| t10 | no metadata still reports the empty block |
| t11 | the `{ :io … }` form: `meta` succeeds, reports empty metadata |

### Fault-injection verified — five separate injections

Each was built and run; **restored and re-run PASS after every one.**

| injection | result |
|---|---|
| revert `meta.rs` entirely (stash the fix) | **FAIL** — `t1` raises `expected block, found symbol` |
| drop the merge: `Merge.global(lref(4), lref(5))` → `local(4)` | **FAIL** — `t3` raises `key 'a' not found in block` |
| discard the opaque outer layer: fallback → `KEmptyBlock.global()` | **FAIL** — t1, t2, t8 false |
| inner opaque layer wins instead of outer: `local(4)` → `local(0)` | **FAIL** — `t9` raises `expected block, found symbol` |
| swap merge order: `Merge.global(lref(5), lref(4))` | **FAIL** — `t4` false (inner wins the conflict) |

## Validation

- `cargo fmt --all` — clean
- `cargo test` — **all green**: 1353 lib + 507 harness + 111 + 12 + 11 + rest,
  0 failed
- `EU_HEAPSYN=1 cargo test --test harness_test` — 507 passed, 0 failed
- `EU_GC_VERIFY=2 EU_GC_STRESS=1 cargo test --test harness_test` — 507 passed,
  0 failed
- Both engines byte-identical on the new test and on the mixed-layer probe
  (`diff` of default vs `EU_HEAPSYN=1` output: identical). Also passes under
  `EU_PREDECODE=0`.
- `python3 scripts/test-doc-examples.py` — 249 passed, 0 failed (the two new
  doc examples are live, not `notest`)
- `eu doc --prelude` regenerated (`docs/reference/prelude/metadata.md`), so
  the prelude-docs CI gate stays green

## Filed while here — eu-frf2o (pre-existing, out of scope)

A metadata-annotated **inline** block literal used as a catenation/call
operand raises `infinite loop detected: binding refers to itself`:

```
p: { :foo k: 1 } raw-meta      # error: infinite loop detected
q: { k: 1 } raw-meta           # -> {}    (no annotation: fine)
spec: { :foo k: 1 }
r: spec raw-meta               # -> foo   (bound first: fine)
```

This hits `raw-meta` identically, and **I reproduced it on master
(a1af1f0b) with an unmodified `meta.rs`** — it is not caused by this change.
The cooked core looks correct, so the fault is downstream of `cook`. Filed
as **eu-frf2o**; test 191 binds its subjects to names first and carries a
comment pointing there.

---

I have not merged this and will not — please route review as the coordinator
sees fit.


gmorpheme:

Reviewed in an isolated worktree at `8c8a4a72` (this branch merged onto master
`1bd8bf6f`). All gates run there, not on the author's machine.

### The fix itself — checked by hand, not by reading the summary

`__META`'s STG body becomes a four-arm nested `case`, and the whole correctness
of it is de Bruijn arithmetic, so I walked every arm index by index rather than
trusting the shape.

- **Both layers blocks** → `Merge.global(lref(4), lref(5))`. This is the
  pre-existing case, and the argument order is `Merge(inner-meta, meta)` —
  identical to before the change. Outer-wins on key conflict is therefore
  preserved, and that case has **no behavioural delta at all**.
- **Outer block, inner opaque** → `local(4)` = the outer metadata.
- **Outer opaque** → `local(0)` = the outer layer itself, unrecursed.

The `case` branches each push a frame binding the constructor's fields, and the
indices above are consistent with that in every arm. I checked the frame depth
at each `lref`/`local` site individually.

**The behaviour delta is pure repair.** Every case this change newly reaches
previously *raised* `expected block, found symbol`. Nothing that used to return
a value returns a different value. That is the property that makes this safe to
land without an owner-filter escalation, and it is why I am comfortable merging
it on my own recorded review: it is defect repair, not an intentional semantic
change.

I did not take the author's word for the test gating. Reverting
`src/eval/stg/meta.rs` to master's version in my own worktree and rebuilding:

```
---- test_204_3aa6s_meta_symbol_metadata stdout ----
204_3aa6s_meta_symbol_metadata...FAIL
test result: FAILED. 0 passed; 1 failed; 516 filtered out
```

Restoring the fix:

```
test test_204_3aa6s_meta_symbol_metadata ... ok
```

The test genuinely gates.

I also checked the fixture's own construction rather than just its verdict: all
11 checks feed the `all-true?` verdict, so none of them can pass silently, and
**t6** (`(stacked raw-meta) keys count = 2`) is doing real work — it is what
stops t4/t5 passing without the recursion actually being exercised.

### One change applied on top of the author's work

`lib/prelude.eu` annotated `meta` as `type: s"a → {..}"`. This fix makes that
annotation **false**: `{ :only k: 1 } meta` now returns the bare symbol `only`,
which is not a block. The PR updated the doc string sitting immediately beside
it but not the annotation.

Relaxed to `s"a → any"`. Two things worth stating so this is not mistaken for a
bigger change than it is: the annotation is currently **inert** — `eu check
--strict` on the fixture exits 0 with no output either way — and relaxing an
annotation can only ever *remove* warnings, never add them. It is a
truthfulness fix, not a behaviour one. But leaving a knowingly-false annotation
in the prelude is how the next person gets misled.

### Fixture renumbering

`191_3aa6s_meta_symbol_metadata` → **`204_...`**. 191-197 were claimed by the
SV3 contract fixtures that landed as 196-202 in #1078, and 203 is #1074's. I
confirmed no `191_3aa6s` reference survives anywhere in the tree.

### Conflict resolution

One conflict, `tests/harness_test.rs`, at the seam where master's five SV3
fixture functions meet this branch's. Resolved **by region**, with each function
keeping its own braces — a concatenate-both-sides resolution on a
brace-structured file silently swallows the trailing brace that the two sides
share, and that failure mode has bitten this queue repeatedly. Verified by
building, not by eyeballing.

### Docs

The new "Reading Metadata Back" section is in both `docs/reference/prelude/
metadata.md` and the `supplements/` source it is generated from, which is the
right pair to touch. The snippets are plain ` ```eu ` fences carrying `RESULT`
expressions, so the doctest CI **executes** them — they are gates, not
illustrations. All 111 doc examples pass.

The note that the built-in monadic tags (`:io`, `:for`, `:let`, `:random`,
`:state`) never reach the runtime this way, because the desugarer consumes them
and rewrites the block into a bind chain, is a genuinely useful clarification —
it is the first question a reader has after seeing `{ :only k: 1 } meta` return
a symbol.

### Gates

Run in the review worktree, with `lib/prelude.blob` regenerated after the
branch switch and `prelude_blob_ok` confirmed **from the build-script output**
rather than inferred from the file existing:

```
cargo:rustc-cfg=prelude_blob_ok
```

- `cargo fmt --all` — clean
- `cargo test` — **0 failed across all 17 test binaries**: lib 1387, harness
  517, doc examples 111, diagnostics invariants 11, wire-format enforcement 8.

Merging once this PR's own CI board and master's are both green.

PR #1082, opened 2026-07-27
fix(build): hash the intrinsic catalogue into the prelude-blob freshness check (eu-3skeg)

gmorpheme:
Closes eu-3skeg (P1, 0.14 release gate). **Touches the blob wire surface — recorded non-author review required before merge.**

## The defect

The blob bakes every prelude global in as `Ref::G(INTRINSIC_COUNT + prelude slot)`, so adding or removing an intrinsic shifts the entire global slot map. But the freshness hash covered only `lib/prelude.eu`, which an intrinsic-adding change need not touch. A blob generated before the new intrinsic was therefore accepted as fresh with an off-by-N slot map, and the binary returned **wrong answers with no warning**.

Two compounding problems, both closed here:

1. The mitigation was a doc comment asking future authors to hand-bump `BYTECODE_WIRE_FORMAT_VERSION`.
2. The constant was **duplicated** in `build.rs:38` and `xtask/src/main.rs:57`, kept in step only by a comment saying "MUST match" — and PRs #1078 and #1079 independently collided on bumping it within a single day.

## What I chose, and why

**Option (b): hash `src/eval/intrinsics.rs` into the blob freshness hash.** Chosen over (a) and (c):

- **(b) over (a) "derive the version from `INTRINSIC_COUNT`"** — a count is not the whole invariant. Reordering or renaming intrinsics, or changing an intrinsic's arity, also corrupts a baked blob while leaving the count identical. Hashing the catalogue covers count, order, names and arities in one term. It also keeps the version constant meaning one clean thing (encoder byte layout) instead of two entangled ones.
- **(b) over (c) "a CI check that fails when `INTRINSIC_COUNT` changes without a bump"** — (c) leaves the human step in place and only nags about it, and it cannot help the local developer who is the one actually hitting the wrong answers (`lib/prelude.blob` is git-ignored, so this bug lives in local builds; CI regenerates the blob unconditionally anyway).

Both files are hashed at the **byte** level rather than parsed for intrinsic names. Extracting names would need a Rust parser inside `build.rs` that could itself drift out of step with the file it parses — reintroducing the failure class in a subtler form. Byte hashing over-invalidates slightly (a comment-only edit to `intrinsics.rs` forces a blob regeneration), but that is the safe direction, the file changes rarely (14 of the 1168 commits between 2026-05-01 and 2026-07-26), and a stale blob degrades to a build warning plus the source-prelude fallback — correct, merely slower.

`build.rs` now also declares `rerun-if-changed` for the catalogue. Without it the whole mechanism is inert: emitting *any* `rerun-if-changed` directive opts cargo out of whole-package tracking, and `intrinsics.rs` is read rather than compiled in, so cargo would never re-run the check. This is the same caching trap the dispatch flagged, and it is part of why the original bug bit even while the crate itself recompiled.

## De-duplication: yes

`BYTECODE_WIRE_FORMAT_VERSION`, the hash recipe, the input paths and the staleness verdict now live once, in **`src/eval/stg/wire_format.rs`**:

- `build.rs` reaches it via `mod wire_format { include!("src/eval/stg/wire_format.rs"); }` — a build script cannot depend on its own crate, which is exactly what tempted the original duplication;
- `xtask` imports `eucalypt::eval::stg::wire_format::*` (its `sha2` dependency is dropped — it no longer hashes anything itself);
- the enforcement tests assert on the same items.

The file carries a comment explaining the constraints that keep it `include!`-able (no `//!` inner doc comments, no `crate::` references, no dependency beyond `sha2`, which is both a dependency and a build-dependency).

## Does this change need a version bump? No.

Deliberately **not** bumped, and I did not touch the value (master is 5; #1078 and #1079 both propose 6 — the owner decides). Two reasons:

1. `BYTECODE_WIRE_FORMAT_VERSION` tracks the serialised code-stream layout, which this change does not touch.
2. A bump would be redundant: the hash *recipe* itself changed (domain-separated, length-prefixed absorption of three labelled inputs), so every pre-existing blob is invalidated automatically.

**Coordination note for whoever merges:** this PR deletes the two constant declarations that #1078 and #1079 both edit, so there will be a textual conflict with whichever of those merges first. Resolution is mechanical — set the single constant in `src/eval/stg/wire_format.rs` to whatever value they land on. The mechanism works for any value.

## How the enforcement is tested

`tests/wire_format_enforcement_test.rs` (8 tests) and `wire_format.rs`'s own unit tests (7):

| Test | Gates |
|---|---|
| `the_wire_format_version_is_defined_exactly_once` | scans every `.rs` in the workspace; exactly one definer, and it is `wire_format.rs` |
| `build_rs_shares_the_one_definition` / `xtask_shares_the_one_definition` | both use the shared recipe and neither constructs its own `Sha256` |
| `the_build_script_reruns_on_the_inputs_it_hashes` | the catalogue and the `include!`d module are in the rerun set |
| `every_tracked_input_path_resolves` | no rerun path is a no-op typo (`BLOB_PATH` exempt — git-ignored artefact) |
| `the_hashed_file_is_the_intrinsic_catalogue` | `intrinsics.rs` still declares `INTRINSICS`/`catalogue()`, so hashing that path still protects something |
| `the_embedded_blob_matches_the_inputs_on_disk` (`cfg(prelude_blob_ok)`) | recomputes the hash from disk at *test* time — fires even if a cached `cfg(prelude_blob_ok)` were carried over a change cargo failed to notice |
| `the_freshness_hash_covers_the_real_inputs`, `intrinsic_table_changes_change_the_hash`, `a_blob_predating_a_new_intrinsic_is_stale` | the core invariant, at both hash and verdict level |

## Fault-injection verification — done

**1. The real defect, reproduced and closed (end to end).** With the intrinsic-table term removed from `blob_source_hash`, blob regenerated under the faulted recipe, a `SCRATCH_FAULT_INJECTION` intrinsic appended to the catalogue, and `cargo build --release`:

```
$ ./target/release/eu -e '[1,2,3] map(inc)'
- 0
- 1
- 2                 # no warning: exactly the reported symptom
```

Restoring the term, same added intrinsic, same rebuild:

```
warning: eucalypt@0.14.0: prelude blob is stale (prelude source, intrinsic table or
bytecode wire format has changed) — run `cargo xtask prelude-compile` to regenerate.

$ ./target/release/eu -e '[1,2,3] map(inc)'
- 2
- 3
- 4                 # source-prelude fallback: correct
```

Restored, blob regenerated, `touch build.rs`, rebuilt: `[2, 3, 4]` in blob mode, and `the_embedded_blob_matches_the_inputs_on_disk` compiles in (8 tests, not 7) confirming `cfg(prelude_blob_ok)` really is on — the mtime trap the dispatch warned about, checked explicitly rather than assumed.

**2. Test-level fault injection**, each applied and reverted individually:

| Injected fault | Failing test(s) |
|---|---|
| drop the `intrinsic-table` term from `blob_source_hash` | `the_freshness_hash_covers_the_real_inputs`, `intrinsic_table_changes_change_the_hash`, `a_blob_predating_a_new_intrinsic_is_stale` |
| re-add a second `const BYTECODE_WIRE_FORMAT_VERSION` to `build.rs` | `the_wire_format_version_is_defined_exactly_once` (reports `["build.rs", "src/eval/stg/wire_format.rs"]`) |
| remove `INTRINSIC_TABLE_PATH` from `BUILD_RERUN_PATHS` | `the_build_script_reruns_on_the_inputs_it_hashes` |

## Validation

- `cargo fmt --all --check` clean; `cargo clippy --workspace --all-targets -- -D warnings` clean.
- **Blob-less** (the CI `Test Suite` configuration): `cargo test --no-fail-fast` fully green.
- **Blob mode**: `cargo test --no-fail-fast` green except three failures — `secondary_labels_do_not_excerpt_library_internals`, `debug_trace_restores_the_uncurated_trace` (both eu-7x0r, named in the workflow comment as known-red under a blob) and `test_193_1tkk_7_12_curated_trace`. I stashed this branch, regenerated the blob at master state and rebuilt: **the identical three fail there too.** Pre-existing, not introduced here.
- **Both engines**: `EU_HEAPSYN=1 cargo test --no-fail-fast` — same three pre-existing failures, nothing new. `EU_HEAPSYN=1 eu test --allow-io tests/harness` — 0 failures.
- **GC**: `EU_GC_VERIFY=2 EU_GC_STRESS=1 eu test --allow-io tests/harness` — 0 failures.
- **Blob determinism** (the `blob-determinism` CI job's invariant): two consecutive `prelude-compile` runs byte-identical, `7bf2c4c5…`.

## Docs

`docs/architecture.md`'s staleness section rewritten (it still said "currently 3"; it now names the three hashed inputs and quotes no version number, so it cannot drift again), plus `blob.rs`'s anatomy table and `source_hash` field doc, and a CHANGELOG entry.


gmorpheme:
## Recorded review (wire-format surface, CLAUDE.md non-author review)

Reviewed by Wicket, not the author. Verified in an isolated worktree off `5dd7c865`. Recommend **merge**, with one follow-up worth a bead (not a blocker).

### What I verified myself, by running it

**The core claim — the staleness verdict genuinely covers the intrinsic catalogue.** Independently fault-injected rather than taken on trust:

1. Fresh blob, clean build → build-script output `cargo:rustc-cfg=prelude_blob_ok`.
2. Appended a **comment-only** line to `src/eval/intrinsics.rs`, and rebuilt **without touching `build.rs`** →
   `cargo:warning=prelude blob is stale (prelude source, intrinsic table or bytecode wire format has changed)` and `cargo:rustc-cfg=prelude_blob_stale`.
3. Reverted → back to `prelude_blob_ok`.

Step 2 is the one that matters and it tests two things at once. It confirms the catalogue is really in the hash, and — because I deliberately did not touch `build.rs` — it confirms the `rerun-if-changed` declaration for `INTRINSIC_TABLE_PATH` actually fires. That second half is the part the PR body correctly identifies as the mechanism being otherwise inert, and it is the easiest thing to get silently wrong.

**Failure direction is safe everywhere I could find.** `read_blob_source_hash` reads the first 32 bytes on the assumption that `source_hash` is `PreludeBlob`'s first field and that postcard emits `[u8; 32]` unprefixed. I checked the struct: `source_hash` is indeed first (`blob.rs:108`). More importantly, if anyone ever reorders it, `build.rs` compares against the wrong bytes, mismatches, and returns `Stale` → source-prelude fallback. Every path in `classify_blob` that is not an exact hash match lands on a non-`Fresh` verdict, and `BlobFreshness::warning()` is total over the enum. A false `Fresh` would need a SHA-256 collision. The design fails toward "slow but correct", which is the right direction for a defect class whose symptom is silent wrong answers.

**Design.** Option (b) is the right call and the reasoning given holds up. Hashing the file rather than parsing it avoids putting a Rust parser in `build.rs` that could drift out of step with the file it parses — that would have reintroduced eu-3skeg in a subtler form. The over-invalidation cost is real but trivial and in the safe direction. Domain-separated, length-prefixed absorption is correct and `inputs_are_unambiguously_separated` gates it. Collapsing the duplicated constant into one `include!`-able home removes the actual root cause rather than papering over it, and the constraints that keep the file `include!`-able are documented in the file itself where the next author will see them. No architectural smell: the mechanism is general over its inputs, not special-cased to intrinsics.

**Not bumping the version is correct.** The recipe changed, so every pre-existing blob is invalidated regardless; bumping as well would be noise. Agreed.

### Finding: one residual gap of the same class (follow-up, not a blocker)

`build-meta.yaml` is baked into the blob — `xtask/src/main.rs:98-147` loads it from disk and pushes its desugared core into `desugared_unit_cores` under the `build-meta` tag — but it is **not** an input to `blob_source_hash`. It is in `BUILD_RERUN_PATHS`, so the build script re-runs, but re-running only recomputes a hash that does not cover it.

Verified: appended a key to `build-meta.yaml`, rebuilt, and the verdict stayed `prelude_blob_ok`. The blob keeps its stale baked `build-meta` core and is embedded as fresh. That is structurally the same shape as eu-3skeg — a blob input outside the freshness hash.

I then checked the severity before calling it, and it is **much narrower than eu-3skeg**. I changed `version:` to `9.9.9-wicket`, rebuilt without regenerating the blob, and `eu -e "__build.version"` reported `9.9.9-wicket`, not the stale value. So `__build` is served at runtime from the `include_bytes!` resource in `src/driver/resources.rs`, which cargo does track — the baked core is not what evaluation reads. The exposure is confined to whatever consumes `desugared_unit_cores`, i.e. the blob-mode typecheck path. No wrong evaluation answers.

Adding a third `absorb` term for `build-meta.yaml` would close it for the cost of one line, and the `io`/`args` pseudoblocks are worth the same look. I would not hold the PR for it — a bead is the right home.

### What I did NOT verify

- I did not re-run the author's full end-to-end fault injection (faulted recipe + appended `SCRATCH_FAULT_INJECTION` intrinsic + release build). My comment-only injection exercises the same hash term and the same verdict path, and a real intrinsic addition is a strict superset of it, but I am not independently confirming their exact transcript.
- I did not audit all 15 enforcement tests line by line; I read the table and spot-checked the two that carry the invariant (`intrinsic_table_changes_change_the_hash`, `a_blob_predating_a_new_intrinsic_is_stale`). Both are genuine — each has a stated fault-injection that flips it, and I confirmed by inspection that the assertion is computed from the code under test rather than restating a constant.
- I relied on the PR's CI for `cargo test` / clippy / fmt across platforms rather than re-running the matrix locally.
- I did not evaluate the merge conflict with #1078/#1079 beyond confirming the PR body describes it accurately; per the queue plan the single constant lands at **7** (two format changes on master's 5).

PR #1085, opened 2026-07-27
feat(diagnostics): golden snapshots + --bless + prelude-source suppression, with 0.13.2 baseline (eu-1tkk.7.4)

gmorpheme:
Implements **eu-1tkk.7.4** — golden diagnostic snapshots, `--bless`, and prelude-source suppression (design spec §5.3, epic eu-1tkk.7).

Release-gating for 0.14: the claim is "diagnostics greatly improved", and this is the evidence for it. Not a suite that says diagnostics are *fine* — a before/after that can be audited.

## The sequencing point

Four diagnostics fixes were approved but unmerged when this was written (#1077, #1079, #1080, eu-og3u6 in progress). Once they land, "before" is unreconstructable in practice. Two captures are therefore **frozen as data** in this PR:

| baseline | ref | what it is |
|---|---|---|
| `0.13.2.snapshots.txt` | tag **0.13.2** (released 2026-07-22) | the last thing users installed |
| `master-a1af1f0b.snapshots.txt` | master **a1af1f0b** | 0.14-dev *before* the pending fixes |

Same 195-fixture corpus, same invocation; only the binary differs. Comparison and interpretation: `docs/superpowers/reports/2026-07-27-diagnostics-snapshot-baselines.md`.

## Headline result (0.13.2 → master a1af1f0b, over the 93 changed fixtures)

| property | 0.13.2 | master a1af1f0b |
|---|---|---|
| prelude source lines excerpted into diagnostics | **198** | **0** |
| fixtures whose primary label points into the prelude | **21** | **0** |
| trace frames pointing into the prelude | **99** | **9** |
| trace frames pointing into a **user** file | **9** | **83** |
| total trace frames | 108 | 92 |
| fixtures carrying a stable error code | 0 | 17 |
| fixtures with **no** primary label at all | 0 | **9** |

Traces used to be 92% prelude frames; they are now 90% user frames, and shorter.

`tests/diagnostics/corpus/nth_out_of_range.eu` — `result: xs nth(10)` on a three-element list.

**0.13.2**

```text
error: tail of empty list
     ┌─ [prelude]
     <4 prelude source line(s) suppressed>
     = guard against empty lists with 'nil?', e.g. 'if(xs nil?, default, xs head)'
     = 'head' and 'tail' are only defined on non-empty lists; use 'nth(0, xs)' or pattern matching if the list may be empty
     = stack trace:
       - nth at [prelude]
       - drop at [prelude]
       - drop at [prelude]
       - drop at [prelude]
         ... 10 frames elided (6x repetition)
       - drop at [prelude]
```

**master a1af1f0b**

```text
  ┌─ tests/diagnostics/corpus/nth_out_of_range.eu:2:12
  │
2 │ result: xs nth(10)
  │            ^^^
  │
  = stack trace:
    - in 'nth' at [prelude]
```

0.13.2 did not merely mislocate: it reported the **wrong error** ("tail of empty list", an internal consequence of `nth`'s implementation). `errors/109_unknown_arg` is the extreme case — a 15-frame all-prelude trace with a 17-line prelude source excerpt becomes a one-frame user-anchored trace pointing at the user's own call.

## How the mechanism works

- **`tests/diagnostics/snapshot_engine.rs`** — std-only capture / normalise / compare engine, `#[path]`-included by *both* the test target and xtask, so a snapshot written by one is byte-identical to one checked by the other.
- **`tests/diagnostics_snapshots.rs`** — the gate. 195 fixtures from `tests/harness/errors/` (179), `tests/diagnostics/corpus/` (9) and the new many-declaration corpus (7). Fixtures are auto-discovered; a fixture with **no** golden fails rather than being skipped. ~35 s (fixtures run 8-way parallel; the sweep runs once per test binary and is shared between tests).
- **`xtask/src/diag_snapshot.rs`** — `--bless`, `--capture` (`--from-ref <ref>` / `--binary <path>`), `--compare`.
- `cargo xtask` alias added to `.cargo/config.toml` so the form used throughout CLAUDE.md actually works (there was no alias; only `cargo run --package xtask --`).

### Runnable against any checkout

`--from-ref` builds `eu` in a throwaway worktree and captures with it. The **corpus** always comes from the current checkout; only the **binary** comes from the ref. An older `eu` embeds its own prelude at its own compile time, so this measures that release's diagnostics rather than a hybrid. That is how the 0.13.2 baseline was taken retroactively.

The facts block is derived from the **human** rendering, not `--error-format json` — which is precisely what makes that possible, since 0.13.2 predates the structured diagnostic model (eu-1tkk.7.1).

### Prelude-source suppression

The rustc `ignore-directory-in-diagnostics-source-blocks` analogue. Prelude line/column numbers and prelude source excerpts are removed — they churn on every edit to `lib/prelude.eu` without saying anything about diagnostic quality. The combinator *name* (`in 'nth' at [prelude]`) is kept, and the **count** of suppressed excerpt lines is retained as a `prelude-excerpt-lines` fact, so a regression that starts excerpting library internals is still visible while the churny text is gone. Exercised by real data: the 0.13.2 baseline suppressed 198 such lines.

Also normalised: ANSI colour, the checkout's absolute path, thread ids in panic banners, and `:LINE:COL` in rustc panic positions (which *file* panicked is kept).

### Both prelude modes — divergence is recorded, not averaged away

Every fixture is run under the blob prelude (what a released binary uses) *and* `--source-prelude`. Agreeing modes get one `=== both prelude modes ===` section; disagreeing modes get `prelude-modes: divergent` and two labelled sections, plus a row in **`tests/diagnostics/DIVERGENCE.md`** — generated by `--bless`, asserted by a test, so a new eu-7x0r-class divergence is a one-line change in one reviewable file rather than something buried in one of 195 `.snap` diffs.

**13 of 195 fixtures currently diverge.** `errors/143_bitwise_float` is the sharpest: the **blob** path — the one released binaries use — renders *no source location* where `--source-prelude` renders one.

The blob half is `#[cfg(prelude_blob_ok)]` and runs in the existing `diagnostics-blob-mode` CI job. Because that gate can vanish silently (a `lib/prelude.blob` with a stale mtime does not re-trigger `build.rs`, `prelude_blob_ok` goes unset, the tests compile out, and the job exits 0 having run nothing), the job now **asserts the test names are present in `--list`** before running them. On the authoring side, `--bless` regenerates the blob itself immediately before building, making that state unreachable.

## Many-declaration corpus (`tests/diagnostics/snapshot-corpus/`)

Seven fixtures whose only distinguishing feature is size, written deliberately against the blind spot that hid **eu-og3u6** (wrong declaration blamed once other bindings render first — correct in a 2-line file, wrong in a 900-line one) and **eu-r4647** (a safety argument resting on "a Smid far outside any range this test's tiny SourceMap will ever register"). Sizes 600–2000 declarations, covering failure-last / failure-first / failure-in-the-middle, a nested user helper, a lookup miss, and both the evaluator and the **type-checker** path.

They earned their keep on the first capture. `many_decls_late_nth` in blob mode:

```text
    ┌─ .../many_decls_late_nth.eu:477:16
    │
477 │ pad0470: 470 + 0
    │                ^
    │
    = stack trace:
      - pad0000 at many_decls_late_nth.eu:7:12
      - pad0455 at many_decls_late_nth.eu:462:10
      - pad0470 at many_decls_late_nth.eu:477:16
```

Three unrelated `padNNNN` declarations blamed for a failure in `result` — the eu-7x0r aliasing shape, live on this path. And `many_decls_early_nth` shows the "correct in a small file" property does **not** survive into blob mode at all. Both are recorded, not asserted: the snapshot diff when #1079 and eu-og3u6 land is the proof they were fixed.

Kept out of the invariant-gate corpus on purpose, so this PR cannot collide with the eu-og3u6 fix in flight.

## Found by the mechanism, filed not fixed

- **eu-1tkk.7.20** (P1) — `171_yaml_large_uint` aborts with a raw Rust panic (`RUST_BACKTRACE` note, `main thread panicked: Any { .. }`), no location, no code, no trace. Green all along because its `.expect` sidecar asserts only `exit: 101` and a substring the panic happens to contain. The clearest single demonstration of why full-output snapshots beat regex sidecars.
- **eu-1tkk.7.21** (P2) — nine fixtures traded a *meaningless* prelude primary for **no** primary. Satisfies invariant (i) by deleting the label. Not a win; reported as such.
- **eu-1tkk.7.22** (P3) — the gate is skipped on Windows (path rendering and CRLF unverified against the goldens); filed so the skip is not silently permanent.

## What these snapshots will and will not prove

**Will**: that a diagnostic changed, exactly how, and how its objective properties moved — all mechanically derived and reproducible from checked-in data.

**Will not**: that a diagnostic is *good*, or that a given change is an improvement. Seventeen fixtures gained an error code because the structured model did not exist at 0.13.2, not because those messages got clearer. Nine deltas are ambiguous. The corpus is our fixtures, not real user programs. Per the design spec's honesty guardrail there is no validated automatic quality metric, and LLM-as-judge specifically failed to predict human outcomes (Santos & Becker, UKICER 2024) — so these are evidence for review, not a score.

Every gate was broken and confirmed to fail, then restored and confirmed to pass:

1. **Diagnostic text changed** — added `FAULT INJECTION` to the divide-by-zero help in `src/eval/error.rs`. `source_prelude_snapshots_match` FAILED, naming `errors/006_div_by_zero`, `errors/041_modulo_by_zero`, `errors/097_div_by_zero_source_loc` with the exact `+`/`-` diff. Restored → passes.
2. **Golden missing** — deleted `errors/006_div_by_zero.snap`. FAILED with `1 fixture(s) have no golden snapshot`. Restored → passes.
3. **Divergence inventory stale** — removed the `errors/143_bitwise_float` row from `DIVERGENCE.md`. `divergence_inventory_is_current` FAILED. Restored → passes.
4. **The two halves gate independently** — corrupted *only* the blob section of `errors/143_bitwise_float.snap`. `blob_prelude_snapshots_match` FAILED, `source_prelude_snapshots_match` passed. Restored → passes.

The normalisation rules are additionally pinned by 12 self-tests, including one asserting **user** source excerpts survive verbatim (a normaliser that ate those would make the goldens worthless) and one for the prelude-excerpt suppression path that today's master output does not reach.

## Gates

- `cargo fmt --all --check` — clean
- `cargo test` **without** a prelude blob (what the main Test Suite job does) — all green; 15 snapshot tests, blob half correctly compiled out
- `cargo test --no-fail-fast` **with** a prelude blob — `diagnostics_snapshots` 17/17 green. Three failures elsewhere are pre-existing blob-mode reds (`diagnostics_invariants` x2 and `harness_test::test_193_1tkk_7_12_curated_trace`, the eu-7x0r class documented in the workflow comment). `git diff origin/master -- src lib build.rs Cargo.toml tests/harness tests/harness_test.rs tests/diagnostics_invariants.rs` is **empty**, so no input to any of those tests was touched here.
- `python3 scripts/test-doc-examples.py` — 245 passed, 0 failed (new docs use `text`/`bash` fences only)
- Toolchain updated 1.97.0 → 1.97.1 mid-verification; snapshots byte-identical across both
- Debug and release builds verified to produce byte-identical diagnostics for all 195 fixtures, so the release-built baselines are comparable with the debug-built goldens `cargo test` produces

## Review guidance

Two hundred `.snap` files is a lot of green diff. The load-bearing files are:

1. `docs/superpowers/reports/2026-07-27-diagnostics-snapshot-baselines.md` — the claim and its caveats
2. `tests/diagnostics/DIVERGENCE.md` — 13 live blob/source bugs
3. `tests/diagnostics/SNAPSHOTS.md` — what is normalised away, and why each rule is safe
4. `tests/diagnostics/snapshot-corpus/README.md` — why the many-declaration fixtures exist
5. `tests/diagnostics/snapshot_engine.rs` — in particular `normalise`, which is where signal could be silently destroyed

**Note for whoever merges after #1077/#1079/#1080/eu-og3u6:** this PR will conflict with them by design — those fixes change diagnostics, so the goldens need re-blessing. That diff is not a nuisance; it is the deliverable. Rebase, run `cargo xtask diag-snapshot --bless`, and the resulting diff is the proof those PRs improved things.


gmorpheme:
## CI: 18/18 green on `a8b2fdb4`

Full rollup from the first run (before the follow-up hardening commit `e966dc95`):

```
AddressSanitizer                            pass   6m28s
Byte-dispatch harness (soak baseline)       pass   2m42s
Bytecode + blob harness (macos-latest)      pass   5m00s
Bytecode + blob harness (ubuntu-24.04-arm)  pass   3m14s
Bytecode + blob harness (ubuntu-latest)     pass   2m51s
Diagnostics blob-mode integration           pass   3m12s
GC-verified harness (x86-64)                pass   3m10s
HeapSyn harness (baseline)                  pass   2m55s
Lint                                        pass   1m31s
Prelude blob determinism                    pass   1m13s
Prelude docs freshness                      pass   1m47s
Security Audit                              pass   3m13s
Source-prelude harness (x86-64)             pass   2m48s
Source-prelude tick-parity tripwire         pass   1m54s
Sequential harness (aarch64 release)        pass   2m39s
Test Suite (ubuntu-latest)                  pass
Test Suite (windows-latest)                 pass
WASM Compilation Check                      pass   4m56s
```

Two results worth calling out:

- **Test Suite (ubuntu-latest) passed.** The 195 goldens were captured on macOS; this is the proof they are byte-identical on Linux. That was the main portability risk in this PR.
- **Diagnostics blob-mode integration passed**, which is the only job that runs the blob half of the snapshots and the new `--list` presence assertion.

Windows passed too — the gate deliberately skips there (eu-1tkk.7.22), so that result confirms the skip is clean rather than that the goldens are portable to Windows.

## Follow-up commit `e966dc95`

Hardening on the baseline capture path, prompted by a live stash incident in the shared repo (`refs/stash` is shared across all worktrees, so stashing to move between refs can restore a stranger's work into the tree a baseline is captured from):

- documented why `build_ref` uses a separate worktree rather than a stash, so it does not get "simplified" into one later;
- `git worktree prune` before `git worktree add`, since the worktree lives under `target/` and `cargo clean` removes the directory while leaving the registration behind;
- README: how to verify a committed baseline was not captured from a contaminated tree.

**Both committed baselines were re-captured from scratch and compared with `cmp` — byte-identical**, so the frozen evidence in this PR is verified reproducible, not merely asserted.

## Fifth fault injection

The CI mtime-trap tripwire was verified the same way as the other four: removed `lib/prelude.blob`, `touch build.rs`, and confirmed the `--list` grep step exits 1 (job fails loudly) instead of the job running zero tests and reporting green. With the blob restored it finds both test names and exits 0.

CI is re-running on `e966dc95`; I will post that rollup when it completes.

gmorpheme:
## Final CI: 18/18 green on `e966dc95`

```
AddressSanitizer                            pass   16m53s
Byte-dispatch harness (soak baseline)       pass    2m48s
Bytecode + blob harness (macos-latest)      pass    3m43s
Bytecode + blob harness (ubuntu-24.04-arm)  pass    3m04s
Bytecode + blob harness (ubuntu-latest)     pass    3m25s
Diagnostics blob-mode integration           pass    3m09s
GC-verified harness (x86-64)                pass    2m49s
HeapSyn harness (baseline)                  pass    2m52s
Lint                                        pass    1m23s
Prelude blob determinism                    pass    1m22s
Prelude docs freshness                      pass    1m54s
Security Audit                              pass    3m15s
Sequential harness (aarch64 release)        pass    2m37s
Source-prelude harness (x86-64)             pass    2m23s
Source-prelude tick-parity tripwire         pass    1m51s
Test Suite (ubuntu-latest)                  pass   13m45s
Test Suite (windows-latest)                 pass   18m12s
WASM Compilation Check                      pass    4m47s

prepare-release / Release Linux x86_64 / Release Linux aarch64 /
Release MacOS / Release Windows             skipping  (expected on a PR)
```

Both runs green — `a8b2fdb4` (18/18) and `e966dc95` (18/18).

Ready for owner review. Not merging, and the bead stays open.

gmorpheme:
## Gatekeeper review — approved, re-blessed, merging on green

`433d863a` — the head of the 0.14 queue, with all fourteen preceding PRs on it
— and re-blessed there with `cargo run --package xtask -- diag-snapshot
--bless`. `lib/prelude.blob` regenerated after the branch switch and

**Blessed 210 snapshots, 0 orphans pruned.**

---

## The blessed diff is the release evidence for the whole 0.14 diagnostics programme

The golden snapshots were taken before thirteen diagnostics PRs landed, so what
this diff records is not "one PR changed some output" — it is **the aggregate
effect of the 0.14 diagnostics work, on 210 real fixtures**. I read every
differing fixture rather than accepting the count.

### The headline: blob and source stopped disagreeing

`tests/diagnostics/DIVERGENCE.md`:

```
- **13 of 195 fixtures diverge.**
+ **4 of 210 fixtures diverge.**
```

**Nine divergences eliminated, fifteen fixtures added, and not one new
divergence introduced.** Every fixture in that file rendered a *different
diagnostic* depending on whether `eu` used the pre-compiled prelude — what a
released binary does — or `--source-prelude`. Same program, same mistake, two
different answers.

Gone: all four `many-decls` fixtures, `errors/139_list_index_oob`,
`errors/181_direct_app_call_site`, `errors/193_1tkk_7_12_curated_trace`,
`provocations/nth_out_of_range`, `provocations/swap_args`.

Four remain, and they are worth naming because they are the honest residue:

| fixture | what still differs |
|---|---|
| `errors/011_assert_pred` | identical facts — diverges in text only |
| `errors/035_nested_fn_trace` | blob blames 2:12, source 1:11 |
| `errors/143_bitwise_float` | **blob has no primary label at all**; source has one |
| `errors/error_176` | primaries now agree; blob has 1 trace frame, source 0 |

`143_bitwise_float` is the one I would put in front of the owner: in blob mode —
the shipped configuration — that diagnostic carries **no source location**.
`error_176` is a partial fix rather than a regression: it was `19:32` vs
`16:22` and the two sides have now converged on `19:32`, leaving only a trace
difference.

### Three distinct improvements, visible in the snapshots

**1. Blob-mode traces gained their library frames (eu-7x0r, #1079).**

```
    +     - in 'nth' at [prelude]
```

added to `139_list_index_oob`, `193_1tkk_7_12_curated_trace`,
`nth_out_of_range`, `swap_args` and others. Before, a blob-mode trace named *no
library combinator at all*.

**2. Both modes stopped blaming a declaration the user never called
(eu-og3u6, #1083).** The clearest specimen:

```
- 7 │ pad0000: 0 + 0
-   │            ^
-     - pad0000 at many_decls_late_nth.eu:7:12
+ 907 │ result: xs nth(10)
+     │            ^^^
+       - result at many_decls_late_nth.eu:907:12
```

`pad0000` is padding. The user never wrote a call there. The same correction
appears in `huge_decls_type_mismatch`, `huge_decls_type_warning`,
`many_decls_early/middle/late_nth`, `many_decls_nested_user_fn`,
`many_decls_lookup_miss` — and, less obviously, in
`083_array_arith_mismatch`, where the trace moves from `a at :2:4` to `c at
:4:4`. I checked that fixture's source: the failing call `arr.add(a, b)` **is**
the declaration `c`, and `a` was merely the first declaration in the file.

**3. A panic became a diagnostic (eu-1tkk.7.20, #1089).**

```
  exit: golden 101 → actual 1
  facts: golden … rust_panic: true
         actual … code: Some("EU-RENDER-UNREPRESENTABLE"), primary: User("…:4:7"), rust_panic: false
```

The golden file had recorded a **Rust backtrace** as the expected output.

### The one category I had to think about

Around eight source-prelude fixtures *lose* their stack trace entirely
(`trace_frames: 1 → 0`): `021_bad_num_parse`, `041_modulo_by_zero`,
`080_array_oob`, `089_numeric_domain`, `090_bad_format_string`,
`105_base64_decode_invalid`, `141_calzdt_year_overflow`,
`150_bad_timezone_source_loc`.

Losing trace frames reads as a regression, so I checked each. In every case the
removed frame is at **exactly the `file:line:col` already carried by the primary
label** — e.g. primary `021_bad_num_parse.eu:1:6` and trace `- bad at
021_bad_num_parse.eu:1:6`. A one-frame trace that repeats the primary label is
noise, and dropping it is the curation eu-og3u6 set out to do. No fixture loses
a frame that named anything the primary did not already say.

### Verification I ran rather than inferred

- No fixture acquires `rust_panic: true`.
- No fixture's primary becomes `None`.
- No `+` line anywhere reintroduces a `pad*` frame.
- 210 snapshots blessed, **0 orphans pruned** — so no fixture silently
  disappeared from the corpus.

---

## The mechanism, reviewed on its own terms

`Facts` — errors, warnings, code, primary, help/note line counts, secondary
labels, trace frame counts split user/prelude, prelude excerpt lines,
`rust_panic` — is the right abstraction. A pure text snapshot would make every
diff a wall of prose; extracting facts means the summary line
(`trace_frames: 1 → 0`) tells you *what kind* of change it is before you read a
character of the rendering. That is what made the eight-fixture category above
recognisable as one category rather than eight separate judgements.

Normalisation is properly tested in its own right — twelve unit tests covering
ANSI stripping, thread ids and panic line numbers, checkout paths, prelude path
rewriting, prelude line-number suppression **while keeping the combinator
name**, and idempotence. `is_idempotent` is the one that matters: a
normalisation that is not idempotent produces spurious diffs forever.

`DIVERGENCE.md` being **generated, not hand-maintained**, with
`divergence_inventory_is_current` asserting it, is the right call. Its header
says both directions are of interest — a new row is a regression of the eu-7x0r
class, a removed row is a fix worth recording — and this bless is nine removed
rows, which is precisely the use it was designed for.

The message on failure is well judged: *"The diff — not the fact that a
snapshot changed — is the thing to review."* Golden-snapshot suites fail when
blessing becomes reflexive, and that line is aimed at exactly that.

## Two changes applied on top of the author's work

**1. The blob-liveness check, kept rather than replaced.** Master's blob-mode
job now runs `scripts/diagnostics-gate.sh` (#1094), which has its own
zero-tests guard. I routed `diagnostics_snapshots` through it **and kept this
branch's bespoke `--list` assertion**, because the two are not equivalent:
the script asks whether the target ran *anything*, and this file has ungated
tests too — so it would stay green with only those running while both
blob-gated tests silently vanished. Naming `blob_prelude_snapshots_match` and
`divergence_inventory_is_current` is the stronger assertion and it is the one
this file needs.

**2. Registered with `gate_liveness` (#1095).** On merge,
`every_conditionally_compiled_test_file_is_registered` failed:

```
these test files are conditionally compiled but are not in REGISTRY:
  tests/diagnostics_snapshots.rs
```

I wrote the entry, recording both what the gated tests assert and *why the
`--list` check exists alongside the target-level guard*, so the next person does
not delete one as redundant.

**That is the third omission #1095's registry has caught** — the first two were
`blob_deprecation_test.rs`, `wire_format_enforcement_test.rs` and
`harness_test.rs` on its own merge. A gate that has caught three real
omissions within hours of existing is carrying its weight.

## Gates

- `cargo fmt --all` — clean
- `scripts/diagnostics-gate.sh diagnostics_snapshots` — exit 0
- `cargo test --test gate_liveness` — 9/9

PR #1086, opened 2026-07-27
fix(core): share, don't copy, duplicated inline arguments (eu-gua64)

gmorpheme:
Fixes **eu-gua64** (P1).

## The defect

`beta_reduce` (`src/core/inline/reduce.rs`) substituted a call's arguments into the callee body with `substs_depth(&body, &mappings, 0)` — no occurrence count, no work-duplication guard. When the callee mentions a parameter more than once, the argument **expression** is copied per occurrence. Copies are separate expressions, so they become separate thunks and share nothing. When the duplicated argument is the caller's own recursive call, cost goes from `T(n) = T(n-1)` to `T(n) = 2·T(n-1)`.

Reproduced on a clean release build of master `a1af1f0b`, in pure user code, on **both** the blob and source-prelude paths:

```eu,notest
mymax(a, b): if(a > b, a, b)
mymax-of(l): if(l tail nil?, l head, mymax(l head, mymax-of(l tail)))
main: range(1, n) mymax-of
```

The blow-up needs the duplicated arm to be *taken*: ascending input returns the second copy and is 2^n, descending only forces the recursion once and is linear. That is why the symptom looked data-dependent.

## `eu dump inlined`, before and after

Before (`mymax-of(TAIL(l))` appears twice, verbatim — two thunks, no sharing):

```
mymax-of = λ(l).
IF(NILP(TAIL(l)), HEAD(l), IF(GT(HEAD(l), mymax-of(TAIL(l))), HEAD(l), mymax-of(TAIL(l))));
```

After (both duplicated arguments bound once; note `HEAD(l)` was being duplicated too):

```
mymax-of = λ(l).
IF(NILP(TAIL(l)), HEAD(l), let __shared_a = HEAD(l);
                               __shared_b = mymax-of(TAIL(l))
    in IF(GT(__shared_a, __shared_b), __shared_a, __shared_b));
```

## The fix — shape (a), let-binding

Both candidate shapes were considered. **(a) let-bind** was implemented; **(b) refuse to reduce** was not needed — (a) is sound here and keeps the inlining.

An argument is substituted directly only when doing so duplicates no work:

- it is **duplicable** — a variable (which already names a single shared thunk), a literal, or an intrinsic reference; or
- its binder occurs **at most once** in the callee body.

Otherwise it is bound once in a `Let` wrapped round the reduced body and every occurrence becomes a reference to that binding. `close_let_scope` performs the de Bruijn shift for both the body and the bound argument expressions as they move inside the new scope — the same adjustment `substs_depth` already performs, applied uniformly at every nesting level.

Occurrence counting needed a per-occurrence traversal, so `collect_free_vars` in `core/expr.rs` is generalised into a reusable `visit_free_vars` with a callback, and `free_vars` is reimplemented on top of it unchanged. That is the whole of the `expr.rs` diff — no behavioural change to `free_vars`.

### Why the sharing `Let` is *genuinely* non-recursive

Worth checking rather than trusting, because **eu-frf2o** (found during this task, see below) turns on exactly this distinction: a group emitted as non-recursive `Let` whose bindings *do* reference siblings resolves those refs in the enclosing environment and silently aliases whatever occupies those slots — which is how you manufacture a self-referential thunk.

The bindings introduced here cannot reference one another:

1. Each binding's value is a **call-site argument expression**, captured before any of the fresh names existed.
2. The binding names are **freshly minted** by `fresh_name`, which excludes every name in `avoid` — the union of `free_vars(body)` and `free_vars(arg)` over *all* arguments — and inserts each chosen name into `avoid` so successive names are distinct from each other too.

So a binding value cannot mention a sibling's name: the name did not exist when the expression was built, and is chosen disjoint from everything the expression does mention. Only the body refers to the bindings. Non-recursive semantics is correct here, not accidentally correct.

Arguments *can* of course reference the same enclosing bindings by de Bruijn index; those are shifted uniformly by `close_let_scope` and are not sibling references.

## Call-site Smid preserved

The existing re-tag is factored into `retag_call_site` and still applied to the reduced application, and the sharing `Let` carries the same `*call_smid`. This is load-bearing for eu-1tkk.7 / eu-og3u6 / eu-8a49h.

Demonstration as requested — `193_1tkk_7_12_curated_trace.eu`, whose primary label points at a user call site through an inlined prelude callee (`nth`). Byte-identical before and after, `┌─` line included:

```
  ┌─ 193_1tkk_7_12_curated_trace.eu:5:12
  │
5 │ result: xs nth(10)
  │            ^^^
  │
  = stack trace:
    - result at 193_1tkk_7_12_curated_trace.eu:5:12
```

## Performance — ticks, not wall clock

Wall-clock figures are deliberately **not** reported. The build machine was under heavy contention during this session (a 0.15 s job measured once at 33.5 s — a 220× excursion; the engine-ab spreads show the same, e.g. `io_loop` bc `[3.394..12.408]`). Ticks are deterministic and machine-independent, so the comparison below stands on its own. Read via HeapSyn `-S`, the same source `cargo xtask engine-ab` uses.

Reproduction, baseline vs fixed:

| N | ticks before | ticks after | allocs before | allocs after |
|--:|--:|--:|--:|--:|
| 16 | 1,460,505 | 3,257 | 246,050 | 417 |
| 18 | 5,835,325 | 3,669 | 983,356 | 459 |
| 20 | 23,333,729 | 4,081 | 3,932,502 | 501 |
| 22 | 93,326,469 | 4,493 | 15,729,008 | 543 |
| 24 | 373,296,553 | 4,905 | 62,914,954 | 585 |

Baseline is exactly ×4.000 per +2 in N — 2^N to four significant figures. Fixed is exactly linear, +412 ticks per +2, i.e. 206 ticks per list element. At N=24 that is a 76,100× reduction in ticks and 107,547× in allocations.

### Canonical suite — `bench/{015..022}`

`cargo xtask engine-ab --runs 3`, baseline vs fixed, HeapSyn ticks and allocations (the deterministic columns):

| bench | ticks before | ticks after | Δ ticks | allocs before | allocs after | Δ allocs |
|---|--:|--:|--:|--:|--:|--:|
| block_merge | 82,541,872 | 82,541,810 | −62 | 480,264 | 480,248 | −16 |
| import_export_yaml | 31,967,231 | 31,967,231 | — | 1,942,766 | 1,942,766 | — |
| import_export_toml | 31,793,907 | 31,793,907 | — | 1,306,616 | 1,306,616 | — |
| string_scale | 45,995,706 | 45,890,729 | −0.23% | 395,650 | 369,407 | **−6.6%** |
| list_scale | 33,156,485 | 33,156,485 | — | 160,676 | 160,676 | — |
| lookup_curve | 1,443,585 | 1,264,965 | **−12.4%** | 5,595,923 | 5,554,583 | −0.7% |
| io_loop | 2,962,309 | 2,962,309 | — | 545,675 | 545,675 | — |
| hof_fold | 84,700,417 | 77,500,446 | **−8.5%** | 8,700,143 | 7,200,150 | **−17.2%** |

Neutral or better everywhere, never worse — sharing removes duplicated thunk allocation rather than adding cost. **Confidence: measured-single**, this machine, per `docs/superpowers/engine-ab/PROTOCOL.md`. No quiet-machine or release-grade claim is made; that session is eu-1hcw and owner-only. The engine-ab ledger (`results.jsonl`) was deliberately **not** committed — both runs were reverted.

All eight benches also pass as harness tests under both engines: `cargo test --test harness_test bench` 8/8, `EU_HEAPSYN=1 cargo test --test harness_test bench` 8/8.

## Tests

**`tests/harness/195_gua64_inline_arg_sharing.eu`** — pure user code, not a prelude combinator, ascending input so the duplicated arm is genuinely taken. Ten checks feeding `RESULT` in the `189_r9oy_union_as_spec.eu` style: the canonical two-use helper, the descending control, three-use and single-use binders, an atomic argument (must *not* be let-bound), an unused parameter that must stay unforced, two uses inside a nested block scope, two uses inside a lambda body closing over an enclosing binding, and a non-atomic argument closing over the enclosing scope.

Because this is a *complexity* regression the verdict alone cannot gate it — the duplicating compiler computes the same answers, just exponentially slower. `test_195_gua64_inline_arg_sharing` therefore runs the file as a subprocess under a **180 s deadline first**, so the defect makes the test *fail* rather than making the suite *hang*, and only then applies the ordinary harness verdict. `size: 30` puts a correct compiler at ~0.15 s (debug, measured) against 2^29 steps for a duplicating one — three orders of magnitude either side of the deadline, which is what makes it robust to a loaded machine.

**Three `core::inline::reduce` unit tests** give a deterministic, instant gate on the core shape: one pinning that the reduced body holds exactly one copy of each argument, one pinning the *boundary* (duplicable and single-use arguments must still be substituted directly — the `Let` is a cost, not a default), and one pinning the call-site `Smid` on both the sharing `Let` and the application it wraps.

### Fault injection — four sites, individually, rebuild each time

Per-site, not wholesale. Fix restored and re-verified green after each.

| # | Injected fault | Result |
|---|---|---|
| 1 | occurrence guard `uses > 1` → `uses > 100` (never share) | `test_195` **FAILS** on the 180 s deadline; `test_beta_reduce_shares_duplicated_non_atomic_argument` **FAILS** |
| 2 | `duplicable` narrowed to drop `Expr::Var` (over-share) | `test_beta_reduce_does_not_share_duplicable_or_single_use_arguments` **FAILS** |
| 3 | sharing `Let` given `Smid::default()` | `test_beta_reduce_sharing_let_keeps_call_site_smid` **FAILS** |
| 4 | `retag_call_site` removed from the reduced body | `test_beta_reduce_sharing_let_keeps_call_site_smid` **FAILS** |

Site 1's failure message names the defect and the function, so a future regression is self-explaining rather than a mystery timeout.

## Blast radius

**`tests/harness/errors/` corpus** — all 179 files, stderr and exit code captured with the release binary before and after, in **both** prelude modes, diffed with ANSI stripped.

- **Blob mode (the shipping default): zero diagnostic differences.** The only textual diff in the whole corpus is a PID in `171_yaml_large_uint.eu`'s pre-existing panic message.
- **Source-prelude mode: 178/179 identical.** One genuine change, reported rather than papered over — no `.expect` file was edited anywhere:

  `error_176.eu`'s primary label moves from `16:22` (inside the callee, `"hello {name}"`) to `19:32` (the user's `//=` call site). Blob mode already reported `19:32` and is unchanged; the source path was the outlier and has moved *into agreement with it*, which is the direction this release wants. The mechanism is visible and benign — `//= = λ(e, v). EXPECT(e, DBG_REPR(v), EQ(e, v), TRUE)` uses `e` twice, so `make-greeting("world")` is now let-bound and evaluated once instead of twice.

  The honest caveat: source mode also **loses** the `called from here` secondary label and the one-line stack trace on that diagnostic. Root cause identified: `target_annotation` (`src/eval/machine/vm.rs:1404`) only matches a top-level `HeapSyn::Ann`, so when a thunk's body becomes `Let(…, Ann(App))` instead of `Ann(App)` the leading `Ann` is hidden and the `Update` continuation is never stamped. Teaching it to look through a leading `Let`/`LetRec` would restore it generally, but that is trace-curation machinery outside this bead's fence and would change diagnostics broadly — **filed as follow-up rather than changed here**. The `.expect` sidecar asserts `got "hello world"` and is unaffected; the test passes in both modes.

**Test suites**

- **Source-prelude mode — fully green**, which is what CI's main `test` job actually runs (it generates no blob): 507/507 harness, 1356 lib, every integration target. `./target/release/eu test --allow-io tests/harness` (the `test-source-prelude` job) exits 0, all PASS.
- **Blob mode** — exactly the three known pre-existing failures and nothing else: `debug_trace_restores_the_uncurated_trace`, `secondary_labels_do_not_excerpt_library_internals` (eu-7x0r, both already excluded by name in `build-rust.yaml`) and `test_193_1tkk_7_12_curated_trace`. 506 harness tests pass including the new one. Confirmed pre-existing by reverting both source files to `origin/master` and rebuilding.
- **HeapSyn engine** — `EU_HEAPSYN=1 cargo test --test harness_test`: 506 pass, same single pre-existing failure.
- `cargo clippy --workspace --all-targets -- -D warnings` clean, no `#[allow]` added. `cargo fmt --all --check` clean.
- `scripts/test-doc-examples.py`: 245 passed, 0 failed, 607 blocks.

## Related, and deliberately not touched

- **eu-pazes** (prelude `max-of`/`min-of` over `foldl`) fixes two *victims* of this defect. Independent; not depended on and not duplicated here.
- **eu-frf2o** (annotated inline block literal in operand position → blackhole) was checked during this task at the coordinator's request and is **unaffected either way** by this change — identical error before and after, and the two passing controls still pass. It is not a core-pass bug: core is correct through `reflatten`; the fault is in STG compilation, where a binder group whose bindings reference siblings is emitted as `Let` rather than `LetRec`. Localised and handed off.

## Documentation

`CHANGELOG.md` under Unreleased → Fixed, and a new *Argument sharing in beta reduction* subsection in `docs/architecture.md` stating the rule, the scope handling, and the non-recursion argument above.

---

Do **not** merge — the owner holds all merge decisions this cycle.


gmorpheme:
Follow-up filed as **eu-7fjiq** (P2, bug) for the one diagnostic finding in this PR — `target_annotation` (`src/eval/machine/vm.rs:1404`) only matches a *leading* `HeapSyn::Ann`, so a thunk whose body is wrapped in a binding group loses the annotation that stamps the `Update` continuation, and with it the secondary label and stack trace. Suggested fix is to look through a leading `Let`/`LetRec`.

Deliberately not done here: `vm.rs` trace machinery was fenced off from eu-gua64, and widening annotation resolution changes diagnostic output across the corpus and deserves its own before/after capture in both prelude modes rather than riding along on an inliner fix.

Blob mode — the shipping default — is entirely unaffected: all 179 `errors/` files byte-identical.

gmorpheme:
Additional evidence for the *"is the sharing `Let` safe?"* question, since eu-frf2o turns on exactly this — the argument in the description is by construction, but it is also checkable empirically. `eu dump stg` on the reproduction:

```
[2] letrec [0] thunk @[57] case ✳2 of
             Cons → ✳0
             Nil → ⊗186()
             _ → HEAD(✳0);
           [1] thunk @[59] let [0] thunk @[61] case ✳2 of
                                 Cons → ✳1
                                 Nil → ⊗187()
                                 _ → TAIL(✳0)
           in seq ✳0 in →✳5(✳0)
in seq ✳0 in seq ✳1 in @[55] let [0] ...
```

The group compiles to a **`letrec`**, not a non-recursive `Let` — so even the hypothetical sibling-reference hazard cannot arise here, independently of the construction argument. Belt and braces.

The `@[55]`/`@[57]`/`@[59]`/`@[61]` annotations visible in the same fragment are the call-site `Ann` nodes surviving through the transformation, which is the other half of the diagnostics story.

Also verified that the harness test's *own* verdict gates, separately from the fault injection of the fix: breaking a single check in `195_gua64_inline_arg_sharing.eu` (`t3: thrice(2 + 3) = 51`) makes `test_195_gua64_inline_arg_sharing` FAIL on the `RESULT` assertion in 2.6s, then restored and green. So both halves of that test bite — the deadline for the complexity regression, and `RESULT` for correctness.

gmorpheme:
## Recorded review (non-author, CLAUDE.md) — eu-gua64

Reviewed by Wicket at `2319d7af`. **Recommend merge.**

### The boundary is correct, and it is the right boundary

The predicate is `uses > 1 && !duplicable(&arg)` → let-bind; everything else substitutes directly. So an argument that is duplicable (a `Var`, `Literal` or `Intrinsic`) or whose binder occurs at most once still goes straight in, and the inliner's existing behaviour is preserved on the overwhelming majority of call sites. That is the conservative direction: the change only ever *adds* sharing where work was being duplicated, never removes an inlining opportunity.

`duplicable` is right to include `Var`: a variable already names a single shared thunk, so copying the reference duplicates a pointer, not a computation. The `uses == 0` case falls through to direct substitution, which drops the argument expression — correct under laziness and unchanged from before.

### The hazard I went looking for, and why it is not there

I did not want to take the "the sharing `Let` is genuinely non-recursive" argument on trust, particularly given eu-frf2o (#1087) turns on exactly that distinction. The specific attack I tried to construct: `fresh_name`'s `avoid` set is built from the body's *free* variables plus the args' free variables, so a name **bound inside** the callee body is not in it. If the callee body contained an inner binding literally named `__shared_a`, could the minted name collide and the substituted occurrences get captured by the inner scope?

**No, and for a structural reason rather than luck.** Core keeps bound variables as de Bruijn indices, and `close_let_scope` → `close_expr_vars` converts only variables that are *free by name*. An inner binding's own references are already bound, so they are invisible to that conversion. The only free-by-name occurrences of the fresh name in the reduced body are precisely the ones the substitution just introduced. The two ways a *free* collision could arise — a free name in the body, or a free name in an argument — are both in `avoid`. So the set is exactly right, not accidentally sufficient.

I wrote a eucalypt probe for this (`g(a): { __shared_a: 7 }.(a + a + __shared_a)`, exploiting the documented rule that block bindings shadow outer scope) and it returned the correct `11`. I am reporting that as weak evidence only: `eu dump inlined` showed the inliner did not beta-reduce that call site, so the probe did not actually exercise `beta_reduce`. The de Bruijn argument above is the real basis for my conclusion.

### Other checks

- `visit_free_vars` / `free_vars` refactor: `free_vars` is reimplemented on top of the new visitor with no behavioural change, which is what the diff shows.
- Sharing a `Let` rather than copying changes how many times an expression is evaluated, never what it evaluates to, under laziness. Errors and effects that would have been raised twice are raised once — observationally identical, since only the first is ever seen.
- No architectural smell: the fix is a general occurrence-count guard, not a special case for the reported victim.
- `retag_call_site` correctly re-tags with the caller's Smid so diagnostics blame the call site, and the new `Let` carries `*call_smid` too.

### What I did NOT verify

- **I did not independently reproduce the 2^n blow-up or re-measure the before/after figures.** The tick counts and `eu dump inlined` output in the PR body are the author's; I read them for coherence and they are internally consistent, but I did not re-run them. Given this project's history with unverified performance claims, someone should re-measure on a quiet machine before any figure from this PR enters a release note.
- I did not verify the fresh-name scheme against a callee body containing a *free* variable named `__shared_<binder>` from an enclosing scope — the code covers it via `avoid`, and I read that code, but I did not construct the case.
- I relied on the PR's CI for the test matrix, clippy and fmt.

gmorpheme:
## Gatekeeper review — approved with one correction applied, merging on green

`783aeca6`, with `lib/prelude.blob` regenerated after the branch switch and

---

### One correction applied: a CHANGELOG figure that does not reproduce

The entry claimed the canonical suite showed **`lookup_curve` −12.4% ticks**.
I re-measured the whole `bench/{015..022}` suite myself, before and after, on
deterministic VM ticks and allocation counts (`eu -S`), and that figure does
not reproduce — I measure **−0.09%**.

Measured, both trees, same binary shape:

| bench | ticks before | ticks after | Δ ticks | allocs before | allocs after | Δ allocs |
|---|--:|--:|--:|--:|--:|--:|
| 015 block_merge | 116,221,655 | 116,221,580 | −75 | 620,340 | 620,316 | −24 |
| 016 yaml | 31,462,013 | 31,462,013 | **0** | 1,815,053 | 1,815,053 | **0** |
| 017 toml | 31,288,675 | 31,288,675 | **0** | 1,178,837 | 1,178,837 | **0** |
| 018 string_scale | 45,747,689 | 45,663,706 | −0.18% | 395,703 | 369,460 | **−6.6%** |
| 019 list_scale | 33,006,893 | 33,006,893 | **0** | 160,716 | 160,716 | **0** |
| 020 lookup_curve | 328,404,994 | 328,113,274 | −0.09% | 11,192,557 | 11,109,877 | −0.74% |
| 021 io_loop | 1,512 | 1,512 | **0** | 506 | 506 | **0** |
| 022 hof_fold | 76,500,835 | 70,500,858 | **−7.8%** | 8,700,183 | 7,200,190 | **−17.2%** |

So: **`hof_fold` −17.2% allocations and `string_scale` −6.6% allocations
reproduce exactly**; `hof_fold`'s tick figure is close (−7.8% against a
claimed −8.5%); and `lookup_curve` −12.4% ticks is off by more than two orders
of magnitude.

The headline claim — **neutral-or-better on ticks, never worse** — is true, and
that is the claim that matters. But an unmeasured figure is exactly the kind of
thing that gets carried verbatim into release notes, so I have corrected the
sentence in the CHANGELOG to the re-measured numbers, stated the method, and
said plainly that the earlier figure was corrected rather than quietly dropping
it.

Note ticks and allocation counts are **build-profile-independent**, so a debug
measurement is as good as a release one for these — which is precisely why they
are the right currency for a claim like this and wall-clock is not.

**The reproduction table in the entry I did not touch**: it is labelled
`HeapSyn -S`, and my numbers differ on ticks (I measured the default bytecode
engine, and against a newer master) while the **allocation column reproduces to
within 0.25%** — 246,654 against a claimed 246,050 at N=16. The load-bearing
property is the shape, and that reproduces exactly: ×4 per +2 in N before,
flat-linear after.

Measured myself, whole fixture, default engine:

```
before:  N=16 1,169,309 ticks   N=18 4,659,607   N=20 18,619,281   (×4 per +2 = 2^N)
after:   N=16     6,802 ticks   N=18     7,408   N=20      8,014   N=24 9,226
```

`+606` ticks per `+2` in N. Exactly linear. And at the fixture's own `size: 30`
the unfixed compiler **does not finish in 120 seconds**, while the fixed one is
instant.

---

### The fix

`beta_reduce` substituted each argument at *every* occurrence of its binder.
Copies of a non-atomic argument become separate thunks that share nothing, so a
callee mentioning a parameter twice doubles the work — and when the argument is
the caller's own recursive call, `T(n) = T(n-1)` becomes `T(n) = 2·T(n-1)`.

The guard is `uses > 1 && !duplicable(arg)`, and both halves are right:

- `duplicable` admits `Var`, `Literal` and `Intrinsic`. A variable already names
  a single shared thunk and the other two cost nothing to re-evaluate, so
  copying them is genuinely free. Anything else is a fresh thunk per copy.
- `uses <= 1` substitutes directly, so the common case pays nothing — no
  spurious `Let`, no regression on code that was fine. The zero-use case still
  drops the argument unforced, which fixture check **t6** gates with a `panic`
  that fires if it is ever evaluated.

**Name capture — the thing I most wanted to be wrong about, and it holds.**
`fresh_name` avoids only the free names of the body and the arguments, not names
*bound inside* the body. That looks like a gap: a previous reduction could have
left a `__shared_a` binder inside. It is safe, and specifically because of the
IR: a name bound inside the body is `Var::Bound` with a de Bruijn index, not
`Var::Free`, so `close_let_scope` cannot capture it. The only free occurrences
of the fresh name are the ones just substituted. Worth stating explicitly since
the safety is not local to `fresh_name`.

`close_let_scope` handles the index shift for both the body and the bound
argument expressions crossing the new scope boundary — fixture checks **t7-t10**
are the ones that would catch a mistake there, covering uses inside a block,
inside a lambda body whose argument closes over an enclosing binding, and an
argument that closes over the enclosing scope.

### Semantic-equivalence checklist (perf PR)

- **No `Smid::default()` in new code** — the reference carries `arg.smid()`,
  the sharing `Let` carries `*call_smid`. ✓
- **Source location preserved** — `retag_call_site` runs before the wrap, so
  the reduced `App` keeps the call site *and* the enclosing `Let` carries the
  same Smid. `test_beta_reduce_sharing_let_keeps_call_site_smid` asserts both,
  which matters because the 0.14 diagnostics work (eu-og3u6, eu-1tkk.7) depends
  on that blame landing at the user's call site. ✓
- **Evaluation order unchanged** — a `Let` binding is a thunk, so nothing is
  forced earlier or later. ✓
- **Error paths unchanged** — no new error construction. ✓
- **Edge cases tested** — zero-use, single-use, duplicable, three-use, nested
  scopes, closure over enclosing scope. ✓

`free_vars` refactored into `visit_free_vars` with a callback is the right
shape: occurrence counts and the name set now come from **one** traversal
definition, so they cannot drift apart. `free_vars` is re-expressed in terms of
it rather than kept as a parallel implementation.

### Tests

- The harness fixture has **ten** checks, all inside the `RESULT` verdict.
- The complexity gate is separate and deliberately runs **first**: with the
  defect present the subprocess is killed at the deadline and the test fails,
  rather than the suite hanging — which is the difference between a gate and a
  timeout.
- The deadline has three orders of magnitude of headroom either side (~0.15s
  correct, 2^29 steps incorrect), so a loaded build machine cannot flip it.
- The note on input order in the fixture is doing real work: the blow-up needs
  the duplicated arm to be *taken*, so `ascending` gates and `descending` is the
  control. Without that comment a later editor could "simplify" the fixture into
  something that no longer reproduces.
- Three unit tests pin the reduced core shape, the no-over-sharing boundary
  (duplicable and single-use arguments must **not** be let-bound — the sharing
  `Let` is a cost, not a default), and the call-site `Smid`.

Disabling the guard (`if false && uses > 1 && …`) in my own worktree:

```
thread 'test_210_gua64_inline_arg_sharing' panicked:
210_gua64_inline_arg_sharing.eu did not complete within 180s
test result: FAILED. 0 passed; 1 failed; 526 filtered out; finished in 180.06s
```

Restored: passes in **2.76s**. The gate is real, and it fails as a deadline
failure rather than a hang.

### Fixture renumbering

`195_gua64_inline_arg_sharing` → **`210_...`**. 195 is
`errors/195_og3u6_trace_anchor`; `tests/harness/` and `tests/harness/errors/`
share one number sequence, and 196-209 are taken. Prose references in the
fixture header and the CHANGELOG updated too.

### Gates

- `cargo fmt --all` — clean
- `cargo test` — **0 failed across every test binary**, lib 1401, harness 527

PR #1087, opened 2026-07-27
fix(stg): emit LETREC for the Meta binder group (eu-frf2o)

gmorpheme:
**Do not merge — owner review.**

Fixes **eu-frf2o** (P0). Blocks eu-lzfj5.

## The defect

```
p: { :foo k: 1 } raw-meta
error: infinite loop detected: binding refers to itself
```

`compile_binding`'s `Expr::Meta` arm builds three bindings into the current
`LetBinder` — the metadata, the annotated body, and the `Meta` node naming
both — but never called `ensure_recursive()`. The group was therefore
realised as a plain `StgSyn::Let`, and `EnvBuilder::from_let`
(`src/eval/machine/env_builder.rs:175`) closes every binding over `next`,
the *enclosing* frame. The `Meta` node's two operands consequently indexed
the outer scope rather than its siblings.

`eu dump stg` before:

```
let [0] Sym(!:foo);
    [1] thunk letrec ... in Block(...);
    [2] `✳0 ✳1                       <-- names [0] and [1], but under LET
in demeta ✳2 of meta → ✳0 _ → ⊗83
```

after:

```
letrec [0] Sym(!:foo);
       [1] thunk letrec ... in Block(...);
       [2] `✳0 ✳1
in demeta ✳2 of meta → ✳0 _ → ⊗83
```

### It was not only a crash

The blackhole is just what you get when outer slot 0 happens to be the
thunk being forced. Inside a function it is a live parameter:

```
probe(a, b): { :foo k: 1 } raw-meta
p: probe(:AAA, :BBB)
---
p: AAA
```

`:AAA` is `probe`'s first argument. No error, no diagnostic — silent data
corruption from valid source. This is why the bead is P0 and why every
check in the regression test asserts a **value** rather than the absence of
an error.

## Which fix, and why

The brief offered two shapes: emit `LetRec`, or build the `Meta` node's refs
against the outer environment so that `Let` stays correct. **The group
genuinely has internal references, so `LetRec` is the honest fix.**

`m` and `b` are results of `compile_binding` *on this binder* — the metadata
symbol is added by `binder.add`, the body by `binder.add_deferred`. They are
siblings by construction. Making the group reference-free would mean
allocating the two operands in a separate enclosing `Let` and the `Meta`
node in a nested one: two frames where one suffices, for no benefit. This
is not a `LetRec` that exists to make escaping references resolve by
accident — under `Let` the references are not "resolving elsewhere", they
are simply wrong.

`ensure_recursive` is also the codebase's established, documented mechanism
for exactly this ("If we're adding inter-referring bindings we need to
ensure that however the binder was created, it will generate a LETREC").
`compile_block`, `compile_list_body`, `compile_list_binding` and
`compile_lookup` all call it. The `Expr::Meta` arm was the one that did not.

That also explains the puzzle of why block-shaped metadata worked:
`{ {doc: "d"} k: 1 }` compiles its metadata through `compile_block`, which
calls `ensure_recursive()` as a side effect. The correct arm was never
"already doing the right thing" — it was being rescued by an unrelated call.
`{ :foo }` works for the same accidental reason inverted: an empty body
needs no allocation group, so there is no sibling reference to get wrong.

## Sibling-group audit

The most valuable part of the task, so in full.

**`compile_binding` — every arm that can add a sibling-referencing binding:**

| arm | sibling refs? | `ensure_recursive`? |
|---|---|---|
| `Expr::Literal` | no (no refs at all) | n/a |
| `Expr::Var` | no — deferred `ProtoVar`, resolved against the binding context | n/a |
| `Expr::Let` | no — `ProtoLet` opens its own binder | n/a |
| `Expr::Lam` | no — `ProtoLambda` opens its own scope | n/a |
| `Expr::App` | no — `ProtoAppGroup` opens its own `synthetic_let` | n/a |
| `Expr::Intrinsic` | no — `Ref::G` | n/a |
| `Expr::List` | yes | yes |
| `Expr::Block` | yes | yes |
| `Expr::Lookup` | yes | yes |
| **`Expr::Meta`** | **yes** | **no — this bug** |
| `Expr::Operator` | delegates | n/a |

`Expr::Meta` was the only offender. **No second instance was found.**

**Other `StgSyn::Let` construction sites in the compiler:**

- `ProtoInline` (`compiler.rs:1512`) and `ProtoFusedInline` (`:1556`) build
  `Let` groups of `value(atom(r))` where every `r` was resolved with
  `take_reference(context)` against the *enclosing* context. No binding names
  a sibling; only the inlined wrapper body (the let body) does, and a let
  body always sees the new frame. `Let` is correct here.
- `compile_body`'s `Expr::Meta` arm (`:1735`) builds the same `with_meta`
  node, but as the **let body**, which `into_stg` always realises in
  `body_context`. The frame is in scope regardless of recursiveness, so this
  arm is correct unchanged and is deliberately left alone.

**Hand-written STG wrappers** (`block.rs`, `render.rs`, `string.rs`, `eq.rs`,
`force.rs`, `array.rs`, `constant.rs`, `wrap.rs`, `meta.rs`): every
multi-binding `let_(...)` was reviewed. In each case the bindings reference
the enclosing frame only and the body references the group — the correct use
of `Let`. Specifically `eq.rs:68` and `eq.rs:149` (`Eq(xh,yh)`, `Eq(xt,yt)`
over a destructured cons, body `And`) and `render.rs:351` (two `Eq` probes
on the same outer `export_sym`, body `Or`). Everything in `block.rs`,
`string.rs`, `force.rs`, `array.rs`, `constant.rs` and `wrap.rs` is
single-binding.

**Optimiser:** `LetIndexTransformation::from_letrec_bindings`
(`optimiser.rs:56`) explicitly refuses to strip a `LetRec` whose bindings
reference the group (`if k < n { return None }`), so the optimiser can never
demote a recursive group with internal references to a `Let`.

**One observation, not fixed here.** `src/eval/stg/syntax.rs`'s `pub mod ex`
is `#[cfg(test)]`, appears to be entirely unreferenced, and its combinators
do not match their names: `ex::s()` computes `g(x)(f(x))` where the S
combinator is `f(x)(g(x))`, and `ex::compose()` returns `f(λx.g(x))` rather
than `λx.f(g(x))`. Dead test fixtures with no runtime reach — flagged for a
follow-up bead rather than touched in a P0 fix.

## Making the invariant self-enforcing

Rather than rely on remembering to call `ensure_recursive()`, `LetBinder::add`
now carries a debug assertion for the underlying rule: **already-determined
syntax embedding a `Ref::L` can only be naming a sibling**, because the
compiler obtains a concrete `Ref::L` solely from this binder's own `add` /
`add_deferred` / `running_ref`, while enclosing-scope references arrive as
deferred `ProtoVar`s. The helper `has_flat_local_ref` models only the flat
node shapes `add` is ever handed and reports "no local ref" for anything
that opens a binding scope, so it can under-report but never raise a false
alarm. Zero cost in release; it fires in `cargo test` the instant a future
arm repeats this mistake. It does not fire anywhere in the current suite.

## Regression test

`tests/harness/192_frf2o_inline_annotated_block.eu`, 17 checks, `RESULT`
computed from all of them via `all-true? then(:PASS, :FAIL)`.

Broken shapes (assert values): annotation off an inline literal; multiple
declarations in the annotated body; the body surviving via `keys`, `values`,
`lookup`, `elements`; **three function-scope cases** including
`second-arg-probe(x, y): [x, { :foo k: 1 } raw-meta]` so an escape onto
parameter 1 is caught as well as parameter 0.

Controls that already worked and must keep working: block metadata;
`"doc"` shorthand; `:suppress`; the `//` route; `{ :foo }`; bound-to-a-name
first (both annotation and body); `meta` through the demeta wrapper.

Plus two unit tests in `compiler.rs`:
`test_meta_argument_compiles_to_letrec` (the annotated body is a Core `Let`,
not a `Block` — a `Block` body would mask the defect by calling
`ensure_recursive()` incidentally, and an earlier draft of this test did
exactly that and passed against the broken compiler) and
`test_non_recursive_binder_rejects_sibling_reference`.

## Fault injection — per site, rebuild each time

All of this was redone from scratch after the machine's disk-full episode,
against fresh builds, using `Edit`/`git diff` only. **No `git stash` was
used at any point in this task.**

**Baseline (fix in place):** unit tests 10/10 pass; `test_192_frf2o` passes.

**Injection A — `binder.ensure_recursive()` removed from the `Expr::Meta` arm:**
- `cargo test --lib eval::stg::compiler` → `test_meta_argument_compiles_to_letrec`
  **FAILED**
- `cargo test --test harness_test test_192_frf2o` → **FAILED**
- release build (no debug assertions, i.e. how the bug shipped): the fixture
  raises `infinite loop detected` at `t1`, and `eu test` reports
  `192_frf2o_inline_annotated_block...FAIL`
- release build, **function-scope checks only** — the silent case, which is
  the one that matters:
  ```
  ---
  raw: AAA
  t7: false
  RESULT: FAIL
  ```
  No error raised; the value assertion is what catches it. A fixture that
  only demanded "no blackhole" would have passed here.
- restored, rebuilt → all pass.

**Injection B — the `LetBinder::add` debug assertion neutered
(`|| true`), `ensure_recursive()` left in place:**
- `test_non_recursive_binder_rejects_sibling_reference` → **FAILED**
  ("test did not panic as expected")
- `test_192_frf2o` → passes, as expected: the guard is a separate safety net,
  not the fix.
- restored, rebuilt → all pass.

After restoring, `git status` and `git diff` are both empty against the
commit.

## Validation

- `cargo test --no-fail-fast`, **blob mode**: the three known pre-existing
  master failures and nothing else —
  `debug_trace_restores_the_uncurated_trace`,
  `secondary_labels_do_not_excerpt_library_internals` (both eu-7x0r) and
  `test_193_1tkk_7_12_curated_trace`. Harness 506 pass / 1 fail (that one).
  These three are **real on master under a blob — the shipped
  configuration — and CI cannot see them**; they are fixed by #1079. Do not
  read this PR's green board as evidence that blob mode is healthy. See
  eu-5yn9v, and the note below.
- `EU_SOURCE_PRELUDE=1 cargo test --no-fail-fast`: harness **507/507 pass**.
  The two `diagnostics_blame_plumbing_test` failures under this env var are
  `blob_mode_*` tests that require a blob by construction; CI's
  `test-source-prelude` job runs only the harness, so this is an artefact of
  forcing the variable across the whole suite, not a regression.
- Release harness, zero failures in each of: default (bytecode + blob);
  `EU_HEAPSYN=1`; `EU_SOURCE_PRELUDE=1`; `EU_SOURCE_PRELUDE=1 EU_HEAPSYN=1`.
- `EU_GC_VERIFY=2 EU_GC_STRESS=1` over the full harness: zero failures, zero
  panics, on **both** engines. The new fixture also runs clean under
  `EU_GC_VERIFY=2 EU_GC_STRESS=1 EU_GC_POISON=1`.
- `cargo fmt --all`; `cargo clippy --workspace --all-targets -- -D warnings`
  clean, no `#[allow]`.

## The green board does not cover blob mode (eu-5yn9v)

Stated explicitly so no reviewer draws the wrong conclusion from 19/19
green. I drew it myself in an earlier draft of this write-up and the
coordinator corrected me; the correction is right, and my own results above
already disproved my inference — the identical machine ran **507/507 under
`EU_SOURCE_PRELUDE=1`** and red under a blob, so the three failures are
blob-conditioned, not platform-conditioned.

Three separate mechanisms each hide this class, and CI is blind to all three
of the failures listed above:

1. **`Test Suite` is deliberately blob-less.** Its only test steps are a bare
   `cargo test` / `cargo test --lib`; there is no `prelude-compile` anywhere
   in the job. The workflow says so in its own words — "the plain 'Test
   Suite' job (deliberately blob-less, to keep the source-prelude fallback
   exercised)". Its greenness is evidence about the *source-prelude*
   configuration and nothing else.
2. **The blob harness jobs do not validate the `errors/*.expect` sidecars.**
   `Bytecode + blob harness` runs `eu test --allow-io tests/harness`, i.e.
   the in-language `.eu` assertions only. `test_193_1tkk_7_12_curated_trace`
   is an *error* test — `tests/harness/errors/193_1tkk_7_12_curated_trace.eu`
   with a `.eu.expect` sidecar asserting exit code and stderr, driven through
   `run_error_test` → `tester::error_test_with_binary` from
   `harness_test.rs`. Only `cargo test` checks it, and `cargo test` never
   runs under a blob.
3. **The blob harness jobs never run `diagnostics_invariants` at all**, which
   is where the two eu-7x0r failures live.

And a fourth, worth naming because it is the one job that looks like it
should catch them: **`Diagnostics blob-mode integration` does build a blob,
but scopes its run to `cargo test --test diagnostics_invariants
corpus_satisfies_invariants -- --exact`** — deliberately excluding
`secondary_labels_do_not_excerpt_library_internals` and
`debug_trace_restores_the_uncurated_trace` by omission, with a workflow
comment naming both and eu-7x0r as the reason.

None of this is caused by or related to this PR; it is recorded here only so
the CI result is read for what it actually covers.

## Note on the bead's third acceptance criterion

`tests/harness/191_3aa6s_meta_symbol_metadata.eu` does not exist on master —
it arrives with PR #1081 (eu-3aa6s, meta-fix), whose files are fenced off
from this branch. Removing its workaround comment belongs to that PR or a
follow-up, not here.

Relatedly: `{ :foo k: 1 } meta` still raises `type mismatch: expected block,
found symbol` on this branch — but it now does so **identically to the
bound-to-a-name control** `named: { :foo k: 1 }; named meta`, which fails
the same way on unmodified master. That is eu-3aa6s's defect in
`src/eval/stg/meta.rs` (fenced), not eu-frf2o's. With this PR the inline and
bound forms are indistinguishable, which is the correct scope boundary; the
bead's `meta` criterion is satisfied once #1081 lands.


gmorpheme:
Reminder for any agent reading this: **do not merge — owner review this cycle.**

gmorpheme:
## Recorded review (non-author, CLAUDE.md) — eu-frf2o

Reviewed by Wicket. Verified in an isolated worktree at `5f1f9afc` with a release build. **Recommend merge.**

### Independently fault-injection verified — I did not take the author's word for it

Ran the bead's own reproduction, then removed the fix (the single `binder.ensure_recursive()` at `src/eval/stg/compiler.rs:1895`), rebuilt, and re-ran.

| | with the fix | fix removed |
|---|---|---|
| `p: { :foo k: 1 } raw-meta` | `p: foo` | `error: infinite loop detected: binding refers to itself` |
| `probe(a,b): { :foo k: 1 } raw-meta` called as `probe(:AAA,:BBB)` | `p: foo` | **`p: AAA`, exit 0, no diagnostic** |
| `test_192_frf2o_inline_annotated_block` | `ok` | `FAILED` |

The second row is the one that matters and it reproduces exactly as the bead describes: valid source, clean exit, no error, and the answer is the caller's first argument. That is silent data corruption, and P0 is the right severity. The harness test genuinely gates — it fails when the code under test is broken, which is the property that actually matters and the one I checked directly rather than inferring.

### The fix is the right shape

`LetRec` is the honest fix and the PR argues it correctly. `m` and `b` are both produced by `compile_binding` *on this binder*, so they are siblings by construction; under a plain `Let`, `EnvBuilder::from_let` closes every binding over the enclosing frame and the `Meta` node's two operands index the outer scope. Those references are not "resolving elsewhere", they are simply wrong. The alternative — building the operands against the outer environment so `Let` stays valid — would cost two frames where one suffices, to preserve an encoding that has no advantage.

**The `debug_assert!` + `has_flat_local_ref` guard is the part I most want to keep.** The one-line fix closes this instance; the assertion closes the *class*, by making any future binder group that adds a sibling-referencing node under a non-recursive `Let` fail loudly in debug and test builds instead of silently miscompiling. Given the symptom here was a plausible wrong value rather than a crash, a general tripwire is worth more than the fix itself. This is the opposite of the special-casing I am meant to flag, so: noted approvingly.

I also checked that the guard is not vacuous — `has_flat_local_ref` covers `Atom`, `Cons`, `Bif`, `App`, `DirectApp`, `Meta`, `LookupLit`, and recurses through `Ann` and `Seq`. The `_ => false` default means a future `StgSyn` variant carrying a `Ref::L` would not be caught, which is a small residual gap, but the enumerated set covers every variant that can currently hold one.

### What I did NOT verify

- I did not audit the four other `ensure_recursive()` call sites (`compiler.rs:1930/2020/2048/2075`) for correctness; I only confirmed the Meta arm was the one missing it.
- I did not check whether any *other* binder group is currently missing `ensure_recursive` — the new `debug_assert!` is what will answer that, across the whole test suite, and CI is green with it in place, which is itself mild evidence there are no other instances on tested paths.
- I ran a release build for the reproduction and `cargo test --release` for the harness test; I did not run the full matrix locally and am relying on the PR's CI for clippy/fmt/cross-platform.
- Interaction with #1081: per the queue plan both are required for eu-frf2o's acceptance and they touch different files. I have not yet verified the combination — I will when #1081 lands ahead of this.

gmorpheme:

Reviewed in an isolated worktree, on this branch merged onto master `6484bcc3`,
with `lib/prelude.blob` regenerated after the branch switch and

### The defect, and why the one-line fix is the right one

`compile_binding`'s `Expr::Meta` arm builds three bindings — the metadata, the
body, and the `Meta` node naming both — but never called `ensure_recursive()`.
Under a plain `Let` every binding is closed over the *enclosing* frame, so the
`Meta` node's two operands indexed the outer scope instead of its siblings.

The two symptoms follow directly from what happened to be in the outer slot:
at the top level it was the thunk being forced, giving a blackhole; inside a
function it was a live parameter, giving a **plausible wrong value** and exit
0. The second is much the worse of the two, and it is what makes this a P1
rather than a cosmetic fix.

`compile_block`, `compile_list_*` and `compile_lookup` all already call
`ensure_recursive()`. This arm is the outlier, so adding the call is
restoring an existing invariant rather than inventing a special case — which
is the distinction I look for before accepting a one-line fix in the STG
compiler.

The explanation of *why nobody hit this sooner* is the part I checked hardest,
because a fix whose "why now" story is wrong usually has the mechanism wrong
too. It holds up: block-shaped metadata worked only because compiling the
metadata block called `ensure_recursive()` **as a side effect**. That is why
the new unit test deliberately uses a Core `Let` body rather than a bare
`Block` — a bare `Block` would mask the defect by accidentally fixing the
group. That comment in the test is doing real work and I would not want it
edited out.

I did not take the PR's word for it. Removing the `ensure_recursive()` call in
my own worktree and rebuilding:

- simple case → `infinite loop detected: binding refers to itself`
- probe case → `p: AAA`, **exit 0** — silent corruption, the caller's first
  argument surfacing as the block's value

With the fix restored, both cases give `p: foo`. The harness fixture **FAILED**
without the fix and passes with it. It genuinely gates.

### The debug assertion is the better half of this PR

`LetBinder::add` now carries a `debug_assert!` that syntax already naming a
sibling may only be added to a recursive group, with `has_flat_local_ref` as
the predicate. This is what stops the *next* call site reintroducing the same
defect, and it is deliberately conservative in the right direction:
scope-introducing nodes (nested let, case branch, lambda form) shift de Bruijn
indices and are reported as "no local ref", so the check **under**-reports.
Under-reporting weakens the assertion; over-reporting would raise false alarms
in a `debug_assert!` and get it deleted. That is the correct trade and the
comment says so explicitly.

I checked that the assertion does not fire on any existing call site: a full
debug-profile `cargo test` — where `debug_assert!` is live — passes 1389 lib
tests and 518 harness tests with zero failures.

The reasoning that any `Ref::L` in already-determined syntax is *necessarily* a
sibling reference is sound: the compiler only ever obtains a concrete `Ref::L`
from this binder's own `add`/`add_deferred`/`running_ref`, while references to
enclosing scopes arrive as deferred `ProtoVar`s resolved against the binding
context. So the predicate is not a heuristic — it is exact for the shapes it
models.

### Tests

Three layers, which is the right shape for a compiler defect:

1. `test_meta_argument_compiles_to_letrec` — asserts on the **compiled STG
   shape** via `meta_group_is_recursive`, so it fails at the point of the
   defect rather than downstream.
2. `test_non_recursive_binder_rejects_sibling_reference` — a `should_panic`
   test on the invariant itself, correctly `#[cfg(debug_assertions)]`-gated
   since the assertion does not exist in release builds.
3. The harness fixture — end-to-end, and the one that catches the silent-value
   case rather than only the blackhole.

`meta_group_is_recursive` walks `Let`/`LetRec`/`Ann`/`Seq`/`Case`/`DeMeta`/
`FusedPrimop`, which covers every node that can enclose the group. It returns
`Some(false)` on finding the group under a plain `Let`, so a regression is a
clear assertion failure rather than a `None` that could be mistaken for "not
found".

### Fixture renumbering

`192_frf2o_inline_annotated_block` → **`205_...`**. 192 is already taken by
`tests/harness/errors/192_1tkk_7_9_function_not_block.eu` in the shared
fixture sequence, and 203/204 went to #1074 and #1081. No `192_frf2o`
reference survives anywhere in the tree.

### Conflict resolution

One conflict, `tests/harness_test.rs`, at the seam with the fixture functions
that landed on master since this branch was cut. Resolved **by region**, each
function keeping its own braces, then rebuilt to verify — a
concatenate-both-sides resolution on a brace-structured file silently swallows
the trailing brace the two sides share.

### Gates

- `cargo fmt --all` — clean
- `cargo test` — **0 failed across all 18 test binaries**: lib 1389, harness
  518, doc examples 111, diagnostics invariants 11, wire-format enforcement 8

PR #1088, opened 2026-07-27
fix(vm): pass settled argument slots through instead of re-aliasing (eu-wpswc)

gmorpheme:
Fixes **eu-wpswc** (P1) — `xs map(f) sum`, and every `data map(f) reduce` pipeline, was quadratic in list length.

> **Measurement note.** Every load-bearing figure below is a **deterministic VM tick count**, identical on any machine. The machine was heavily contended tonight (load average ~19, and the disk filled at one point), so wall-clock figures appear only as illustration and are labelled **measured-single on a contended machine**. No release-grade or quiet-machine claim is made anywhere — that is eu-1hcw's and the owner's.

## The defect

Folding a lazily-*mapped* list was O(N²). Folding the same list directly was linear. Neither `map` alone nor the fold alone was affected — only the combination, because only a fold forces the mapped *elements*.

Ticks fit exactly (bytecode pre-decoded, blob prelude, master `a1af1f0b`):

```
xs map(sq) sum     ticks = 0.875·N² + 248.25·N − 6
xs map(sq) count   ticks = 218.75·N + 19            (linear)
```

Allocations were **exactly linear in both arms** (25.75·N + c, identical numbers). So this was **re-evaluation, not re-allocation**. `--suppress-demand-analysis` did not change the shape, ruling out a strictness divergence.

## Root cause

This is the eu-e3c3i mechanism — an unmemoised `Atom{Ref::L}` alias chain on a lazily-threaded parameter — reappearing at a call shape the eu-e3c3i fix (`6a902030`, `eager_args` at self-recursive call sites) cannot reach.

`map` is written with a **curried** self-call:

```
map(f, l): if(l nil?, l, cons(l head f, l tail map(f)))
```

so the recursive call compiles to a partial application plus an apply, not a saturated self-call (`eu dump runtime`, global 452):

```
[1] thunk let [0] thunk 452(*0);      # map(f) — PAP, arity 2, 1 arg
              [1] thunk 280(*1)       # tail(l)
          in *0(*1)                   # PAP applied
```

Compare `foldl`, whose recursive call is the saturated `DirectApp` `→437(*2, *0, *1)`.

Two alias links are minted per recursion level: `create_arg_array` wraps the `Ref::L` naming `f` at the partial application, and the PAP trampoline (`pap_syn` / `encode_pap_templates`) re-wraps the already-materialised supplied argument on every re-application. Those alias closures are built with `update() == false`, so the `Atom{Ref::L}` handler pushes **no `Update` continuation** for them and nothing ever collapses the chain. Resolving `f` at depth k walked 2k links; forcing every element cost Σ2k.

Confirmed at trace level (`EU_HEAPSYN=1 -d`, N = 10/20/30, per-signature second differences): total second difference **175 ⇒ a = 0.875 exactly**, carried entirely by two alias signatures whose env chain grows one frame per level. The raw trace shows the literal alternating walk terminating at the `sq` lambda:

```
(*2|[×1]→[×2]→[×2]→[×2]→[×2]→[×4])
(*1|[×1]→[×2]→[×2]→[×2]→[×4])
(*2|[×1]→[×2]→[×2]→[×2]→[×4])
...
(*1|[×2]→[×4])
λ{1}(… sq …)
```

**User-code discriminator, no prelude involved** — a hand-written `mymap` with a curried self-call reproduces the prelude numbers *tick for tick* (116,744 / 342,869 / 1,123,244 / 3,996,494 at N = 250/500/1000/2000); the same algorithm with a saturated self-call is exactly `255.24·N`.

## The fix

Three functions, all argument-array construction. An argument naming a **settled** slot — one whose closure can never change — is passed through to the callee by value instead of being wrapped in a fresh alias over the caller's frame.

Settled = not an update thunk, **and** either:

* **arity > 0** (a lambda or PAP), or
* **an `Atom` node** (a pure indirection — which is what an alias closure itself is).

Both clauses are load-bearing and are gated separately (see the tests). The first stops a chain forming on a literal lambda; the second stops one forming when the function parameter arrives as an alias, which is what happens when it is a *thunk* that evaluates to a function.

Everything else keeps its alias, so its representation is **bit-identical to before**.

### Why this is sharing-neutral (condition 3)

The alias exists for exactly one reason: memoisation. If a slot holds an update thunk, the callee must force it *through the caller's slot* so the WHNF is written back there and every reader shares the work. Update thunks therefore keep their alias, unchanged — **nothing that is memoised today stops being memoised**.

For the two settled shapes:

* **A lambda or PAP is never forced and never updated.** A callee holding one by value applies it exactly as it would through an alias. Applying the same PAP from two sites builds two independent saturated frames either way — the PAP closure is not the sharing point for its supplied arguments; those live in the PAP's own env frame, which both applications continue to share *by pointer*, so a supplied thunk is still memoised at exactly one slot. This is precisely the hazard I flagged for `app_eager` in the PAP trampoline, and the chosen fix does not inherit it: the trampoline is untouched, and a PAP's *contents* are never copied — only the PAP closure value (a code pointer plus an env pointer).
* **An `Atom` node is a pure indirection.** Passing it on rather than re-aliasing it changes nothing about where a force lands: an alias onto a thunk elsewhere still navigates to *that* owning frame and updates *that* slot.

**Black holes** need no explicit exclusion. A black hole is the one transient in which a non-updateable closure occupies a slot whose contents are about to change; it has `arity == 0` and is not an `Atom`, so it matches neither clause. `tests/harness/191` exercises this deliberately with `nats: cons(0, nats map(inc))` — a self-referential list whose mapped spine is built from a slot that is black-holed while its own head is produced.

### What I deliberately did **not** do, and why (an honest negative result)

My first implementation used the wider predicate "any non-update-thunk closure is settled". It is sound for *evaluation* — but it **breaks the IO-run driver**, which navigates argument structure statically (`driver::io_run::peel_meta` / `block_list_inner`, and the bytecode twin `BytecodeMachine::block_field_values`) on the assumption that it can chase an `Atom` indirection and record the container frame it came from. Handing it a bare `Cons` instead breaks that walk. Measured: **14 harness tests fail**, all `io.*`, on all three dispatch paths, with `IO spec block did not evaluate to a Block constructor`. Verified pre-existing-free by reverting: baseline passes.

Narrowing to the two clauses above fixes the defect completely while leaving every argument the navigator looks at bit-identical. The wider predicate is a real further opportunity, blocked on making that navigator independent of argument aliasing — recorded as follow-up on eu-wpswc, not attempted here.

Other rejected alternatives, with measurements rather than argument:

* **Rewrite `map`'s recursion in `lib/prelude.eu`** — fenced, and it fixes one victim rather than the defect: any user combinator with a curried self-call stays quadratic.
* **Flatten curried application spines in the STG compiler** — prototyped and measured. It only *halves* the quadratic coefficient (3,996,494 → 1,979,499 at N = 2000, still quadratic), because after the inliner unrolls the combinator the recursive call is to an anonymous lambda literal and self-recursion detection cannot fire.
* **`app_eager` in the PAP trampoline** — removes only one of the two links, and unconditional eager resolution of a supplied argument can lose memoisation when a PAP is applied twice.
* **Alias-chain path compression in the `Atom{Ref::L}` handler** — general, but strictly more invasive (mutating env slots mid-instruction in both engines) for no extra coverage.

## Results

Deterministic ticks, fitting **every measured point exactly** at N = 500/1000/2000/4000:

| dispatch | before | after |
|---|---|---|
| bytecode pre-decoded | `0.875·N² + 248.25·N − 6` | `230.75·N − 21` |
| bytecode byte (`EU_PREDECODE=0`) | `0.875·N² + 259.25·N − 10` | `241.75·N − 25` |
| HeapSyn (`EU_HEAPSYN=1`) | `0.875·N² + 271.25·N − 5` | `248.75·N − 17` |

The quadratic coefficient is **exactly 0.875 on all three paths**, and all three are exactly linear after.

**Condition 5 — the named victims.** `max-map`/`min-map` are `l map(f) max-of`, which pp-fix measured as still quadratic after the eu-pazes prelude rewrite (which is why `tests/harness/195_pazes_max_of_linear.eu` is restricted to 3-element lists). They are cured:

| | before | after |
|---|---|---|
| `xs max-map(sq)` (bc-pd) | `0.5·N² + 356.5·N − 166` | `325·N − 166` |
| `xs reverse map(sq) max-of` (bc-pd) | quadratic (153,060 / 415,685 / 1,269,060 / 4,288,310) | `361.75·N − 185` |

Illustrative wall time, **measured-single on a contended machine**: `xs map(sq) sum` at N = 20000 went 118,793 ms → 141 ms; `xs max-map(sq)` at N = 20000 went 11,625 ms → 199 ms.

### Canonical bench suite 015–022

Ticks, one binary and one blob per arm, **measured-single**; ticks are deterministic and reproduced exactly on a repeat run.

| bench | bytecode before | bytecode after | HeapSyn before | HeapSyn after |
|---|---|---|---|---|
| 015_block_merge | 82,296,413 | 4,031,897 | 82,541,872 | 4,227,329 |
| 016_import_export_yaml | 31,460,913 | 7,018,463 | 31,967,231 | 7,510,864 |
| 017_import_export_toml | 31,287,589 | 6,845,139 | 31,793,907 | 7,337,540 |
| 018_string_scale | 45,747,189 | 2,730,395 | 45,995,706 | 2,943,907 |
| 019_list_scale | 33,006,461 | 1,390,917 | 33,156,485 | 1,510,936 |
| 020_lookup_curve | 164,202,155 | 23,865,943 | 1,443,585 | 1,221,888 |
| 021_io_loop | 2,941,275 | 1,832,740 | 2,962,309 | 1,831,763 |
| 022_hof_fold | 76,500,403 | 70,500,358 | 84,700,417 | 77,100,367 |

All eight PASS in both engines, before and after. **I am deliberately not headlining these**: they are a side effect of the defect having been pervasive across pipelines that thread a function through a recursion, they are ticks not wall time, and the engine-A/B story belongs to eu-1hcw and the owner. The `020_lookup_curve` bytecode/HeapSyn asymmetry is pre-existing on both arms and is not touched by this change.

## GC and reachability (condition 2)

**What changes in the object graph.** Before, the callee's argument array held a freshly allocated one-word `Atom{Ref::L(i)}` closure whose env pointer was the **caller's frame**; reaching the actual argument took two hops (array → alias → caller frame → slot i). After, for a settled slot the array holds the target closure directly: one hop, and the caller's frame is no longer referenced *by this edge at all*.

So the change **strictly reduces retention**, in two ways:

1. **One fewer heap object per settled argument per call** — the alias closure is not allocated. Nothing else can reference it (it is created and immediately stored), so this is a pure reduction in the live set.
2. **One fewer edge into the caller's frame.** Previously a callee that outlived its caller kept the caller's whole env frame — and transitively everything that frame reached — alive purely so the alias could resolve `Ref::L(i)`. Now it retains only the argument it actually uses. For the defect's own shape this is the difference between retaining N nested frames and retaining one lambda.

No new edges are created. The target closure was already reachable from the caller's frame at the moment of the copy, and a `Closing` is a two-word value (code pointer + env pointer) copied by value into a GC-managed `Array`, exactly as `create_arg_array_eager` has done since eu-e3c3i — so the scan set gains no new *shape*, only fewer members. Nothing is mutated in place, so no write barrier is involved that was not already involved: the array is populated before the frame that holds it is allocated.

The one reachability *risk* the change could pose is the black-hole transient, and the predicate excludes it by construction (see above).

**Empirically**, `EU_GC_VERIFY=2 EU_GC_POISON=1 EU_GC_STRESS=1` over the full 176-file harness corpus, on each dispatch path, run after the disk was reclaimed:

```
[pd]   ran=176 gc_panics=0
[byte] ran=176 gc_panics=0
[hs]   ran=176 gc_panics=0
```

plus the four fold-over-map fixtures at N = 4000 under the same three flags on all three paths, all producing identical correct results (`mapsum`, `thunkfn`, `maxmap`, `revmapmax` — 12 runs, 12 clean exits).

## Engine agreement (condition 1)

The three builders implement the identical predicate, and each carries a doc comment saying they must not drift; `machine::env_builder`'s module doc is the single normative statement and both bytecode twins point at it.

**Output byte-identity across `pd` / `EU_PREDECODE=0` / `EU_HEAPSYN=1`** over the whole corpus (355 files, `tests/harness/*.eu` + `tests/harness/errors/*.eu`, stdout+stderr+exit code): **353 identical, 2 explained**:

* `070_bench_validation` prints VM tick counts, which differ per dispatch path *by design* — that is what the bench harness measures.
* `171_yaml_large_uint` differs only in the OS thread id inside a panic message.

(The comparison normalises the `Report generated at …/<uuid>/report.html` line, which is a fresh UUID per run and differs between two runs of the *same* binary.)

**Tick growth converges on all three paths** — each is exactly linear after the fix and each was exactly `0.875·N²` before, as tabulated above.

## Relationship to `eager_args` (condition 4) — recorded, not acted on

`eager_args` is **partially subsumed and still doing independent work**. Not removed in this PR.

* **Subsumed:** for a settled argument at a self-recursive call site the two now do the same thing, and the settled-slot path gets there without needing the compiler to have identified self-recursion — which is exactly why it reaches `map`'s curried call, where `eager_args` cannot fire.
* **Still independent:** `create_arg_array_eager` resolves *every* `Ref::L` from the env, **including update thunks**. That is strictly more aggressive and not sharing-neutral — it copies a thunk closure by value, so the callee's force updates the callee's slot and the caller's slot keeps the thunk, and the work can be done twice. The settled-slot predicate deliberately never does that. So `eager_args` still changes behaviour for thunk arguments at self-recursive sites, and removing it would be a separate, measurable change with its own risk.

My recommendation, for a separate bead: measure whether `eager_args`'s thunk-copying behaviour buys anything now that the alias chains are gone, and if it does not, retire the flag in favour of the sharing-neutral predicate. Not in scope here.

## Related beads

* **eu-pghae** (P3, "interposed map defeats PR #1016 HOF-fold fusion") — **the same defect**, diagnosed as a fusion gap. The mechanism is not fusion: allocations were exactly linear, so nothing extra was being built; the cost was re-walking an alias chain. `range(N) map(f) foldl(+, 0)` is exactly linear on all three paths after this change. Recommend linking to eu-wpswc and closing with it.
* **eu-9wq0s** — **not** the same. That is a blob-vs-source *divergence*; this defect reproduces identically on both paths (source-prelude `map(sq) sum` at N = 20000 measured ~9.0 s before, matching blob), and the full suite is green under `EU_SOURCE_PRELUDE=1` after.
* **eu-gua64** (quill) — **independent**. Different complexity class (2^n vs N²); allocations exactly linear here where beta-duplication would multiply them; disjoint code paths — this is fixed entirely inside the VM's argument-array builders with `src/core/inline/` untouched. quill's branch is still not on `origin`, so the re-run I promised is outstanding; it does not gate this.
* **eu-14ya9** — not applicable: that hole is intrinsic-only and no intrinsic is involved here.

## Tests

**`tests/fold_over_map_growth_test.rs`** — the complexity-class gate. Compares deterministic **tick growth** across a 4× size increase (linear ≈ 4.0, the defect ≈ 11.0, cap 6.0). Ticks rather than a wall-clock deadline so it cannot flake on a loaded runner. The spawned `eu` inherits `EU_HEAPSYN`/`EU_PREDECODE` from the test process — the same arrangement `tick_parity_test.rs` uses — so all three CI engine configurations gate it independently. Two arms, one per clause of the predicate.

**`tests/harness/191_wpswc_settled_slot_passthrough.eu`** — the semantics pass-by-value must not disturb: results across `map`/`max-map`/`min-map`/nested maps, laziness over self-referential mapped structures (`ones`, `nats: cons(0, nats map(inc))`), and a partial application reused from two sites. 16 checks, `RESULT` computed from all of them.

### Fault injection — done, per site, with a rebuild each time

| injection | result |
|---|---|
| `machine::env_builder::is_settled` → `false` | `EU_HEAPSYN=1` **FAIL** (ratios 10.76 / 10.55); default and `EU_PREDECODE=0` PASS |
| `bc_is_settled` (byte dispatch) → `false` | `EU_PREDECODE=0` **FAIL** (10.89 / 10.68); default and HeapSyn PASS |
| `bc_is_settled_pd` (pre-decoded) → `false` | default **FAIL** (11.02 / 10.84); `EU_PREDECODE=0` and HeapSyn PASS |
| drop the `Atom` clause from all three | thunk arm **FAIL** on all three paths; lambda arm PASSES on all three |
| all restored | PASS everywhere |

Each site fails **only** its own dispatch path, which is what makes the "must not drift" claim checkable rather than aspirational. The fourth injection gates the second clause of the predicate independently of the first.

Harness 191's verdict was separately confirmed to gate (breaking one check gives `FAIL`, rc=1; restoring gives `PASS`, rc=0).

**One honest gap:** harness 191 does **not** catch the over-broad-predicate variant — its 16 checks all pass under it. What catches that is the existing `io.*` corpus (14 tests), which is how I found it. Recorded here rather than papered over.

## Validation

Full `cargo test --no-fail-fast`, four configurations:

| configuration | result |
|---|---|
| default (bytecode pre-decoded) | 1358 lib + 506 harness + 111 typecheck + all others pass; **3 pre-existing failures** |
| `EU_PREDECODE=0` | same 3 pre-existing failures, nothing else |
| `EU_HEAPSYN=1` | same 3 pre-existing failures, nothing else |
| `EU_SOURCE_PRELUDE=1` | harness **507/507**, diagnostics_invariants **3/3**; 2 `blob_mode_*` failures, pre-existing |

The 3 pre-existing failures are `secondary_labels_do_not_excerpt_library_internals`, `debug_trace_restores_the_uncurated_trace` and `test_193_1tkk_7_12_curated_trace` — the eu-7x0r blob-mode diagnostics trio that **PR #1079 fixes**. I verified they fail identically with this change reverted, on the same binary. The 2 source-prelude failures are `diagnostics_blame_plumbing_test`'s `blob_mode_*` cases, which assert blob-mode behaviour and cannot pass when `EU_SOURCE_PRELUDE=1` forces the source path; also verified identical on baseline.

`errors/` corpus unchanged in both prelude modes and all three dispatch paths.

`cargo fmt --all` clean. `cargo clippy --workspace --all-targets -- -D warnings` clean, no `#[allow]` added.

### Side finding for the coordinator (asked for, not fixed here)

Those same two `diagnostics_invariants` tests **also fail under `EU_HEAPSYN=1` on unmodified master** — a second axis on the same assertions. I checked PR #1079's branch (`fix/clarion-eu-7x0r-blob-blame`, `d33218b3`) directly under `EU_HEAPSYN=1`, with its own regenerated blob: `diagnostics_invariants` **3 passed, 0 failed** and `harness_test` **506 passed, 0 failed**. So it is **one defect visible along two axes**, and #1079 closes both. No separate HeapSyn diagnostics bead is needed. I did not touch `tests/diagnostics/`.

## Fences

This change touches **none** of the fenced areas: nothing in `src/core/inline/`, `lib/prelude.eu`, `src/eval/stg/meta.rs`, `src/eval/stg/compiler.rs`, `wire_format.rs`/`blob.rs`/`build.rs`/`xtask`, `src/driver/check.rs`, `tests/diagnostics/`, or `src/eval/error.rs`/`src/common/sourcemap.rs` trace curation. **No wire-format change** — `BYTECODE_WIRE_FORMAT_VERSION` is untouched and the blob is byte-identical (598,731 bytes, sha256 `c2ff6873a510a91f…`) before and after.

## Review

**Do not merge — owner review.** This touches the VM and memory behaviour, so per `CLAUDE.md` it also needs a recorded review comment from someone other than the author before any merge. I am not merging it and I have not closed the bead.


gmorpheme:
COMPOSITION TEST vs eu-gua64 / PR #1086 — CLEAN. No merge-order finding.

Method: threw away branch off #1088 (91e03b95), merged origin
fix/quill-eu-gua64-beta-reduce-sharing at 2319d7af. Only CHANGELOG conflicted
(both add an Unreleased entry); src/ and tests/harness_test.rs auto-merged with
no code conflict. Regenerated lib/prelude.blob and rebuilt for every arm --
#1086 touches src/core/inline/, which changes blob content, so a stale blob
would have invalidated the comparison. All figures are deterministic VM ticks.

Full 2x2 factorial, bc-pd (both other dispatch paths agree; see below).

  wpswc fixture: range(1,N) map(sq) sum, N=4000
    master              14,992,994    0.875N^2 + 248.25N - 6
    #1086 alone         14,992,994    IDENTICAL TO MASTER, tick for tick
    #1088 alone            922,979    230.75N - 21
    combined               922,979    IDENTICAL TO #1088 ALONE, tick for tick

  gua64 fixture: mymax-of over range(1,N), N=22
    #1088 alone         74,451,652    still exponential (1,165,264 / 4,655,308 /
                                      18,614,728 / 74,451,652 at N=16/18/20/22,
                                      x4 per +2)
    #1086 alone              4,085    187N - 29
    combined                 3,806    175N - 44

So the two fixes are strictly orthogonal in both directions: quill's has ZERO
effect on the wpswc quadratic (identical to master to the tick), mine has zero
effect on the gua64 exponential, and each defect needs its own fix. Nothing is
subsumed and there is no merge-order dependency.

Where they interact at all, they compose ADDITIVELY rather than antagonistically:
the combination runs the gua64 fixture ~6.4% cheaper than #1086 alone (175N vs
187N), because once the exponential is gone my change also removes the alias
work from the remaining linear traversal. Nothing gets worse in either direction.

All three dispatch paths under the combination:
  mapsum   bc-pd 115,354 / 230,729 / 461,479 / 922,979 at N=500/1000/2000/4000
           bc-byte 120,850 / 241,725 / 483,475 / 966,975
           hs 124,358 / 248,733 / 497,483 / 994,983
  thunkfn  bc-pd 120,847 / 241,722 / 483,472 / 966,972 (and byte/hs likewise)
  gua64    bc-pd 2,756 / 3,106 / 3,456 / 3,806 / 4,156 at N=16/18/20/22/24
           bc-byte 2,889 / 3,257 / 3,625 / 3,993 / 4,361
           hs 2,988 / 3,368 / 3,748 / 4,128 / 4,508
Every one of those is exactly the #1088-alone figure for the wpswc fixtures, and
exactly linear for the gua64 fixture.

Values identical across all three dispatch paths under the combination, and
identical to the #1088-alone values: mapsum and thunkfn 21325334000, maxmap and
revmapmax 15992001, minmap 1, gua64 23.

tests/fold_over_map_growth_test.rs passes on all three dispatch paths under the
combination (2/2 each).

Full cargo test --no-fail-fast on the combination: default engine 1361 lib + 507
harness + 111 typecheck; EU_HEAPSYN=1 the same. The only failures are the same
three pre-existing eu-7x0r blob-mode diagnostics
(secondary_labels_do_not_excerpt_library_internals,
debug_trace_restores_the_uncurated_trace, test_193_1tkk_7_12_curated_trace),
which PR #1079 fixes. Nothing new appears in the combination that is absent from
either PR alone -- in particular no repeat of the diagnostics stack's
"neither fix alone sufficed" pattern.

Throwaway branches deleted; nothing pushed. Worktree back on
fix/furnace-eu-wpswc-fold-over-map at 91e03b95, clean, blob regenerated and
byte-identical to the PR (598,731 bytes, sha256 c2ff6873a510a91f...), headline
figures re-verified after restore.

gmorpheme:
## Recorded review (non-author, CLAUDE.md — GC/memory surface) — eu-wpswc

Reviewed by Wicket at `origin/fix/furnace-eu-wpswc-fold-over-map`, release build. **Recommend merge.**

This is an unusually well-evidenced PR: deterministic ticks rather than wall clock, a user-code discriminator that reproduces the prelude numbers tick for tick, rejected alternatives with measurements attached rather than assertions, and an honest negative result. The measurement discipline is right and the "measured-single on a contended machine" labelling of the wall figures is the correct call.

### The PAP hazard — probed empirically, and it holds

The safety argument turns on one claim: a PAP passed by value rather than through an alias cannot lose memoisation of its *supplied* arguments. The author flags this themselves as what killed the `app_eager` alternative, so I tested it rather than reading it.

Built one PAP with an expensive supplied argument and saturated it twice, against a control that builds two independent PAPs:

```eu,notest
sum-to(n): range(0, n) sum
add(a, b): a + b
p: add(sum-to(3000))
main: [p(1), p(2)]              # shared
# control: [add(sum-to(3000))(1), add(sum-to(3000))(2)]
```

Deterministic ticks:

| | shared | separate (control) |
|---|---|---|
| **with #1088** | 371,625 | 742,881 |
| without #1088 (master + #1087) | 374,660 | 748,921 |

The shared case sits at almost exactly half the control in both arms. If passing the PAP by value had duplicated its supplied argument, the shared figure would have risen toward the control's ~742k; it did not move. **Memoisation of PAP-supplied arguments is preserved**, which is the crux of condition 3. (#1088 is also marginally cheaper, ~0.8%, but that is noise-adjacent and I am not claiming it.)

### The predicate

`is_settled` is `!update() && (arity() > 0 || code is Atom)`. Both clauses are genuinely load-bearing and the reasoning is sound: update thunks keep their alias, so **nothing that is memoised today stops being memoised** — that is the property that makes this safe, and it is stated as the invariant rather than left implicit. Black holes need no special case (arity 0, not an `Atom`), and the author is right that this falls out rather than being arranged.

I verified the predicate is implemented identically at all three sites: `src/eval/machine/env_builder.rs:87`, `src/eval/bytecode/machine.rs:1725` and `:1741`. Same three tests, same order, same outcomes.

### Finding: nothing mechanically enforces the three-site invariant

The module doc says "the three must not drift". That is a comment asking future authors to keep three copies in step — **structurally the same mechanism that failed in eu-3skeg**, where a "MUST match" comment was the only thing keeping `BYTECODE_WIRE_FORMAT_VERSION` synchronised between `build.rs` and `xtask`, and #1078/#1079 collided on it within a day. #1082 is in this same queue precisely to delete that pattern.

The mitigation here is real and better than eu-3skeg's: the three paths are separately exercised in CI (bytecode pre-decoded, `EU_PREDECODE=0`, `EU_HEAPSYN=1`), so a drift should surface behaviourally rather than silently. That is why I am not blocking on it. But a shared predicate — or at minimum a test that asserts the three agree on a table of closure shapes — would be cheap and would close it properly. Worth a follow-up bead rather than a change here.

### The negative result is the right call

Discovering that the wider "any non-update-thunk is settled" predicate breaks the IO-run driver's static navigator — 14 `io.*` harness failures on all three dispatch paths, verified pre-existing-free by reverting — and then *narrowing* rather than shipping the wider predicate with a workaround, is exactly right. Recording it as follow-up rather than attempting it here is also right.

### What I did NOT verify

- **I did not reproduce the quadratic-to-linear result or re-derive any of the fitted coefficients.** The `0.875·N²` fits, the bench-suite table, and the `max-map`/`min-map` figures are all the author's numbers. I checked them for internal consistency and read the methodology, but this review is not independent confirmation of them, and given this project's history none of those figures should enter a release note without a quiet-machine re-measurement under `docs/superpowers/engine-ab/PROTOCOL.md`.
- I did not audit the GC/reachability section (condition 2) beyond reading it; my PAP probe covers sharing, not the object-graph claims.
- I did not review `tests/harness/191_wpswc_settled_slot_passthrough.eu` for gating quality, including the `nats: cons(0, nats map(inc))` black-hole case — I read the description only. Note it will need renumbering off 191 when it lands; that slot is contested and master is now past it.
- I relied on the PR's CI for the full matrix, clippy and fmt.

gmorpheme:
**Owner approval: approved for merge.**

Reviewed personally by the owner under the `wicket.md` owner-review filter (GC/memory layout change), and released for merge on the recorded review at https://github.com/curvelogic/eucalypt/pull/1088#issuecomment-5091333923.

The load-bearing evidence for the record: the PAP probe measured 371,625 ticks for the shared case against 742,881 for the two-PAP control, with the 2:1 ratio preserved both with and without the change — so PAP-supplied argument memoisation is intact. That was the crux of the safety argument (and the hazard that ruled out the `app_eager` alternative), measured in deterministic ticks rather than argued.

Also confirmed at review: `is_settled` is consistent across all three sites — `src/eval/machine/env_builder.rs`, and `make_arg_array` / `make_arg_array_pd` in `src/eval/bytecode/machine.rs`. Nothing enforces that consistency automatically, which is the same mechanism that failed in eu-3skeg; the mitigation is that the three dispatch paths are separately exercised in CI (bytecode-pd, `EU_PREDECODE=0`, `EU_HEAPSYN=1`), so a future drift shows up behaviourally rather than silently.

gmorpheme:
## Merge-time review (Wicket) — updated onto master `7f7b5e45`

Merging on the owner's approval (#issuecomment-5095359183). This comment records what changed at merge and one finding that corrects a claim in the PR body.

### Conflict resolution

Two conflicts, both additive, resolved by region: `CHANGELOG.md` (both Unreleased *Fixed* entries kept) and `tests/harness_test.rs` (both test blocks kept — note the two sides shared a trailing `}`, so a naive keep-both nests one test inside another; the seam was repaired by hand).

**Fixture renumbered `191` → `221`.** `191` is taken in the shared `tests/harness` + `tests/harness/errors` sequence by `errors/191_m93j_lookup_on_function.eu`, and master is long past it. The test function was renamed to match — a test name disagreeing with its fixture number is how the next renumber goes wrong.

### FINDING — the two clauses are not separately gated, contrary to the PR body

The PR states: *"Both clauses are load-bearing and are gated separately (see the tests)."* The CHANGELOG said *"one arm per predicate clause"*. I fault-injected rather than accepted this, on all three dispatch paths:

| injection | lambda arm | thunk arm |
|---|---|---|
| `Atom` clause off, `arity > 0` kept | **passes** | **FAILS** (ratio 10.88 vs cap 6) |
| `arity > 0` clause off, `Atom` kept | passes | passes — **on all three paths** |
| whole predicate off | **FAILS** (11.02) | **FAILS** (10.84) |

So the `Atom` clause is load-bearing and the thunk arm gates it independently, exactly as claimed. The `arity > 0` clause is **not** load-bearing for the complexity class, and nothing gates it: a lambda that is re-aliased once becomes an `Atom` node, which the *next* level then passes through, so the chain still cannot grow beyond depth 1. The lambda arm gates the pass-through as a whole (it fails when the predicate is disabled entirely), so neither test is vacuous — but it does not protect the clause the PR says it protects.

I measured what the clause is actually worth rather than just calling it redundant: **−1,599 ticks on `growth-large` and −3,198 on `growth-thunk-large`**, identical on the byte and HeapSyn paths (~0.4–0.8%, scaling linearly). That is a real constant-factor saving, and it is safe — the owner's PAP probe covers precisely this shape — so **the code is unchanged**. What changed is the documentation, so that a green suite is not misread as evidence the clause survives. This is the fourth claim in this queue written while the mechanism underneath it was still moving.

### `tests/gate_liveness.rs` — the #1095 meta-gate fired on contact

`tests/fold_over_map_growth_test.rs` is `cfg(not(target_arch = "wasm32"))` and was unregistered, so the merge failed `every_conditionally_compiled_test_file_is_registered`. Registered with `Supplier::NativeTargetsOnly`, matching its sibling `tick_parity_test.rs`. That gate has now caught a genuine omission on every PR it has touched.

### What I verified rather than accepted

- **The added `unsafe`.** `create_arg_array` now holds `&EnvFrame` from `unsafe { environment.as_ref() }` across `array.push`, which allocates. That is only sound if a moving collection cannot run mid-instruction — `src/eval/machine/vm.rs:2329` documents exactly that ("GC can only run between ticks … no allocation path reachable from here triggers an inline collection"), and the pre-existing sibling `create_eager_arg_array` relies on the same invariant with the same idiom. Sound, and consistent with the surrounding code.
- **Error paths unchanged.** `None if eager` still returns `BadEnvironmentIndex`; a lazy arg naming an out-of-range slot still falls through to the alias closure exactly as before.
- **Fixture 221 genuinely gates** — falsifying check `t9` fails `cargo test`. Sixteen checks, all feeding `RESULT`.
- **Full `cargo test` green on all three dispatch paths.** Two `diagnostics_snapshots` failures appear under `EU_HEAPSYN=1` and `EU_PREDECODE=0` — **pre-existing, not this PR**: clean master at `7f7b5e45` fails them identically, and the diverging fixture sets are byte-for-byte the same with and without #1088 (7 fixtures under HeapSyn, 6 under byte dispatch). Reported separately; see below.
- **`EU_GC_VERIFY=2`** on both growth targets on all three paths, and `EU_GC_VERIFY=2 EU_GC_POISON=1` on fixture 221 on both engines: exit 0, no verifier panic.
- No `Smid::default()` in new code.

### Separate pre-existing finding, for the owner — not a blocker here

The #1085 golden diagnostic snapshots are **engine-sensitive and validated on the default engine only**. On clean master they fail under both `EU_HEAPSYN=1` (7 fixtures) and `EU_PREDECODE=0` (6), all `--source-prelude`, all losing a single `stack trace:` frame. CI cannot see this: `tests/diagnostics_snapshots.rs` is `cfg(prelude_blob_ok)`, only the `diagnostics-blob-mode` job supplies a blob, and that job runs the default engine — so under the HeapSyn and byte-dispatch jobs the file compiles to zero tests and reports `ok. 0 passed`. This is the same shape as the three "a green signal that was not measuring what everyone assumed" findings already recorded this release.

PR #1089, opened 2026-07-27
fix(export): YAML/TOML u64 overflow aborts the process instead of diagnosing (eu-1tkk.7.20)

gmorpheme:
Fixes **eu-1tkk.7.20**. Do not merge — owner review.

## The panic

```
$ eu 171_yaml_large_uint.eu          # data.n == 9999999999999999999
thread '<unnamed>' (22509971) panicked at src/export/yaml.rs:46:29:
number 9999999999999999999 is too large to represent as a YAML integer (max i64)
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

thread 'main' (22509944) panicked at src/bin/eu.rs:43:10:
main thread panicked: Any { .. }
```

Exit 101, no location, no code, no trace, and a thread id that changes
every run. Now:

```
error[EU-RENDER-UNREPRESENTABLE]: cannot represent this value in YAML output: the integer 9999999999999999999 is above 9223372036854775807, the largest integer a YAML integer scalar can carry
  ┌─ 171_yaml_large_uint.eu:4:7
  │
4 │ main: data.n
  │       ^^^^
  │
  = render to a format that can carry the value — 'json', 'text' and 'eu' output all keep integers of this magnitude
  = to keep the exact digits in this format, convert the value to a string first with 'str', e.g. 'n str'
  = stack trace:
    - main at 171_yaml_large_uint.eu:4:7
```

Exit 1, and stdout is now empty (it wasn't at first — see below).

## TOML had the identical bug

The audit turned up that the same input hit `unreachable!` at
`src/export/toml.rs:32` and aborted the same way — TOML integers are
signed 64-bit too. That is a one-liner of the identical shape, so it is
fixed here rather than deferred, with its own fixture (194).

## Design

`Emitter` gains two defaulted methods:

- `format_name()` — "YAML", "TOML"
- `unrepresentable(&Primitive) -> Option<String>` — default `None`

The emit intrinsics ask before emitting any scalar, so the failure becomes
`ExecutionError::UnrepresentableValue(Smid, format, reason)` carrying
`machine.annotation()`, and flows through the ordinary diagnostic path
(primary label, notes, curated trace). `OwnedCaptureEmitter` forwards both
methods, so `render-as :yaml` is covered too — without that the guard
would have had a hole.

New stable code **`EU-RENDER-UNREPRESENTABLE`**, catalogued in
`docs/reference/error-codes.md` and in `eu error <CODE>` per the
"To add a new code" note in that file.

### Why error rather than coerce

The options were error, emit as a string, or emit unquoted. **Error**, for
the eu-odkp reason: silently changing a number's type or magnitude on
export means the document claims something the user never evaluated. No
other exporter has a convention for unrepresentable values to follow —
JSON, text and `eu` all carry the value fine, so YAML and TOML are the
only formats with a limit here.

The now-unreachable conversion branches in `from_primitive` no longer
abort either: each is driven by the same predicate used by
`unrepresentable`, so the two cannot disagree, and the fallback keeps the
exact digits (`Yaml::Real` is emitted unquoted and unaltered).

### Partial-document flush

Removing the abort exposed a second defect: the driver calls
`stream_end()` on the error path, so `TableAccumulator` flushed a
half-built document and `eu -x yaml big.eu > out.yaml` wrote `--- {}`
into `out.yaml` alongside the error. The accumulator now tracks
`OutputDocumentEnd` and discards a document abandoned mid-render.

## Sidecar

`171_yaml_large_uint.eu.expect` asserted only `exit: 101` and
`"too large to represent as a YAML integer"` — both of which the panic
satisfied, which is why the harness was green on it for its whole life.
It now asserts the code, the message, and the **primary label's
file:line:col**, anchored on the `┌─` line:

```
┌─[^\n]*171_yaml_large_uint\.eu:4:7
```

`[^\n]*` cannot cross a line, so a diagnostic whose location survives only
in the `stack trace:` note does not match. Verified directly: stripping the
`┌─` line from the real stderr and re-running the pattern gives no match,
while the unmodified stderr matches.

This is the fixture-local part only; **eu-ziq8o** still owns the
corpus-wide hole.

## Fault injection

Every injection rebuilt; `df -h /System/Volumes/Data` checked before
starting (58 GiB free) and again mid-run (41 GiB).

| # | Injection | Result |
|---|---|---|
| A | `yaml_integer_overflow` → always `None` | `test_error_171` **FAIL** → restore → **PASS** |
| B | `toml_integer_overflow` → always `None` | `test_error_194` **FAIL** (both engines) → restore → **PASS** (both) |
| C | `machine.annotation()` → `Smid::default()` | **PASS** — see below |
| D | blank line inserted, shifting `main:` to line 5 | `test_error_171` **FAIL** → restore → **PASS** |

**Injection C did not gate, and I am reporting it rather than papering
over it.** With the Smid stripped the output is byte-identical, because
`to_diagnostic` falls back to searching the traces and recovers the same
location. So this fixture gates *that a correct primary label is present*
(A, B, D) but does not independently gate the `machine.annotation()`
plumbing — the two sources agree here. Worth knowing for eu-ziq8o.

## Validation

| prelude | engine | result |
|---|---|---|
| source | bytecode | all green (507 harness, 1353 unit) |
| blob | HeapSyn | same 3 |

The three are `diagnostics_invariants::secondary_labels_do_not_excerpt_library_internals`,
`diagnostics_invariants::debug_trace_restores_the_uncurated_trace`, and
`test_193_1tkk_7_12_curated_trace` — exactly as briefed.

Also clean under `EU_GC_VERIFY=2 EU_GC_STRESS=1` (harness suite; only the
known blob-mode 193 failure). `cargo fmt --all --check` and
`cargo clippy --workspace --all-targets -- -D warnings` both clean, no
`#[allow]`.

Fixture 194 initially used `main: { n: data.n }` and failed under HeapSyn,
which drops the primary label entirely for a render error inside a block
literal while bytecode keeps it. The fixture now uses the top-level scalar
shape that both engines annotate identically; the engine gap is filed as
**eu-1tkk.7.26**.

## Audit of the rest of export/import — not fixed here

Filed, not taken on:

- **eu-1tkk.7.23** (P1) — **EDN export silently corrupts** integers above
  `i64::MAX`: `9999999999999999999` exports as `10000000000000000000`,
  rounded through `as_f64`. Not a panic, so no abort, but a data-fidelity
  bug of the eu-odkp class. EDN integers are arbitrary precision, so this
  should be fixed losslessly rather than made an error.
- **eu-1tkk.7.24** (P1) — **`eu -x html` panics on any ordinary block
  document**. `printf 'main: { a: 1 b: 2 }' | eu -x html` aborts at
  `markup.rs:170`. The markup emitter only accepts hiccup-shaped input.
  No fixture covers `-x html` failure at all.
- **eu-1tkk.7.25** (P1) — **`eu … | head` panics on broken pipe**.
  Confirmed for yaml, json, text, eu and html; toml and edn carry the same
  `.expect` on their write. Deliberately not fixed here: `Emitter::emit`
  returns `()`, so this needs either a fallible-emit refactor across every
  emitter and call site, or SIGPIPE handling in `main()` — a design
  decision, not a one-liner.
- **eu-1tkk.7.26** (P2) — HeapSyn primary-label gap described above.
- **eu-1tkk.7.28** (P2) — **TOML renders `null` as `""`**, so
  `{ a: null b: "" }` exports as two identical empty strings. Same
  fidelity family, and `Emitter::unrepresentable` added here is the
  mechanism to fix it — but unlike the integer case, which only ever
  panicked, this would change working behaviour, so it needs a decision
  rather than a drive-by change.
- **eu-1tkk.7.27** (P2) — import-side invariant panics enumerated
  (`import/yaml.rs:41,112,145,366`, `import/xml.rs:183,196`,
  `import/error.rs:80`). None reachable from ordinary input; csv, edn,
  toml, text, git and stream have no production panic sites.

Not touched: any fenced file.


gmorpheme:

`6b8b6b10`, with `lib/prelude.blob` regenerated after the branch switch and

### The best evidence that this is a real fix is in the diff of an old sidecar

`tests/harness/errors/171_yaml_large_uint.eu.expect` on master read:

```
exit: 101
stderr: "too large to represent as a YAML integer"
```

That fixture **pinned the panic as expected behaviour**. `exit: 101` is a Rust
abort, and a test asserting it is a test asserting that the process dies. This
PR replaces it with an assertion on an ordinary diagnostic — code, message,
primary label — and `exit: 1`.

Turning a codified panic into a codified diagnostic is exactly the direction
`CLAUDE.md`'s "Panics Are Critical" section calls for, and it is worth saying
out loud that the *old* test was the bug's accomplice.

### The design

The load-bearing decision is that each format gets **one** overflow predicate
(`yaml_integer_overflow`, `toml_integer_overflow`) used by both the rejection
path and the serialiser fallback, so the two cannot disagree. That is the right
shape: the alternative — a check at the emit intrinsic and an independent guard
in `from_primitive` — is precisely how a "can't happen" branch drifts out of
sync with the check that was supposed to make it unreachable.

The `Emitter` trait grows two defaulted methods, `format_name` and
`unrepresentable`. Defaulting them to `"output"` and `None` means every other
emitter is untouched and no format gains a restriction it did not have — the
change is opt-in per format, which is the correct blast radius.

`emit_scalar` in `src/eval/stg/emit.rs` is the single chokepoint, and I checked
that **every** scalar-emitting path now routes through it: `emit_set`,
`emit_vec`, `emit_ndarray_data` at rank 0 and rank 1, and the `Sym`/`Str`
arms of `EmitNative`. Each of those functions had to change signature to
`Result<(), ExecutionError>` and each propagates. No path was left behind — a
missed one would be a silent hole, since the fallback would then quietly
stringify instead of raising.

The error carries `machine.annotation()`, which is what gives the diagnostic
the source location of the value being rendered rather than a bare message —
and both sidecars anchor on the `┌─` **primary-label** line
(`…:4:7`), not merely on the message text, so a regression that keeps the
wording but loses the location fails the test. That is the right assertion and
the same standard the eu-8a49h and eu-og3u6 fixtures set earlier in this queue.

### The fallbacks

Both serialisers keep a fallback for the value the emit path now rejects,
documented as unreachable. I went back and forth on whether an unreachable
branch is worth having at all, and I think it is, for one specific reason: the
branch it replaces was `unreachable!()`, i.e. a panic. Replacing a panic with a
lossless representation is strictly better even if the branch is never taken,
because "never taken" is a claim about today's call graph.

The choices are sound. YAML uses `Yaml::Real(n.to_string())`, which is emitted
unquoted and unaltered, so the exact digits survive. TOML uses
`Value::String(n.to_string())`, which changes the type but preserves the digits
— TOML has no unquoted-scalar escape hatch, so that is the best available. Both
say what they do and why.

The `serde_json::Number` reasoning holds: `NegInt` is bounded below by
`i64::MIN`, so there is no underflow case to handle, and `as_u64().is_some() &&
!is_i64()` is exactly "above `i64::MAX`".

### `table.rs` — the quiet fix

`TableAccumulator` now tracks `document_open` and discards a document still
open at `OutputStreamEnd`. Without it, the driver's stream-end after a failed
render flushes a half-built (typically empty) document to stdout **alongside**
the error.

This is the part a reader could skim past, and it is what makes the fix
user-visible rather than merely internal: an error that also prints a truncated
document to stdout is worse than the panic in one respect — a caller piping the
output gets something that looks like data.

Making `emit_scalar` stop acting on the emitter's answer
(`.filter(|_| false)`), in my own worktree:

```
test test_error_171 ... FAILED
test test_error_211 ... FAILED
test result: FAILED. 0 passed; 2 failed; 525 filtered out
```

Restored:

```
test test_error_171 ... ok
test test_error_211 ... ok
```

Both gate, and both gate on the *same* mechanism, which is the point of having
the TOML companion: the two formats share one code path and one error variant,
so a fix that only reached YAML would be caught.

### Documentation

`EU-RENDER-UNREPRESENTABLE` is added to `docs/reference/error-codes.md` and to
the `CATALOGUE` in `src/driver/error_codes.rs`, which is what the file's own
"to add a new code" comment asks for — section, `code()` mapping, catalogue
entry. The notes on the error ("render to a format that can carry the value",
"convert to a string first with `str`") are positive guidance with concrete
alternatives rather than an enumeration of what not to do.

### Renumbering applied

`errors/194_toml_large_uint` → **`errors/211_...`**, because 194 is
`errors/194_8a49h_library_blame`. The test function was still called
`test_error_194` after the fixture moved, so I renamed it to `test_error_211`
— a test whose name and fixture number disagree is how the next renumber goes
wrong.

### Conflict resolution

Three conflicts, all resolved **by region** and all additive: the
`EU-EVAL-CONTRACT` and `EU-RENDER-UNREPRESENTABLE` entries coexist in
`docs/reference/error-codes.md`, in the `CATALOGUE`, and in
`ExecutionError::code`. Verified by building, not by checking that the markers
were gone.

### Gates

- `cargo fmt --all` — clean

### Note for the queue, not for this PR

**#1091 is branched off this PR** and carries its two commits. Once this merges,
#1091 needs its remaining fixtures renumbered against the new master rather than
merged as-is — I am handling that on my side, but flagging it here so nobody
resolves the overlap twice.

PR #1090, opened 2026-07-27
fix(diagnostics): stop blob global-slot Smids destroying user call sites (eu-1tkk.7.21)

gmorpheme:
Fixes **eu-1tkk.7.21** (raised to P1). Filed along the way: **eu-1tkk.7.29** (P1, separate defect — see below).

**Do not merge — owner review.**

## Root cause

eu-1tkk.7.11 stamped a `Smid::global_slot` identity onto every blob-reconstructed prelude global (`StandardRuntime::globals`, and the matching xtask bytecode pre-encode) so the Phase 2 blame classifier had something to key on for prelude frames.

Both machines treat a closure's annotation as *"the source location now in effect"* and overwrite the caller's on entry — `bytecode::machine::handle_op{,_predecoded}` and `vm::Machine::step`:

```rust
if closure.annotation().is_valid() { state.annotation = closure.annotation(); }
```

A `Smid::global_slot` value **is** valid but denotes an identity, not a position: `SourceMap::source_info_for_smid` deliberately refuses to resolve it. So every call from user code into a prelude combinator replaced the caller's genuine call-site Smid with a location-free one.

For a combinator that fails immediately — no continuation stack, no annotated environment — that call site was the **only** user-file Smid in existence. `to_diagnostic` had nothing to blame anywhere and emitted no primary label at all.

`EU_ERROR_TRACE_DUMP=1` on `errors/094_io_no_flag`, blob mode, release:

```
0.13.2   error smid: smid=39 file=3 span=109..111 ann="result"   → ┌─ 094_io_no_flag.eu:4:9
master   error smid: smid=2147483705 (no source info)            → no label
```

`2147483705` is `0x8000_0039` — the `GLOBAL_SLOT_TAG` bit over prelude slot 57 (`io.shell`).

This is the predicted class (a synthetic annotation displacing the call-site `Ann`) in its blob-mode form.

## The fix, and why it is this shape

The stamp only ever pays for itself when the slot's name is in the blame table. `SourceMap::classify_frame` resolves a global-slot Smid through `slot_to_name` → `blame_by_name` and returns `FrameKind::Transparent` when **either** lookup misses — which is exactly what an *unstamped* (`Smid::default()`) prelude frame classifies as anyway.

Eleven of ~548 prelude globals declare a blame contract (`head`, `tail`, `lookup`, `lookup-or`, `drop`, `nth`, `foldl`, `foldr`, `map`, `filter`, `mapcat`). For the other ~537 the stamp changed **no** classification while destroying **every** call site into them.

So the stamp is gated on the blame table, in both places that mint it — they must agree or `embedded_bytecode_matches_fresh_encode` fails. eu-1tkk.7.11's frame naming is untouched for the combinators that actually use it (`test_193`'s `in 'nth'` path is byte-identical, verified below).

**2 stamping sites changed** (`StandardRuntime::globals` → HeapSyn path; xtask bytecode pre-encode → bytecode path). Each is independently fault-injected below. No intrinsic wrappers needed changing — `annotated_lambda` no longer appears in any of them; that earlier fix had already landed.

Notes:
- The xtask blame table moves from step 9b to step 7b so the pre-encode can consult it. Pure reordering, same expression.
- **No wire-format change.** The blob's shape is unchanged — only which forms carry an annotation. `lib/prelude.blob` is gitignored and regenerated by every CI job that needs it, so there is no stale-artefact path. `BYTECODE_WIRE_FORMAT_VERSION` is untouched.

## Before / after — all nine, blob mode, against 0.13.2

Primary-label line (`┌─`) only. Release builds, `prelude_blob_ok` confirmed in build-script output for every binary; 0.13.2 built from its own tag with its own `prelude-compile`.

| fixture | 0.13.2 blob | master blob | **this PR** |
|---|---|---|---|
| 047_str_plus_plus | none | none | none |
| 072_io_fail | none | none | none |
| **094_io_no_flag** | `eu:4:9` | **none** — regression | **`eu:4:9`** ✅ |
| 106_io_check_fail | none | none | none |
| 115_render_as_invalid_format | none | none | none |
| 124_dbg_io_map | n/a (exit 0, `▶` output) | n/a | n/a |
| 125_dbg_io_function_wrap | n/a (exit 0, `▶` output) | n/a | n/a |
| 142_render_unknown_format | none | none | none |
| **168_io_shell_missing_cmd** | `eu:3:7` | **none** — regression | **`eu:3:7`** ✅ |

### The brief's premise needs correcting

The dispatch stated six blob-mode regressions and cited `┌─ 072_io_fail.eu:3:7` as the 0.13.2 rendering for `072`. Measured against a real 0.13.2 blob build, **there are two**: `094` and `168`. `168`'s 0.13.2 location is `168_io_shell_missing_cmd.eu:3:7` — the same `3:7`, which looks like where the `072` attribution came from.

The bead's own wording is the accurate one: the nine went *`primary: prelude` → `primary: none`*, and that is a **source-mode** statement. In source mode `047`/`094`/`115`/`142`/`168` had prelude primaries (correctly suppressed now); `072`/`106` never had a location in either mode; `124`/`125` produce no diagnostic at all (exit 0 `dbg` output, byte-identical to 0.13.2 in both modes).

So `047`/`072`/`106`/`115`/`142` are **not** blob regressions and are **not** this class. I verified that directly: with the global-slot stamping removed entirely (experiment, not shipped) all five remain locationless. Their causes are different and are recorded below.

## Snapshot delta — nothing else moved

PR #1085 is unmerged, so `cargo xtask diag-snapshot` does not exist on master. I captured the equivalent by hand: **every** `tests/harness/errors/*.eu` (179 fixtures) × 2 targets × 2 prelude modes = 716 runs, uniform invocation, master binary vs this branch.

The complete diff is:

- `168_io_shell_missing_cmd` gains its four primary-label lines
- thread-id noise from the pre-existing `src/export/yaml.rs:46` panic (eu-1tkk.7.20)

`094` is not in that diff only because the uniform capture passes `--allow-io` to everything, so it never reaches the no-flag path; it is covered by the targeted capture in the table above.

Source-prelude mode across the nine is **byte-identical** to master, as it must be — the change only affects blob reconstruction.

## Tests

**Where the location is asserted, and why it is not the `.expect` sidecar.** My first push pinned `file:line:col` in the `.expect` files and CI failed it — correctly. The "Test Suite" job is *deliberately* blob-less, to keep the source-prelude fallback exercised, and an `.expect` sidecar has no notion of prelude mode: one regex has to hold in both. These locations only exist under a blob. In source-prelude mode the annotation is a real prelude source position, which `to_diagnostic` correctly suppresses, so there is no `┌─` label to pin. Exactly and only my three sidecars failed (504 passed, 3 failed); the production change was green everywhere else.

So the location assertions live in `tests/diagnostics_blame_plumbing_test.rs`, gated wholesale on `#[cfg(prelude_blob_ok)]` — the existing home for blob-mode Smid-reconstruction assertions, run by CI's "Diagnostics blob-mode integration" job.

- **new** `blob_mode_immediate_prelude_failure_blames_the_user_call_site` — calls into the prelude from inside a *user-defined helper*, so it pins that the label tracks the real call site (`4:12`) rather than the top-level target
- **new** `blob_mode_io_permission_refusal_blames_the_user_call_site` — the `094` specimen, pins `2:9`
- both read the `┌─` **primary-label line** via `primary_label_position`. A `stack trace:` note renders locations as `- name at file:line:col` with no `┌─`, so neither can pass on the note alone.
- **new** fixture `tests/harness/errors/194_1tkk_7_21_blob_call_site.eu` + `test_194_1tkk_7_21_blob_call_site` — mode-agnostic wording assertion, gates that the fixture runs in both modes
- 5 unit tests for `should_stamp_slot` (declared / undeclared / unnamed / out-of-range / no blame table)

`094`/`168`/`194` sidecars keep mode-agnostic wording assertions with a comment pointing at where the location is really pinned. The remaining fixtures keep wording-only too: five have no location to pin (see eu-1tkk.7.29 and below), two produce no diagnostic. Pinning "no location" would freeze in a defect. Corpus-wide location assertions remain eu-ziq8o's.

One pre-existing footgun fixed while I was in there: `#[cfg(prelude_blob_ok)]` gates that file on the blob being *embedded* (build-time), but `EU_SOURCE_PRELUDE` is a *runtime* override that makes the spawned `eu` ignore it — so `EU_SOURCE_PRELUDE=1 cargo test --test diagnostics_blame_plumbing_test` failed all four tests, including the two eu-1tkk.7.11 ones that predate me. CI never hits it (its source-prelude jobs generate no blob, so the file compiles out), but a developer with the variable exported sees four spurious failures. Both child-process helpers now clear it.

## Fault injection

Each site reverted independently, with a **full rebuild including `prelude-compile`** each time (the bytecode annotations are baked at blob-compile time, so a rebuild without it proves nothing), and `prelude_blob_ok` re-confirmed from the build-script output on every cycle. Re-run cleanly after I noticed a first attempt had overlapped the cargo target lock with another build.

Harness gate (`test_error_094`, `test_error_168`, `test_194_1tkk_7_21_blob_call_site`):

| state | default (bytecode) | `EU_HEAPSYN=1` |
|---|---|---|
| fix in place | 3 pass | — |
| **inject 1**: `should_stamp_slot` defeated in `runtime.rs` | 3 pass¹ | **3 FAIL** |
| **inject 2**: gate defeated in xtask pre-encode | **3 FAIL** | — |
| restored | 3 pass | — |

¹ Not a weak test: the bytecode engine executes the blob's *pre-encoded* bytecode, so `StandardRuntime::globals()` feeds the HeapSyn path only. Site 1 is genuinely load-bearing there; site 2 is load-bearing for the default engine. Each site fails on the engine it actually serves.

Blob-gated location gate (`diagnostics_blame_plumbing_test`), which also demonstrates the fix does not undo eu-1tkk.7.11:

| state | the 2 new call-site tests | the 2 existing `nth` identity tests |
|---|---|---|
| fix in place | pass | pass |
| **inject**: xtask gate defeated | **FAIL** | **pass** |
| restored | pass | pass |

That second column is the point: with the stamp restored to everything, `nth`'s global-slot identity still resolves to a declared `:boundary` — so gating the stamp on the blame table takes nothing away from Phase 2.

## Validation

Full `cargo test` in blob mode, plus the gating tests under `EU_HEAPSYN=1`, `EU_PREDECODE=0`, and `EU_SOURCE_PRELUDE=1`. Exactly the three pre-existing failures the brief predicts, and nothing else:

- `diagnostics_invariants::secondary_labels_do_not_excerpt_library_internals`
- `diagnostics_invariants::debug_trace_restores_the_uncurated_trace`
- `harness_test::test_193_1tkk_7_12_curated_trace`

The gating tests pass under all four configurations — default, `EU_HEAPSYN=1`, `EU_PREDECODE=0`, `EU_SOURCE_PRELUDE=1`.

All three are fixed by #1079. `.github/workflows/build-rust.yaml` (the `diagnostics-blob-mode` job) already documents the first two as known blob-mode failures. For `test_193` I checked the actual stderr rather than trusting the label: **byte-identical between the master binary and this branch**, which also confirms the `nth` boundary path is undisturbed.

`cargo fmt --all` clean; `cargo clippy --workspace --all-targets -- -D warnings` clean, no `#[allow]`.

## Filed: eu-1tkk.7.29 (P1) — not fixed here

While tracing the residual cases I found a **worse** defect, on master as well as this branch. `StgArena` reconstruction elides `Ann` nodes and `Lambda` annotations on the blob path, but **not** the inline Smids that `ArenaStgSyn::DirectApp { smid, .. }` and `ArenaStgSyn::LookupLit { smid, .. }` carry in their own fields — those are copied through verbatim as raw xtask `SourceMap` indices.

Today they usually resolve to nothing (`115`/`142` carry `smid=3055` and print no label). But user-file Smids are minted in the **same** `u32` space from 1, so a large enough user file makes the stale index valid:

```
# 800 pad declarations, then: result: "hello" render-as(:badformat)

error: unknown render format 'badformat'
  ┌─ big.eu:1:9
  │
1 │ pad1: 1 + 1
  │         ^
```

A confidently wrong location is worse than none. Not fixed here: it is a different root cause, the fix needs an explicit blob-baked discriminator (`reconstruct_form` is also used by the `prelude_slot_override` path, whose Smids *are* valid), and it overlaps `resolve_trace_entry` in #1079/#1084.

Residual causes for the other locationless fixtures, for the record: `072`/`106` raise with `Smid::default()` — the annotation is never set, not clobbered; `047` fails inside `head`, a *declared boundary* combinator, so it keeps its stamp by design and the competition between trace naming and call site is real there — resolving it needs the two-field split in the machines, which is fenced from me.

## Fences

`src/eval/bytecode/machine.rs`, `src/common/sourcemap.rs`, `src/eval/stg/compiler.rs`, `src/eval/stg/meta.rs`, `src/eval/machine/env_builder.rs`, `src/core/inline/`, `src/core/expr.rs`, `src/export/yaml.rs`, `lib/prelude.eu`, `src/driver/check.rs`, `tests/diagnostics/`, `BYTECODE_WIRE_FORMAT_VERSION` — **none touched**. The fix is deliberately upstream of the machines: it stops the misleading annotation being minted rather than teaching the machines to ignore it.

## On eu-7fjiq — different defect, left alone

`target_annotation` (`vm.rs:1404`) is a **fallback**, consulted only when `self.annotation` is already invalid; it fails to *find* an `Ann` that is present but not leading. Mine is the opposite polarity: a valid-but-location-free Smid **overwriting** a good annotation that was found correctly. `target_annotation` is never reached in any of my fixtures — the dumps show the annotation being *set* on entry, so the fallback never runs. Decisively, eu-7fjiq is source-mode-only (blob byte-identical across all 179) and mine is blob-mode-only: they live in different prelude representations, so no single fix reaches both. Left untouched.


gmorpheme:
## Verification of the #1079/#1083 tension — CONFIRMED, and broader than previously reported

Reviewed as part of clearing the 0.14 queue. I was asked to verify or refute the earlier finding that this PR is in tension with #1079/#1083, **not** to act on it. Leaving this PR open for the owner to decide.

### What I ran

Built an integration in an isolated worktree off master `a1af1f0b`:

- `#1079` (eu-7x0r) + `#1080` (eu-8a49h) + `#1083` (eu-og3u6) + `#1084` (eu-r4647)
- regenerated `lib/prelude.blob` via `cargo run --package xtask -- prelude-compile`
- rebuilt and confirmed `cargo:rustc-cfg=prelude_blob_ok` in the build-script output (not inferred from the file existing)

Then measured the five `many_decls_*` fixtures from #1085's snapshot corpus, before and after merging this branch, regenerating the blob and rebuilding each time.

### Result: 4 of the 5 fixtures regress

| fixture | without #1090 | with #1090 |
|---|---|---|
| `many_decls_late_nth` | `result` @ **907:12** correct | `pad0470` @ 477:16 **wrong** |
| `many_decls_early_nth` | `result` @ **5:12** correct | `pad0469` @ 475:14 **wrong** |
| `many_decls_middle_nth` | `result` @ **455:12** correct | `pad0469` @ 475:14 **wrong** |
| `many_decls_nested_user_fn` | `result` @ **606:12** correct | `pad0469` @ 475:10 **wrong** |
| `many_decls_lookup_miss` | `result` @ 604:9 correct | `result` @ 604:9 unchanged |

The earlier report identified only `many_decls_late_nth`. It is three fixtures worse than that.

### Mechanism

The correct frame is not lost — it is *outvoted*. With #1090 the trace becomes:

```
= stack trace:
  - result at many_decls_late_nth.eu:907:12     <- correct, still first
  - in 'nth' at [prelude]:1386:1
  - pad0455 at many_decls_late_nth.eu:462:10    <- leaked
  - in 'nth' at [prelude]:1386:1
  - pad0470 at many_decls_late_nth.eu:477:16    <- leaked, becomes primary
  - in 'nth' at [prelude]:1386:1
```

Narrowing the `Smid::global_slot` stamp to only the bindings carrying a declared blame classification leaves the other prelude globals un-stamped. `cargo xtask prelude-compile` reports `blame classifications retained in blob: 11` against `352` prelude bindings — so 341 globals lose the stamp. Their baked `DirectApp`/`LookupLit` Smids are indices into *xtask's* `SourceMap`, and in any user file whose own `SourceMap` grows past them they resolve against unrelated user declarations again. That is exactly the aliasing class #1079's `StgArena::rebase_baked_smid` was introduced to close; the stamp was load-bearing beyond naming.

### CI visibility — this is better than previously reported

The earlier note said CI cannot see this. That is true only of the specific order "#1090 alone on master":

- **#1090 on master today:** 19/19 non-skipping checks green. Confirmed.
- **#1090 on top of the integration:** fails loudly, on both engines.

```
failures:
    many_binding_unit_anchors_blame_on_the_calling_declaration_bytecode
    many_binding_unit_anchors_blame_on_the_calling_declaration_heapsyn
test result: FAILED. 1 passed; 2 failed        (diagnostics_trace_anchor_test)

failures:
    blob_mode_trace_frames_cannot_alias_unrelated_user_declarations
test result: FAILED. 6 passed; 1 failed        (diagnostics_blame_plumbing_test)
```

So once #1083 and #1079 are on master, this tension is **CI-visible and cannot land silently**. The committed 10-padding-declaration harness fixture is indeed too small to see it — the 900-declaration generated fixture in `tests/diagnostics_trace_anchor_test.rs` is what catches it.

### Suggested direction (not a merge decision)

eu-1tkk.7.21's own goal — a prelude combinator that fails immediately must still blame the user's call site — looks sound and its fixture is a genuine gap. The problem is the chosen remedy: narrowing the stamp trades a real fix for a regression in a larger class. Worth exploring whether the call site can be preserved *without* dropping the stamp on the other 341 globals, e.g. by keeping the stamp universal and making the machine not treat a global-slot annotation as displacing a live call-site annotation on entry.

Rebasing this branch onto master once the queue lands will surface the three failures above directly.

gmorpheme:
Closing this PR without merging.

**Reason: it is in direct tension with #1079/#1083, which have now merged.** Verification was run on an integration branch of #1079+#1080+#1083+#1084, with the blob regenerated per arm and `prelude_blob_ok` confirmed from build-script output. Four of the five `many_decls_*` corpus fixtures regress with this PR applied:

| fixture | without #1090 | with #1090 |
|---|---|---|
| `late_nth` | `result@907:12` | `pad0470@477:16` |
| `early_nth` | `result@5:12` | `pad0469@475:14` |
| `middle_nth` | `result@455:12` | `pad0469@475:14` |
| `nested_user_fn` | `result@606:12` | `pad0469@475:10` |
| `lookup_miss` | correct | unchanged |

**Mechanism:** the correct frame is not lost, it is outvoted. Narrowing the `Smid::global_slot` stamp leaves 341 of 352 prelude globals un-stamped, so their baked `DirectApp`/`LookupLit` Smids alias user declarations again; two bogus pad frames land *after* the correct one and the primary label takes the last.

**This was not visible to the author.** Measured against master, where those fixtures were already wrong, the change is an improvement. It is only against the merged #1079/#1083 blame fixes that it regresses — and #1090 alone on master passes every test, including its own and #1083's, because the committed harness fixture has too few padding declarations for the leak to reach a different one.

**The good news:** now that #1083 has landed, the tension is no longer silent. `diagnostics_trace_anchor_test` and `diagnostics_blame_plumbing_test` fail loudly on both engines if this mechanism is reintroduced, so any future attempt is gated by CI rather than by manual verification.

**eu-1tkk.7.21 remains open and unfixed.** The six IO fixtures — `072_io_fail`, `094_io_no_flag`, `106_io_check_fail`, `124_dbg_io_map`, `125_dbg_io_function_wrap`, `168_io_shell_missing_cmd` — had a correct blob-mode location at 0.13.2 and still have none. This PR restored all six, which is better than its own author measured, so the diagnosis here is sound and worth reusing; it is the *mechanism* that cannot stand. A replacement fix must restore those locations without un-stamping the globals that suppress the annotation leak.

Closed by the coordinator on the owner's instruction.

PR #1091, opened 2026-07-27
fix(export): html shape panic, EDN integer corruption, TOML null conflation (eu-1tkk.7.24/23/28)

gmorpheme:
**Stacks on #1089** (eu-1tkk.7.20) — branched from it, not from master, so
the `Emitter::unrepresentable` hook it introduces is present. Review #1089
first; the diff against that branch is 23 files, and against master 27.
Do not merge — owner review.

Three defects from #1089's audit: **eu-1tkk.7.24**, **eu-1tkk.7.23**,
**eu-1tkk.7.28**.

## 1. `eu -x html` panicked on any ordinary document (eu-1tkk.7.24, P1)

```
$ printf 'main: { a: 1 b: 2 }\n' > h.eu && eu -x html h.eu
thread '<unnamed>' panicked at src/export/markup.rs:170:30:
unexpected block end event in state OddAttrAccumulation([], "main")
thread 'main' panicked at src/bin/eu.rs:43:10: main thread panicked: Any { .. }
```

`eu -x html file.eu` with no target renders the whole unit, whose root is a
block — so **every default html invocation aborted**. A bare scalar was the
mirror image: exit 0 and zero bytes, silently.

`Emitter::unacceptable(&Event)` lets an emitter reject an event given what
it has already received — document *shape*, where `unrepresentable` is about
a single value. The markup emitter now reports `EU-RENDER-SHAPE`:

```
error[EU-RENDER-SHAPE]: cannot render this document as html: the value to render is a block, but markup output needs a hiccup element — a list whose first item is the tag
  = html output renders hiccup markup: a list of tag, attribute block and contents, e.g. '[:div, { id: "top" }, "hello"]'
  = select the markup value with a target or '-e', e.g. 'eu -x html -e page.eu'; rendering a whole unit gives html the unit's block, which is not markup
  = to render arbitrary data instead, choose a format that accepts any shape, such as 'yaml', 'json' or 'text'
```

Both `panic!`s in `markup.rs` are gone; the arms they guarded now drop the
event and are unreachable behind the check.

**On the missing source location.** This diagnostic usually has none, and I
have not manufactured one. `EU_ERROR_TRACE_DUMP=1` shows the error's Smid
invalid and both traces empty: the document root events come from the render
pipeline, not a user expression. Where an annotation *is* recorded it is the
root expression's own span, so it is correct when present. Giving it a
location reliably is annotation-resolution work, which is fenced to another
agent. The 195/196 sidecars say this explicitly and tell readers not to copy
them as the eu-ziq8o pattern.

### The coverage gap that let it survive

There was **no test of `-x html` of any kind** — success or failure. Added:

- `tests/harness/191_html_export.eu` — the success path, asserting exact
  serialised html for four hiccup shapes (element with attributes, bare tag,
  nesting, text escaping) via `render-as(:html)` so the assertions are
  ordinary string equalities. This also covers the capture-emitter path.
- `errors/195_html_block_shape.eu`, `errors/196_html_scalar_shape.eu`.

## 2. EDN silently corrupted large integers (eu-1tkk.7.23, P1)

`9999999999999999999` exported as `10000000000000000000` — a wrong number,
no panic, nothing to notice it by. Fixed **losslessly**, not by erroring:
EDN integers are arbitrary precision, so `Value::BigInt` emits the exact
digits with EDN's `N` suffix. `num-bigint` becomes a direct dependency,
already in the tree via `edn-format` (whose own requirement is 0.4.3), so
no new supply-chain surface.

**The same `as_f64` pattern across every exporter, as asked:** json passes
the `serde_json::Number` straight through; text and eu format it directly;
yaml and toml short-circuit on the range predicate from #1089 before
reaching `as_f64`. **EDN was the last lossy one.** The integer guidance in
the catalogue claimed EDN was safe before it actually was — corrected.

## 3. TOML rendered null as `""` (eu-1tkk.7.28, P2)

`{ a: null b: "" }` produced two identical `""`, indistinguishable to any
reader and on re-import.

**Decision: error**, consistent with the integer case. The alternative worth
weighing was omitting the key — TOML's own way of spelling "no value" — but
that is also lossy in eucalypt terms (re-import yields a missing key, so
`data.a` becomes a lookup failure rather than null) and it silently changes
the document's shape. Erroring loses nothing and lets the caller decide;
the notes tell them how:

```
error[EU-RENDER-UNREPRESENTABLE]: cannot represent this value in TOML output: TOML has no null; every key must have a value
   ┌─ 197_toml_null.eu:15:7
   = give the key a value the format can carry, or drop it from the block before rendering
   = 'yaml' and 'json' output both have a null, and 'edn' renders it as 'nil'
```

This is the one behaviour change here that could affect working code. Anyone
exporting nullable data to TOML currently gets `""` — but cannot distinguish
it from a real empty string, so cannot be relying on it correctly.

## Design note: emitters now own their remediation

My first cut selected the null-vs-integer notes in `error.rs` by testing
whether the reason string **contained "null"**. That is exactly the
silently-wrong-on-rewording failure this work is supposed to remove, so I
replaced it: `unrepresentable`/`unacceptable` return a `Rejection { reason,
notes }` and `error.rs` just forwards the notes. This does change #1089's
hook signature — deliberate, and called out because that PR is under review.

## Fault injection — per site, rebuilt each time

Disk checked before each long run (58 → 32 GiB free, never near the limit).

| Site | Injection | Result |
|---|---|---|
| html block-root guard | disabled | 195 **FAIL** (caught by the second guard, different message) |
| html, full pre-fix state | both guards off **+ `panic!` restored** | 195 **FAIL — exit 101**, the original abort |
| html scalar-root guard | disabled | 196 **FAIL — exit 0**, the original silent-empty behaviour |
| html serialiser | doctype altered | 191 **FAIL** |
| EDN BigInt | `as_f64` fallback restored | 192 **FAIL**, output `10000000000000000000` |
| TOML null guard | disabled | 197 **FAIL — exit 0**, output `""` |

All restored and re-verified PASS.

**Two things this found that a pass/fail check alone would not:**

1. **Fixture 195's first draft could not have caught its own bug.** A single
   flat block never reached the `panic!` — it fell out of the state machine
   as empty output and exit 0. The panic needs a block *inside* a block,
   which is what the unit wrapper produces. The fixture is now nested, and
   only then does the injection give exit 101.
2. **Fixture 192's `t-edn-not-rounded` was too weak.** It compared against
   `"10000000000000000000N"`, but the bug emitted no `N` suffix, so the
   assertion was satisfied *by the bug*. Now a substring check, and it fails
   under the injection alongside `t-edn-exact`.

**Injection C repeated on 197** (strip the error's Smid): still **PASSES**,
same as 171 — `to_diagnostic` recovers the location from the traces. Noted
in the sidecar and added to **eu-ziq8o** as a comment, per your instruction,
along with the `┌─[^\n]*file\.eu:L:C` anchor and the strip-the-label-line
verification as the pattern to adopt corpus-wide.

## Validation

| prelude | engine | result |
|---|---|---|
| source | bytecode | all green (512 harness) |
| blob | HeapSyn | same 3 |

The three are the two `diagnostics_invariants` and
`test_193_1tkk_7_12_curated_trace`. Clean under `EU_GC_VERIFY=2
EU_GC_STRESS=1`. `cargo fmt --all --check` and `cargo clippy --workspace
--all-targets -- -D warnings` clean, no `#[allow]`.

Fixture 197 initially failed under HeapSyn — the eu-1tkk.7.26 block-literal
label gap again, same as 194 in #1089. Moved to the document root where both
engines agree; noted on that bead as related, and otherwise left alone as
instructed.

## Fixture numbers taken

`tests/harness/`: **191**, **192**. `tests/harness/errors/`: **195**, **196**,
**197** (plus **194** already in #1089). The two directories number
independently — errors/ was at 194, harness/ at 190. Happy to renumber in a
coordinated sweep.

## Not touched

**eu-1tkk.7.25 (broken pipe)** — proposal below, awaiting your decision, as
instructed. **eu-1tkk.7.26** — left alone, noted as related.
**eu-1tkk.7.27** — no action. No fenced file touched.


gmorpheme:

`17f84b21` (which now carries #1089), with `lib/prelude.blob` regenerated after
the branch switch and `prelude_blob_ok` confirmed from the build-script output.

### Three defects, and the EDN one is the worst of them

The html panic and the TOML null conflation are both bad, but **the EDN
rounding is the one that should worry a reader most**: `9999999999999999999`
was exported as `10000000000000000000`. No panic, no error, no exit code —
just a different number in the output file. Every other defect in this queue
announces itself; this one does not.

The fix is right in kind, not just in effect: EDN integers are
arbitrary-precision by specification, so `Value::BigInt` is the *correct*
representation rather than a workaround, and the `N` suffix is the spec's own
notation for it. This is not "make the error go away", it is "use the type the
format actually has".

**Fixture 213's second check is the best piece of test design in this PR.**

```
t-edn-not-rounded: not(edn-out str.contains?("10000000000000000000"))
```

with the reasoning stated in the fixture: the old fallback emitted
`10000000000000000000` **without** an `N` suffix, so an inequality against the
suffixed form would have been satisfied *by the bug itself*. That is the kind
of thing that is invisible unless someone thinks about what the broken output
actually looked like, and it is why I am confident this test gates rather than
merely passes.

The boundary checks are right too: `123` and `i64::MAX` itself must keep the
plain integer form, so the BigInt path cannot swallow ordinary integers.

### The html shape fix

`unacceptable(event)` is the shape-level analogue of #1089's
`unrepresentable(primitive)`, and generalising the rejection into a `Rejection`
type carrying notes is the right move — the alternative would have been a
second, parallel mechanism for the same job.

The state predicate is precise. `at_root` is `stack.is_empty() &&
result.is_none()`, so it fires only at the document root and cannot reject a
block or scalar appearing legitimately *inside* an element. The two extra
guards cover exactly the two states that previously reached a `panic!` — an
attribute block still open where an element was expected, and a block closed
where an attribute block was expected.

Both former `panic!` arms become `return`, documented as unreachable because
`unacceptable` rejects first. Same judgement as #1089 and same reasoning: an
unreachable branch that drops an event is strictly better than one that aborts
the process, because "unreachable" is a claim about today's call graph.

The remediation notes are genuinely useful rather than boilerplate — they name
the hiccup shape with a concrete example, explain that rendering a whole unit
gives html the unit's *block* (which is the actual mistake most people will
have made), and point at formats that accept any shape. Positive guidance with
alternatives, which is the house style.

**The scalar case is a behaviour change worth naming explicitly.** Rendering a
bare scalar as html previously produced **no output and exit 0**. It is now a
diagnostic. Someone piping `eu -x html` on a scalar and checking the exit code
will see a change — but silently producing nothing and reporting success is not
behaviour worth preserving, and I would rather this be stated in the review than
discovered in the release notes.

### Fixture 212 is the missing half

Before this PR, **nothing exercised `-x html` or `render-as(:html)` at all**.
That is why the markup emitter could abort the process on any non-hiccup
document without a single test noticing — the exporter had error coverage
arriving in the same PR as its first success coverage.

The success fixture pins exact serialised output for four shapes (element with
attributes, bare tag, nesting with attribute rendering, and text escaping),
which is the right level: a shape-only assertion would not have caught the
escaping.

### `num-bigint` as a direct dependency

New direct dependency, but **already in the tree transitively via
`edn-format`**, whose own requirement is `0.4.3`. So this adds no new
supply-chain surface — it promotes an existing transitive dependency to direct
so the exporter can name `BigInt`. The `Cargo.toml` comment says exactly that,
which is the right thing to leave for the next person reading the manifest.
The Security Audit job passes.

### `INTEGER_RANGE_NOTES` in `export/mod.rs`

Shared remediation for YAML and TOML, hoisted so the two cannot drift. Same
instinct as the per-format overflow predicates in #1089 and correct for the
same reason. Note it now also mentions `edn` as a format that carries these
integers — which is only true *because* of this PR, so the two halves are
consistent.

Two injections in my own worktree, each targeting a different one of the three
defects:

| Injection | Result |
|---|---|
| EDN `BigInt` arm removed (the `as_f64` rounding returns) | `test_213_edn_large_integer` **FAILED** |
| `MarkupEmitter::unacceptable` returns `None` | `test_error_214` **and** `test_error_215` **FAILED** |

Restored, all five fixtures pass (212, 213, 214, 215, 216). Each fix has a test
that fails without it.

### Fixture renumbering applied

All five of this PR's own fixtures collided with master:

- `191_html_export` → **212**
- `192_edn_large_integer` → **213**
- `errors/195_html_block_shape` → **214**
- `errors/196_html_scalar_shape` → **215**
- `errors/197_toml_null` → **216**

Test functions renamed to match (`test_error_195/196/197` →
`test_error_214/215/216`), and the two fixture doc strings that still opened
with the old numbers corrected. A test whose name, doc string and fixture
number disagree is how the *next* renumber goes wrong.

Because this branch is stacked on #1089, its eu-1tkk.7.20 commits are now
ancestors via master, so only this PR's own fixtures needed touching — the
overlap was not resolved twice.

### Gates

- `cargo fmt --all` — clean

PR #1092, opened 2026-07-27
fix(export): report write failures instead of panicking; quiet exit on a closed pipe (eu-1tkk.7.25)

gmorpheme:
**Stacks on #1091**, which stacks on #1089. Base is the #1091 branch. Review
those first; the diff against #1091 is 17 files. Do not merge — owner review.

Fixes **eu-1tkk.7.25**. Option B as directed, with A's special case folded in.

```
$ eu -x yaml big.eu | head -2
thread '<unnamed>' panicked at src/export/yaml.rs:127:44:
failed to write YAML output: Os { code: 32, kind: BrokenPipe, ... }
thread 'main' panicked at src/bin/eu.rs:43:10: main thread panicked: Any { .. }
```

Now: exit 0, nothing on stderr, in all six document formats.

## Constraint 1 — exit 0, and the grep you asked for

**Nothing asserts 141 or relies on a signal death.** The error sidecars assert
only `exit: 0` (10 fixtures) and `exit: 1` (168) — no 101, no 141. `SIGPIPE`,
`BrokenPipe` and bare `141` appear nowhere in `tests/`, `.github/`, `scripts/`
or `xtask/`. Five scripts use `set -o pipefail`, none of them pipes `eu`.
Exit 0 is safe.

`ExecutionError::OutputStreamClosed` is recognised in `diagnose`, beside the
existing Ctrl-C case — one place, at the driver boundary, not per emitter.

## Constraint 2 — EPIPE must not swallow real errors

`OutputStreamClosed` and `OutputWriteFailed` are **separate variants**, and
only the former reaches the quiet-exit branch. Verified both ways:

- **Unit** (`src/export/mod.rs::write_failure_tests`): every emitter driven
  through a `Write` that fails with a chosen `ErrorKind`. `BrokenPipe` must
  classify as a pipe; `OutOfMemory`, `PermissionDenied` and `WriteZero` must
  not. This is how the ENOSPC case gets covered at all — it cannot be
  provoked from a shell without filling a filesystem.
- **Injection of exactly the failure mode you named**: widening
  `is_broken_pipe` to `matches!(self, RenderError::Io(_))` — the broad
  `Err(_) => exit(0)` — fails 2 of the 3 unit tests with
  `yaml: OutOfMemory was misclassified as a broken pipe`.
- **End-to-end**: injecting a forced `OutOfMemory` into the yaml write path
  gives `error: failed writing output: injected ENOSPC-like failure` and
  **exit 1** — both unpiped and piped into a short reader, so the pipe
  handling does not mask it. A forced `Serialisation` error in the json path
  likewise gives a diagnostic and exit 1.

## Constraint 3 — the non-IO `.expect`s, and reaching toml/edn

All converted, none left on the write path:

| Site | Now |
|---|---|
| `yaml.rs` emit | `RenderError::Serialisation` |
| `json.rs` serialise | `RenderError::Serialisation` |
| `eu.rs` render + UTF-8 decode | `RenderError::Serialisation` |
| `markup.rs` serialise | propagated |
| `yaml/json/toml/edn/eu/text` writes | propagated |

`RenderError` gains `Serialisation` because in those cases nothing is wrong
with the stream — the document could not be turned into bytes.

**toml and edn are now reached, and that mattered.** See below.

## Constraint 4 — per-format injection, and a test that could not fail

`tests/broken_pipe_test.rs` drives the real binary per format: spawn, consume
a prefix, drop the read end, check exit and stderr.

**The first version could not fail for two formats.** It read a *line*. EDN
renders any document as a **single line**, and TOML rendered the fixture as
**two** — so `read_line` swallowed the whole document and the child finished
writing successfully. Restoring the pre-fix `.expect` in `toml.rs` left the
test **passing**; same for `edn.rs`. Reading a fixed 64-byte prefix instead
leaves >200 KB unwritten in every format. Injection per format, rebuilt each
time:

| Injection | Result |
|---|---|
| `yaml.rs` write `.expect` restored | **FAIL — exit 101** |
| `json.rs` | **FAIL — exit 101** |
| `toml.rs` | **FAIL — exit 101** (passed before the fixture fix) |
| `edn.rs` | **FAIL — exit 101** (passed before the fixture fix) |
| `text.rs` | **FAIL — exit 101** |
| `eu.rs` | **FAIL — exit 101** |
| `markup.rs` serialise `.expect` | **FAIL** (html unit test) |
| `is_output_stream_closed` → `false` | **FAIL** — quiet-exit test only, correctly not the never-panics test |
| `is_broken_pipe` widened to any `Io` | **FAIL** — 2 of 3 unit tests |

All restored and re-verified green.

## The fenced file is untouched

`src/eval/bytecode/machine.rs` needed no edit. Its three capture-lifecycle
calls are on a concrete `OwnedCaptureEmitter`, so inherent `stream_start` /
`stream_end` on that type shadow the now-fallible trait methods and stay
infallible. That is correct on the merits, not a dodge: a capture writes into
a `Vec<u8>`, whose `Write` cannot fail. They carry a `debug_assert` on the
inner result so a future change that makes it fallible is caught.

## Design notes

- `stream_end` is where the document-producing emitters actually write, so a
  failure there must not be lost because evaluation itself succeeded — that
  is how a truncated file gets reported as success. `finish_stream` folds it
  into the run's result; an error already in flight wins, being earlier.
- `wasm_pipeline.rs` reports both stream errors as `PipelineError` rather
  than ignoring them, even though its sink is in-memory today.

## Validation

| prelude | engine | result |
|---|---|---|
| source | bytecode | all green (512 harness, 1356 unit, 2 broken-pipe) |
| blob | HeapSyn | same 3 |

Clean under `EU_GC_VERIFY=2 EU_GC_STRESS=1`. `cargo fmt --all --check` and
`cargo clippy --workspace --all-targets -- -D warnings` clean, no `#[allow]`.
Disk 65 GiB throughout.

## Still open in export, reported as unproven rather than claimed

Three panics remain in `src/export/`, all on the block-key path:
`json.rs:12` and `toml.rs:61` (`as_str().unwrap()`) and `eu.rs:33`
(`panic!("non-string/symbol used as block key in eu export")`).

I tried to reach them and **could not**: YAML keys that look boolean, float
or date (`true:`, `1.5:`, `2020-01-01:`) all import as strings, so `as_str()`
succeeds. Recording that as unreproduced rather than repeating the
"reasoned but unreached" claim that turned out to be half-right for toml/edn
last round. Happy to file a bead if you want it tracked.


gmorpheme:

`c21e6d72` (which now carries #1096, the retarget of #1091), with
`lib/prelude.blob` regenerated after the branch switch and `prelude_blob_ok`
confirmed from the build-script output.

I have also **retargeted this PR at `master`** — it was opened against
`fix/furnace-export-fidelity-html-edn-toml`, one link in the
#1089 → #1091 → #1092 → #1093 stack. Its predecessors are on master now, so
the diff here is its own two commits alone.

### The defect

`eu … | head` — everyday CLI use — aborted with a Rust backtrace and **exit
101** in every format that writes a document. The document emitters buffer the
whole document and write it once at `stream_end`, so the write happens at a
point where nothing was checking the result.

### The design decision that matters

`Emitter::emit` now returns `Result<(), RenderError>`. That is a wide,
mechanical change across every emitter, and the temptation with a broken-pipe
fix is to special-case EPIPE at the one place it was observed. This does the
opposite: it makes *every* write failure representable, then classifies at the
driver boundary.

That distinction is the whole fix, and the PR gets the failure mode the right
way round:

```rust
pub fn is_broken_pipe(&self) -> bool {
    matches!(self, RenderError::Io(e) if e.kind() == io::ErrorKind::BrokenPipe)
}
```

A closed pipe is not an error — the reader has what it wanted — so it becomes
`OutputStreamClosed` and a silent exit 0. **Everything else is reported.** Had
this instead been "treat write failures at stream end as benign", a full disk
would exit 0 and the user would believe a truncated file was complete — which
is *strictly worse than the panic it replaces*, because the panic at least
announced itself.

The PR's own unit test says exactly that, and it is the test I would have
asked for:

```rust
/// The guard that matters: a real write failure must NOT be classified as
/// a broken pipe. Were it to be, the driver would exit 0 on a full disk
/// and the user would believe a truncated file was complete — strictly
/// worse than the panic this replaced.
fn a_real_write_failure_is_not_mistaken_for_a_broken_pipe()
```

`RenderError::Serialisation` as a variant distinct from `Io` is right too: a
serialiser refusing a document is not an output-stream problem, and collapsing
the two would make `is_broken_pipe` answer a question about the wrong thing.

### `finish_stream` — the subtle half

```rust
match (emitter.stream_end(), result) {
    (Err(e), Ok(_)) => Err(stream_write_error(e)),
    (_, other) => other,
}
```

Two properties, both correct and neither obvious:

- A write failure at `stream_end` is **not lost** when evaluation itself
  succeeded. That is precisely how a truncated file gets reported as success,
  and it is the case a naive `let _ = stream_end();` would miss.
- An error already in `result` **wins**, being earlier and more specific. So a
  genuine evaluation error is not masked by the write failure it caused.

`stream_end` is called on all four driver paths (bytecode, HeapSyn, and both
IO-yield branches), and every one now goes through `finish_stream`. I checked
each call site rather than the two the diff draws the eye to — a missed one
would silently keep the old swallow-the-error behaviour on that path only.

### Test design

Two layers, correctly divided:

- **Unit** (`src/export/mod.rs::write_failure_tests`) drives a `Write` that
  fails with a chosen `ErrorKind`. That reaches **ENOSPC and
  PermissionDenied**, which cannot be provoked from a shell without filling a
  filesystem. Six formats × three error kinds, plus `html` separately because
  it writes through a `MarkupSerialiser` rather than `writeln!`.
- **Integration** (`tests/broken_pipe_test.rs`) drives the real binary,
  because the bug lived in the interaction between emitter, driver stream
  lifecycle and process exit — a unit test over an emitter cannot reach that.

**The best thing in this PR is a constant and its comment:**

```rust
/// Deliberately a byte count and not a line: EDN renders any document as a
/// *single* line, and TOML renders this one as two, so a `read_line` would
/// swallow the whole document and the child would finish successfully —
/// leaving those formats untested. Verified by injection: with a line-based
/// read, restoring the pre-fix `.expect` in `toml.rs` or `edn.rs` left this
/// test passing.
const BYTES_TO_CONSUME: usize = 64;
```

That is an author who wrote the test, checked it actually failed without the
fix, found it did **not** for two of the six formats, and fixed the *test*.
Recording the near-miss in the source is worth more than the fix itself,
because the next person to touch this file now knows why it is a byte count.

`BIG_DOCUMENT` producing ≥200 KB so the write is guaranteed still in progress
when the pipe closes is the same care applied to the other end of the race.

Making `is_broken_pipe` classify every `Io` error as a pipe, in my own
worktree:

```
test a_real_write_failure_is_not_mistaken_for_a_broken_pipe ... FAILED
test html_reports_write_failures_too ... FAILED
test every_format_reports_a_broken_pipe ... ok
test result: FAILED. 1 passed; 2 failed
```

Exactly the right two fail, and the third correctly still passes — a broken
pipe *is* still reported as a broken pipe under that injection, so a test that
also failed would have been testing something other than what it claims.

Restored: 3/3 unit tests and both integration tests pass.

### Gates

- `cargo fmt --all` — clean

PR #1093, opened 2026-07-27
fix(export): report non-text block keys instead of panicking (eu-1z503)

gmorpheme:
**Stacks on #1092** → #1091 → #1089. Base is the #1092 branch; the diff
against it is 13 files. Kept separate rather than amended into #1092 so that
PR's verified state stays intact. Do not merge — owner review.

Fixes the block-key panics. **The bead is `eu-1z503`** — the id I was given
(`eu-tqbcx`) does not resolve; I raised `eu-1z503` from P2 to P1 and recorded
the reproduction there.

## Reachable — and your hunch about where to look was half right

You were right that the import path was the wrong place, and right about the
reason: **every importer normalises or rejects non-text keys.** EDN in
particular does *not* let one through — `src/import/edn.rs::value_to_key`
returns `SourceError::InvalidBlockKey` for anything that is not a Character,
String, Symbol or Keyword. So the EDN route is closed too.

The way in is the second thing you suggested — constructing the block so the
key never passes an importer — and it needs nothing exotic, just the
documented prelude constructors, which take an arbitrary key:

```
` { target: :main }
main: kv-block(1, "a")
```

```
eu -x json f.eu  →  panicked at src/export/json.rs:12:23   exit 101
eu -x toml f.eu  →  panicked at src/export/toml.rs:61:23   exit 101
eu -x eu   f.eu  →  panicked at src/export/eu.rs:33:18     exit 101
```

Also reproduced with a boolean key, a list key, and via `block([[1, "a"]])`.
**P1 by the panics rule**, and fixed here.

## The fix

`TableAccumulator::expecting_key` exposes what the accumulator already knows
— whether the next scalar lands in key position — so the three emitters whose
keys must be text reject it through the `unacceptable` hook added for
eu-1tkk.7.24. The three panic sites become fallbacks that keep the key's own
rendering, unreachable behind the guard.

```
error[EU-RENDER-SHAPE]: cannot render this document as JSON: a block key is the number 1, but JSON object keys must be text
  = give the block symbol keys, e.g. 'kv-block(:one, "a")', or convert the key to text with 'str' before rendering
  = 'yaml' and 'edn' output both support keys that are not text, and render this block unchanged
```

A **list or block used as a key** is covered too — it arrives as a sequence
or block start rather than a scalar, so `as_key` saw a `Value::Array` and
panicked just the same.

**Reported rather than coerced to `"1"`**, for the eu-1tkk.7.28 reason: a
string key is not the key that was evaluated, and the difference survives
into anything re-reading the output. Worth noting the type checker already
emits `type mismatch calling 'kv-block'` for these, so this confirms a
mistake it has flagged rather than refusing something idiomatic.

**yaml and edn are deliberately unchanged** — both genuinely represent such
keys (`1: a`, `{1 "a"}`). Fixture 193 pins that, so hardening the other three
cannot quietly take these with it.

## Fault injection

| Injection | Result |
|---|---|
| json guard off **+ original `unwrap` restored** | 198 **FAIL — exit 101** |
| toml guard off **+ original `unwrap` restored** | 199 **FAIL — exit 101** |
| eu guard off **+ original `panic!` restored** | 200 **FAIL — exit 101** |
| all text keys rejected (over-rejection) | 193 **FAIL** |
| only `Str` keys rejected | 193 **FAIL** |

All restored and re-verified green.

**Two of my injections were no-ops and I nearly counted them.** Deleting
`Primitive::Sym` or `Primitive::Str` from the match arm does not reject those
keys — they fall through to the catch-all `_ => return None`, so the guard
still accepted them and 193 still passed. Rejecting them *explicitly*
instead makes 193 fail. Same shape as the last two rounds: the injection
looked like it proved the test had teeth when it proved nothing.

That correction also settled a real question: the `Str` arm is **live, not
dead code**. `kv-block("a", 1)` emits a `Primitive::Str` key, distinct from
the symbol keys a literal block produces, and rejecting only `Str` fails 193.
Fixture 193 now asserts both accepting arms.

Note the guard-off-only injections give **exit 0**, not 101 — because the
panic sites are already replaced by fallbacks. Restoring the original
`unwrap`/`panic!` as well is what reproduces the pre-fix abort, and is what
proves these fixtures would have caught the bug.

## Validation

| prelude | engine | result |
|---|---|---|
| source | bytecode | all green (516 harness) |
| blob | HeapSyn | same 3 |

Clean under `EU_GC_VERIFY=2 EU_GC_STRESS=1`. `cargo fmt --all --check` and
`cargo clippy --workspace --all-targets -- -D warnings` clean, no `#[allow]`.

## A separate bug found while writing the fixture — eu-1tkk.7.31 (P1)

A c-string containing doubled braces makes the **whole declaration vanish**,
silently, exit 0:

```
$ printf 'a: c"{{X}}"\n' > f.eu && eu f.eu
---
{}
```

The plain-string form `"{{X}}"` correctly gives `{X}`, and
`docs/guide/string-interpolation.md` documents `{{`/`}}` as *the* literal
brace escape while presenting c-strings as the C-escape variant — so a
reader has every reason to expect this to work. Silent loss of a declaration
is the class this cycle keeps finding, so I filed it as P1 rather than
folding it in here. Fixture 193 uses the `\u007b`/`\u007d` escapes as the workaround
and says why.

## Fixture numbers taken

`tests/harness/`: **193**. `tests/harness/errors/`: **198**, **199**, **200**.


gmorpheme:

`7c0b07f8` (which now carries #1092), with `lib/prelude.blob` regenerated after
the branch switch and `prelude_blob_ok` confirmed from the build-script output.

I have **retargeted this PR at `master`** — it was opened against
`fix/furnace-broken-pipe-io-errors`, the last link in the
#1089 → #1091 → #1092 → #1093 stack. Its predecessors are on master now, so
the diff here is its own two commits alone.

### The fix

Three `AsKey::as_key` implementations aborted the process on a key that was
not text: `panic!` in `eu.rs`, `as_str().unwrap()` in `json.rs` and `toml.rs`.
Reachable from ordinary source — `kv-block(1, "a")` — so this is user-facing,
not an internal invariant.

The shape is now familiar from #1089 and #1091, and the consistency is the
point: a **shared** `unrepresentable_key` helper in `export/mod.rs` behind the
same `unacceptable(event)` hook, with `TableAccumulator::expecting_key()`
supplying the one bit of state the emitters lack. Three formats, one predicate,
one message. Had each emitter grown its own key check the three would have
drifted, and the earlier PRs in this stack established exactly why that
matters.

`expecting_key()` reading `matches!(self.stack.last(),
Some(Expectation::EvenBlockAccumulation(_, _)))` is precise: even accumulation
is the key position, odd is the value position, so a non-text **value** is
untouched. That asymmetry is the whole correctness of the check and it is one
line.

### The decision to reject rather than coerce

This is the judgement call in the PR and I think it is right.

Coercing `1` to `"1"` would make the export succeed, and the doc comment gives
the reason not to: *a string key is not the key that was evaluated, and the
difference survives into anything that re-reads the output.* That is the same
argument the TOML-null fix made in #1091, applied consistently — and consistency
here is worth more than the individual call, because a user who learns that
eucalypt reports rather than silently converts can trust the other formats too.

The observation that **the type checker already warns on such a key** is what
settles it: this is confirming a mistake the toolchain has flagged, not
refusing something idiomatic.

The remediation notes are concrete and correct — symbol keys via
`kv-block(:one, "a")`, or `str` to convert — and they name `yaml` and `edn` as
formats that carry non-text keys **and render this block unchanged**. That last
clause matters: it is only true because YAML has complex keys and EDN has
arbitrary keys, and fixture 217 asserts it rather than leaving it as a claim in
an error message.

### Fixture 217 is the half that makes this a specification

The three error fixtures say what is rejected. 217 says what is **accepted**,
in the formats that can carry it. Without it the PR would read as "non-text
keys are bad"; with it, it reads as "non-text keys are a property of the
*format*, not of the value" — which is the true statement, and the one a user
needs to act on.

### The fallbacks

Each `as_key` keeps a non-panicking fallback, documented as unreachable because
`unacceptable` rejects first: `self.to_string()` for JSON and TOML, and the
value's own eu rendering for `eu.rs`. Same judgement as #1089 and #1091 and
correct for the same reason — replacing a panic with a lossless rendering is
better even in a branch believed unreachable, because "unreachable" is a claim
about today's call graph and this is the third PR in a fortnight to discover a
route nobody expected.

`render_eu_value(other).unwrap_or_else(|_| "?".to_string())` is the only one
that can lose information, and only in a doubly-unreachable case.

Making `unrepresentable_key` never reject (`if !expecting_key || true`), in my
own worktree:

```
test test_error_218 ... FAILED
test test_error_219 ... FAILED
test test_error_220 ... FAILED
test result: FAILED. 0 passed; 3 failed
```

Restored, all four fixtures (217, 218, 219, 220) pass. All three error
fixtures gate, and each covers a different one of the three former panic
sites — so no third of this fix can regress with the suite green, which is the
same per-site standard #1077 met.

### Fixture renumbering applied

- `193_non_text_block_keys` → **217** (193 is
  `errors/193_1tkk_7_12_curated_trace`)
- `errors/198_json_block_key` → **218**
- `errors/199_toml_block_key` → **219**
- `errors/200_eu_block_key` → **220**

198-200 collided with the SV3 fixtures in `tests/harness/` — the two
directories share one number sequence. Test functions renamed to match
(`test_error_198/199/200` → `test_error_218/219/220`), the cross-reference in
`test_error_219`'s doc comment updated, and the fixture header's "see
errors/198-200" corrected to 218-220. Nothing in the tree still refers to the
old numbers.

### Gates

- `cargo fmt --all` — clean

PR #1094, opened 2026-07-27
fix(diagnostics): engine-aware invariant gate + ratcheted blob-mode CI job (eu-1tkk.7.17, eu-0cc1)

gmorpheme:
Three defects in the diagnostics invariant gate and the CI job that runs it, all the same class: a gate reporting green while asserting less than it claims.

## eu-0cc1 — `TRACE_BUDGET` was restated, not imported

`src/eval/error.rs` documents its constant as "shared by the curated human/JSON trace and the objective invariant gate … centralised here so the two cannot drift apart". `tests/diagnostics_invariants.rs:8` then declared its own `const TRACE_BUDGET: usize = 12`, which made that sentence false. Now imported.

## eu-1tkk.7.17 — the bead's symptom is gone; the real defect is worse

The bead says `metadata_span.eu` and `hof_bad_arg.eu` are marked `xfail` and trip under `EU_HEAPSYN=1`. Neither is marked any more — **no corpus fixture is `xfail` at all today**. So the "you fixed it, lock it in" arm was already vacuous on *both* engines, and `assert!(heapsyn || unexpected_pass.is_empty())` was relaxing an assertion that could never fire. On top of that, CI never ran this target under `EU_HEAPSYN=1` in the first place, so nothing would have noticed either way.

Replaced with per-engine markers — `xfail_engines = ["bytecode"]` — so a fixture broken on one engine stays a live guard on the other. The blanket HeapSyn relaxation is deleted, and a new step in the *HeapSyn harness (baseline)* job runs the gate under `EU_HEAPSYN=1`. I will rewrite the bead to describe the defect that actually exists.

## eu-oxtcq mechanism 3 — the blob-mode job was an allowlist with a comment where enforcement should be

The job ran `cargo test --test diagnostics_invariants corpus_satisfies_invariants -- --exact`, beside a comment reading "Widen the `cargo test` invocations below as each lands". Nothing enforced that. Two tests were excluded by omission and any new test in either target would have been excluded the same way. And a `--exact` filter matching nothing exits 0:

```
$ cargo test --test diagnostics_invariants no_such_test -- --exact
test result: ok. 0 passed; 0 failed; 3 filtered out
$ echo $?
0
```

so a rename would have deleted a gate in silence.

`scripts/diagnostics-gate.sh` runs each target **whole** with a ratcheted exclusion list, then requires every excluded test to **still fail**. It also fails when a target contributes zero tests.

**Merge-order note, deliberate and load-bearing.** Once #1079 (eu-7x0r) lands, the two excluded tests start passing and this job goes **RED** until their two lines are deleted from the workflow. That is the enforcement working — it is how "widen as each lands" stops being a hope. The fix is to delete the lines. Do not loosen the ratchet, do not re-add an allowlist.

The stale claim in that comment is gone too: it asserted typecheck fixtures 010/011/013/063/085 were "actively red under blob mode (eu-rqwh, in progress)". eu-rqwh closed on 2026-07-24. I ran the full `cargo test --test harness_test` with a blob present: all five pass, and the only failure in 506 tests was `test_193_1tkk_7_12_curated_trace`, itself eu-7x0r.

## Two further vacuity guards inside the gate

* `parse_meta` rejects an unrecognised sidecar key rather than skipping it. The old parser ignored anything it did not recognise, so `xfail_engine`, `xfails` or the retired `xfail` would have been a marker that silently did nothing.
* The verdict is a pure function with a unit test for each of its four cases, so the `UnexpectedPass` arm is provably reachable whatever the checked-in corpus happens to contain (eu-oxtcq mechanism 9).

## Fault injection — per site, rebuilt each time

| Site | Injection | Result |
|---|---|---|
| `TRACE_BUDGET` import | crate constant → 0, renderers pinned to 12 (simulates the drift) | **RED**: `(v) trace length 1 > budget 0`. Reverting only the import, same injection → **GREEN**. Both directions proven. |
| per-engine `xfail` | `xfail_engines = ["heapsyn"]` on a fixture that passes | **RED** under `EU_HEAPSYN=1`, **GREEN** under the default engine. The old code relaxed precisely this away. |
| strict sidecar parse | append `xfail = true` to a real sidecar | **RED**: `unrecognised sidecar key "xfail"` |
| zero-fixture sweep | corpus extension filter → `"zzz"` | **RED**: `swept no corpus fixtures` |
| ratchet, fixed test | add a passing test to the exclusion list (simulates eu-7x0r landing) | **exit 1**, naming it and the bead |
| ratchet, renamed test | list a name that does not exist | **exit 1**, same message |
| zero-test target | `rm lib/prelude.blob && touch build.rs`, rebuild, build-script output confirmed `cargo:rustc-cfg=prelude_blob_stale` | `diagnostics_blame_plumbing_test` compiles to **0 tests**; script **exit 1**. The previous `cargo test --test …` exited **0**. |

## Verification

`cargo fmt --all`, `cargo clippy --workspace --all-targets -- -D warnings` clean. The gate is green in all four configurations — blob and blob-less × default engine and `EU_HEAPSYN=1`. Baseline of exactly three pre-existing blob-mode failures on master (`secondary_labels_do_not_excerpt_library_internals`, `debug_trace_restores_the_uncurated_trace`, `test_193_1tkk_7_12_curated_trace`, all eu-7x0r) confirmed and unchanged.

Part of the eu-oxtcq audit. Beads: eu-1tkk.7.17, eu-0cc1.


gmorpheme:
## Gatekeeper review — approved with one change applied, merging on green

`c21e6d72`, with `lib/prelude.blob` regenerated after the branch switch and

### The change applied: the ratchet fired, so I emptied the list

This branch was written while eu-7x0r was open, and its exclusion list named
the two tests it made fail:

```
scripts/diagnostics-gate.sh diagnostics_invariants \
  secondary_labels_do_not_excerpt_library_internals:eu-7x0r \
  debug_trace_restores_the_uncurated_trace:eu-7x0r
```

**eu-7x0r landed as #1079, and both tests now pass under a blob.** The script
requires every named entry to *still fail*, so carrying those two lines would
have turned this job red the moment it merged.

I deleted them, which is what the script's own comment instructs — "the fix is
to remove the line, never to loosen the check". The exclusion list is now
**empty**, and that emptiness is the mechanism's first live result rather than
a weakening of it.

Worth dwelling on, because it is the point of the PR: this branch would have
gone red on merge, and it would have gone red *for the right reason* — a
recorded known-failure got fixed and nobody updated the list. That is exactly
the class of drift the old allowlist could not detect, demonstrated on the
first PR to carry the new mechanism.

I also routed `tests/blob_deprecation_test.rs` — added to this job by #1074
while this branch was open — through the same script rather than leaving it as
a bare `cargo test`, so it gets the whole-target treatment and the
zero-tests-is-red check too. Leaving it outside would have recreated in
miniature the exact asymmetry this PR exists to remove.

### The mechanism

The old job named the single test it wanted:

```
cargo test --test diagnostics_invariants corpus_satisfies_invariants -- --exact
```

Three separate ways that fails silently, and the PR identifies all three:

1. A test **added** to the target is excluded by omission. The comment beside
   it said "widen as each lands"; nothing enforced that, and nothing did.
2. Two tests were already excluded by omission — not by decision.
3. **A filter that matches nothing exits 0.** So renaming a test removes it
   from CI while the job stays green. This is the sharpest of the three,
   because a rename is a routine, well-intentioned edit.

`scripts/diagnostics-gate.sh` inverts the default: run the whole target with
`--skip` per known failure, then run each known failure **alone** and require
it to still fail. New tests are included automatically; a fixed known failure
turns the job red until its line is deleted.

The zero-tests guard is the part I would not have thought to add and is the
better half of the script:

```sh
if ! echo "$output" | grep -qE 'test result: ok\. [1-9][0-9]* passed'; then
```

A `cfg`-gated file that compiles to zero tests reports `ok. 0 passed` and exits
0. `blob_deprecation_test` is `#![cfg(prelude_blob_ok)]`, so in a build where
the blob generation silently failed, the whole file would vanish and the job
would stay green. The `[1-9][0-9]*` pattern is doing real work.

### I verified the ratchet actually gates, rather than reading it

Empty list, all three targets:

```
diagnostics_blame_plumbing_test  exit=0
diagnostics_invariants           exit=0
blob_deprecation_test            exit=0
```

Then I named a **passing** test as a known failure, which is precisely the
situation the ratchet exists to catch:

```
$ ./scripts/diagnostics-gate.sh diagnostics_invariants corpus_satisfies_invariants:eu-fake
::error::diagnostics_invariants::corpus_satisfies_invariants now PASSES (or no longer exists) under this configuration.
::error::Remove it from the exclusion list in .github/workflows/build-rust.yaml so the gate guards it — eu-fake.
$ echo $?
1
```

Exit **1**. The ratchet is real, not decorative — which matters more than usual
here, because a gate mechanism that does not itself gate is worse than none: it
buys confidence without providing it.

`diagnostics_invariants` now reports **12 passed** under a blob, including
`debug_trace_restores_the_uncurated_trace`, which independently confirms the
exclusion removal was correct rather than optimistic.

### Shell details

`set -uo pipefail` with `"${skips[@]+"${skips[@]}"}"` — the expansion that
keeps an empty array from tripping `set -u` on older bash. Given the list is
now empty, that is load-bearing today rather than defensive, and I confirmed it
by running the script with no entries.

Deliberately **not** `set -e`: the script needs to inspect a non-zero `cargo
test` status rather than die on it, and it captures `status=$?` explicitly. The
right call, and not an oversight.

`--include-ignored` on both invocations means an `#[ignore]`d test cannot hide
from the gate either — a fourth escape route, closed without comment.

The script is checked in `100755`, so CI can execute it.

### One forward reference, correct but worth noting

The zero-tests comment points at `tests/gate_liveness.rs` "for the general form
of this check", which arrives in **#1095**. Harmless as a comment, and the two
PRs are adjacent in the queue — but if #1095 were dropped the reference would
dangle. Flagging so it is a decision rather than an accident.

### Documentation

The corrected comment block is unusually good and I have kept it intact: it
records what the old mechanism was, why it failed, what replaced it, and — in
the parenthetical — that an *earlier revision of the comment itself* made a
false claim about eu-rqwh which has since been corrected and verified. A
comment that documents its own past inaccuracy is more trustworthy than one
that quietly drops it, and given this queue has already turned up four stale
mechanism claims, that habit is worth encouraging.

### Gates

- `cargo fmt --all` — clean
- `scripts/diagnostics-gate.sh` — all three targets pass, ratchet verified to
  exit 1 on a fixed known failure

PR #1095, opened 2026-07-27
test(ci): fail loudly when a cfg-gated test file contributes zero tests (eu-oxtcq)

gmorpheme:
Item A of the eu-oxtcq audit, and the one with the widest reach.

## The failure mode

`build.rs` emits `prelude_blob_ok` only when `lib/prelude.blob` is present and hashes against the current `lib/prelude.eu`. A test file gated on that cfg therefore compiles to *nothing at all* without a blob — and a test binary containing no tests reports

```
test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

and exits 0. **Absent coverage is indistinguishable from passing coverage.** Measured on master, blob-less: `tests/diagnostics_blame_plumbing_test.rs` contributes 0 tests and `cargo test` is green. That grows to three files as #1074 and #1085 land.

This is survivable while a CI job reliably supplies the cfg. It stops being survivable the moment the `cargo xtask prelude-compile` step in that job is reordered, renamed or dropped: the gated assertions vanish and nothing anywhere goes red.

## What `tests/gate_liveness.rs` enforces

1. **Registration.** Every conditionally compiled file under `tests/` must appear in `REGISTRY`. This is the part a one-off audit cannot give you — the audit catches today's gates, the registry catches tomorrow's. Four files are registered, each with the cfg it needs, the supplier of that cfg, and a stated reason.
2. **No rot.** A registered file that has gone, or that no longer carries the cfg it is registered for, fails. A registry describing a world that no longer exists is worse than no registry.
3. **A supplier that actually supplies.** For every blob-gated file, the CI job named in its entry must exist in `.github/workflows/build-rust.yaml` **and** must run `cargo xtask prelude-compile`. This is the check that turns a dropped build step from silence into a red build.
4. **An explicit demand.** `EU_REQUIRE_PRELUDE_BLOB=1` asserts the cfg is genuinely on, for a job that prefers to state the requirement at runtime.

**No workflow change.** Check (3) is deliberately *static* — it reads the workflow file as data — so it needs no blob, no environment variable and no CI wiring. It runs in every configuration, including the deliberately blob-less "Test Suite" job, which is exactly where you want a dropped-step detector to live. It also means this PR does not touch `build-rust.yaml` and so does not conflict with #1094.

## Fault injection — per site, clean tree each time, reverted and confirmed clean after

| Injection | Result |
|---|---|
| Append `#[cfg(prelude_blob_ok)] #[test] fn injected() {}` to `tests/git_import_test.rs` | **exit 101** — `these test files are conditionally compiled but are not in REGISTRY` |
| Rewrite the gate in `diagnostics_blame_plumbing_test.rs` to `cfg(all())` | **exit 101** — `REGISTRY says … is gated on cfg(prelude_blob_ok), but its source does not contain that predicate` |
| **Delete the "Generate prelude blob" step from the `diagnostics-blob-mode` job** | **exit 101** — `CI job diagnostics-blob-mode is the supplier for … but it does not run cargo xtask prelude-compile. Without that step build.rs sets prelude_blob_stale, the gated file compiles to ZERO tests, and the job reports green having asserted nothing` |
| Rename the `diagnostics-blob-mode` job | **exit 101** — `there is no such job in …/build-rust.yaml` |
| `EU_REQUIRE_PRELUDE_BLOB=1` on a blob-less build | **exit 101** — names `prelude_blob_stale` and the remedy |

The third row is the one the audit was after.

## Verification

Green in both prelude modes — blob-less (`cargo:rustc-cfg=prelude_blob_stale`) and with a blob (`cargo:rustc-cfg=prelude_blob_ok`), the verdict read from the build-script output at `target/debug/build/eucalypt-*/output`, never inferred from the blob file existing. `cargo fmt --all` and `cargo clippy --workspace --all-targets -- -D warnings` clean.

Note for anyone reproducing: restoring a blob with `mv` or `cp -p` preserves its mtime, so cargo does not rerun `build.rs` and the stale cfg persists — `touch build.rs` after restoring. That warning is recorded in the file's doc comment.

## Not in this PR

`tests/gate_liveness.rs` is itself never conditionally compiled — it is the thing that detects conditional compilation — and is excluded from its own scan. If it ever did compile to zero tests, the zero-test check in `scripts/diagnostics-gate.sh` (#1094) would catch it.

Bead: eu-oxtcq (item A).


gmorpheme:
## Gatekeeper review — approved with three registry entries added, merging on green

`88ea0fff`, with `lib/prelude.blob` regenerated after the branch switch and

### It failed the moment I merged master into it, and that is the review

```
these test files are conditionally compiled but are not in REGISTRY in tests/gate_liveness.rs:
  tests/blob_deprecation_test.rs
  tests/harness_test.rs
  tests/wire_format_enforcement_test.rs
```

All three gained their gates *after* this branch was cut —
`blob_deprecation_test.rs` from #1074, `wire_format_enforcement_test.rs` from
#1082, and `harness_test.rs`'s blob-core-versus-source comparison from
eu-r4647. **Three real omissions caught on the mechanism's first contact with
the tree it is meant to guard.**

I have written the three entries, each naming the cfg, the supplying CI job
and — the part the registry actually earns its keep on — *what the tests assert
that is meaningless without the gate*. `harness_test.rs` and
`wire_format_enforcement_test.rs` are the interesting pair: both are mostly
**ungated** files with a single gated test inside, so neither would vanish
wholesale. The rationales say so explicitly, because a future reader seeing
those paths in a registry of "conditionally compiled files" would otherwise
reasonably conclude the whole file disappears without a blob.

This is the second PR in this queue whose own mechanism failed on arrival — the
other being #1094's ratchet, which had recorded eu-7x0r failures that #1079 had
since fixed. In neither case is that a defect. It is the difference between a
gate and a comment.

### The failure mode being closed

Stated better in this file's header than I could:

> a test binary containing no tests reports `ok. 0 passed … 0 filtered out`
> and exits 0. **Absent coverage is indistinguishable from passing coverage.**
> Measured on master: with no blob present,
> `tests/diagnostics_blame_plumbing_test.rs` contributes 0 tests and
> `cargo test` is green.

And the reason it matters is the sentence after: it is survivable while a CI job
reliably supplies the cfg, and stops being survivable the moment someone
reorders, renames or drops the `cargo xtask prelude-compile` step — at which
point the assertions vanish and *nothing anywhere goes red*.

I hit precisely this failure mode myself earlier in this queue: a stale blob
produced a false red that mimicked a genuine aliasing regression. The inverse —
a silent green — is worse, because nothing prompts you to look.

### The four checks, and why the third is the good one

1. **Registration** — a new gate cannot be added without a reviewer seeing an
   entry appear. The header makes the right distinction: an audit catches
   today's gates, a registry catches tomorrow's.
2. **No rot** — a registered file that no longer exists, or no longer carries
   the cfg it is registered for, fails. The registry cannot drift into
   describing a world that no longer exists.
3. **A supplier that actually supplies** — for every blob-gated file, the named
   job must exist in `.github/workflows/build-rust.yaml` **and** must run
   `cargo xtask prelude-compile`. This is the one that turns a dropped build
   step from silence into a red build, and it is **static**: no blob, no
   environment variable, no CI wiring, so it runs in every configuration
   *including the deliberately blob-less "Test Suite" job*. A check that only
   worked where the blob exists would be useless precisely where it is needed.
4. **An explicit demand on request** — `EU_REQUIRE_PRELUDE_BLOB=1` for a job
   that wants to state the requirement at runtime.

### Details worth noting

`gate_markers()` is built by `concat!("#!", "[cfg(")` **so that this file does
not match its own scan through its own doc comments**. That is a real problem
(the header quotes the marker several times) solved in one line, and a reader
who "tidied" it into a literal would break the file silently.

The file is deliberately never itself conditionally compiled — it is the thing
that detects conditional compilation, so gating it would be self-defeating —
and the header says so, along with the backstop:
`scripts/diagnostics-gate.sh`'s zero-test check would catch it if it ever did.
Two mechanisms covering each other's blind spot, from #1094 and #1095
respectively, which is why the pair belongs together.

Tracking `cfg(prelude_blob_ok)` and `cfg(prelude_blob_stale)` **wherever they
appear** — not just at crate level — is what makes `harness_test.rs` and
`wire_format_enforcement_test.rs` show up at all: a gated test inside an
ungated file disappears just as quietly as a whole target, and only a marker
scan finds it.

Deleting the `blob_deprecation_test.rs` registry entry I had just added:

```
test every_conditionally_compiled_test_file_is_registered ... FAILED
these test files are conditionally compiled but are not in REGISTRY in tests/gate_liveness.rs:
  tests/blob_deprecation_test.rs
test result: FAILED. 8 passed; 1 failed
```

Exactly one test fails and it names exactly the removed file. Restored: 9/9.

Note the other eight tests correctly still pass — a registry omission is not a
rot failure and not a supplier failure, so a test that also failed would have
been testing something other than what it claims.

### Gates

- `cargo fmt --all` — clean
- `cargo test` — **0 failed across every test binary**: lib 1404, harness 537,
  doc examples 111, gate_liveness 9

PR #1098, opened 2026-07-31
fix(vm): force Meta body through thunk-forcing ceremony (eu-bc34x)

gmorpheme:
## Summary

A metadata-annotated binding (`` ` :suppress``, a doc string, etc.) compiles to `Meta{meta, body}`. `return_meta`'s "other" branch resolved `body` with a bare, passive env-slot read instead of the standard black-hole-then-`Update` ceremony every other thunk reference uses — so if `body` named a genuinely unevaluated computation, it silently never memoised: every reader re-ran it in full (multiplicative instead of additive cost).

Two candidates named in the bead were ruled out first: `is_settled`/`create_arg_array` only affects call-argument aliasing (irrelevant here); widening `is_whnf()` alone changes nothing, since `Continuation::Update`'s existing branch correctly writes the original metadata-tagged form back (preserving `` ` ``/`meta` visibility), not a resolved value.

**Fix**: route `body` through the same enter/black-hole/`Update`-push logic as `Atom{Ref::L}`, in both engines. Pure runtime change — blob bytes byte-identical before/after.

A review pass caught an ordering bug in the first version (the outer continuation must go back on the stack *before* `enter_meta_body` runs, not after, or its `Update` push lands in the wrong position and the outer consumer's result overwrites the memoised slot). Neither symptom this exposed was a pre-existing bug — both were this fix's own wiring; `hoist.rs` was investigated and exonerated (eu-096pd, closed invalid).

## Verification

- `tests/harness/errors/*.eu` diff: **0/194 differ**, blob × source × three dispatch engines
- `cargo test --release`: full suite green, both engine defaults, 539 harness tests, all 8 canonical benches
- Ticks match the additive baseline within ~60-tick noise everywhere (was 206,212–207,977 multiplicative)
- `EU_GC_VERIFY=2 EU_GC_STRESS=1`: clean
- New regression test (`tests/harness/228_bc34x_meta_body_memoisation.eu` + `test_228_bc34x_meta_body_memoisation`), 30s deadline gate, **fault-injection verified twice** (the ordering bug, and the full pre-fix binary — both FAIL correctly, both restored and re-confirmed PASS)

Full measurements: `docs/superpowers/reports/2026-07-31-eu-bc34x-meta-memoisation-fix.md`

## Test plan

- [x] `cargo test --release` (default + `EU_HEAPSYN=1`)
- [x] `cargo clippy --workspace --all-targets -- -D warnings`
- [x] `cargo fmt --all`
- [x] `tests/harness/errors/*.eu` before/after diff, all mode/engine combos
- [x] `EU_GC_VERIFY=2 EU_GC_STRESS=1`
- [x] Fault injection (both known failure modes)

**Do not merge — owner review** (touches GC/VM machinery, per CLAUDE.md's recorded-review requirement).

### DIFF
--- a/docs/superpowers/reports/2026-07-31-eu-bc34x-meta-memoisation-fix.md
+++ b/docs/superpowers/reports/2026-07-31-eu-bc34x-meta-memoisation-fix.md
@@ -1,0 +1,114 @@
+# eu-bc34x: metadata-annotated bindings lose thunk memoisation — root cause and fix
+
+Furnace, 2026-07-31.
+
+## Root cause
+
+A binding carrying metadata (a doc string, `:suppress`, etc.) compiles to
+`Meta{meta, body}`. `is_whnf()` treats `Meta` as unconditionally WHNF, so the
+binding is a non-updateable Value form — cheap to re-enter, by design.
+`return_meta`'s "other" branch (any consumer that isn't itself destructuring
+the metadata) used to resolve `body` with a bare, passive environment-slot
+read (`HeapNavigator::resolve` / `resolve_ref`) instead of entering it
+through the standard black-hole-then-`Update` ceremony every other reference
+to an updateable thunk goes through. If `body` names a genuinely unevaluated
+thunk, that thunk's own persistent slot never got an `Update` continuation
+registered, so it never memoised: every reader independently re-ran the full
+computation.
+
+Two other candidates named in the bead were ruled out:
+- **is_settled/create_arg_array**: only controls whether a fresh alias
+  wrapper is built when a slot is passed as a call argument; irrelevant here
+  since an alias onto the Meta wrapper is harmless either way.
+- **is_whnf() alone**: even if the Meta wrapper were compiled as an
+  updateable Thunk, `Continuation::Update`'s existing (unmodified) branch
+  writes the *original, unresolved* Meta form back into the binding's own
+  slot (correct — preserves `` ` ``/`meta` visibility), so widening
+  `is_whnf()` changes nothing about whether the underlying computation
+  memoises.
+
+## Fix
+
+Route `body`'s resolution through the same enter/black-hole/`Update`-push
+logic as `Atom{Ref::L}` (`enter_local`/`enter_global` in
+`src/eval/bytecode/machine.rs`; a new `enter_meta_body` helper mirroring
+them in `src/eval/machine/vm.rs`), leaving the Meta node's own slot,
+`is_whnf()`, and `is_settled`/`create_arg_array` untouched.
+
+## A second, ordering bug found during review
+
+The first version of this fix called `enter_meta_body` *before* pushing the
+outer continuation back onto the stack. If `body` names an updateable
+thunk, `enter_meta_body` pushes its own `Update` continuation for that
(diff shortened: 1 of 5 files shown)


gmorpheme:
## Wicket review — merged

Reviewed in isolation at `/tmp/eu-wicket-1098` (head `42b960a4`), plus a baseline `origin/master` (`642b5e2d`) build at `/tmp/eu-wicket-1098-baseline` for independent corpus comparison.

**Scope check.** `git diff origin/master...HEAD --stat` (not the stale local `master` ref, which was behind) shows exactly 5 files: `src/eval/machine/vm.rs`, `src/eval/bytecode/machine.rs`, the new `tests/harness/228_bc34x_meta_body_memoisation.eu` fixture, its `harness_test.rs` wrapper, and the fix report. Matches the stated "one line per engine plus a new helper" shape.

**Code review.** `enter_meta_body` in both engines mirrors the existing `Atom{Ref::L}`/`enter_local`/`enter_global` ceremony exactly — the bytecode version reuses the pre-existing `enter_local`/`enter_global` helpers verbatim (no duplication); the HeapSyn version in `vm.rs` reimplements the same black-hole+`Update` sequence inline, and I compared it line-by-line against the `Atom{Ref::L}` arm in `handle_instruction` — matches, including capturing `self.closure.env()` *before* reassignment for the same reason `handle_instruction` does. `Ref::V` correctly keeps the old passive-resolve path (trivially WHNF, no memoisation concern).

**Fault injection — independently repeated, not taken on trust:**
- Reintroduced the ordering bug (push `other` after `enter_meta_body` instead of before) in both engines → `test_228_bc34x_meta_body_memoisation` FAILs in **both** the default bytecode engine and under `EU_HEAPSYN=1` (a `LookupOnFunction`/similar wrong-value error, not a timeout — matches the described symptom). Reverted, rebuilt → PASSes in both.
- Fully reverted `vm.rs`'s `enter_meta_body` to the old bare `resolve` (HeapSyn only, since bytecode has its own separate `return_meta`) → confirmed timeout past 20s (was going to be >40s per the report). Reverted, rebuilt → PASSes.

**Corpus-diff claim, independently reproduced.** Ran all 194 `tests/harness/errors/*.eu` fixtures directly against both binaries (fixed vs. `origin/master` baseline) with matched relative paths: **0/194 differ**, default engine. Also spot-checked the 20 fixtures that actually carry backtick metadata annotations under `EU_HEAPSYN=1`: **0/20 differ**.

**Gates:** CI terminal and green (19 pass, 5 skipped — release/prepare-release jobs, expected off-release-branch). `cargo fmt --all --check` and `cargo clippy --workspace --all-targets -- -D warnings` clean. `cargo test --release test_error` (179 tests) green. Read `eu-bc34x` and closed-invalid `eu-096pd` — the bead's own root-cause note matches this PR's account exactly; nothing to reopen.

**Owner-hold note:** the PR body's "Do not merge — owner review" line was the author self-applying CLAUDE.md's GC/VM-touching filter. The owner has released it for this cycle under "bug fixes proceed; intentional semantic changes stop" — this is a one-line-per-engine bug fix with no intended semantic change, confirmed by the corpus-diff and tick-parity evidence above. Merging as gatekeeper.

PR #1100, opened 2026-07-31
fix(diag): target_annotation looks through Let/LetRec (eu-7fjiq); narrow DIVERGENCE.md engine claim

gmorpheme:
## Mode

DIRECTED (owner-sanctioned brief relayed by coordinator) — reviewed and merged by wicket in the normal way, not the owner personally.

## Summary

- **eu-7fjiq**: `target_annotation` (both `vm.rs`/HeapSyn and `bytecode/machine.rs`, byte-dispatch + pre-decode) now looks through a leading `Let`/`LetRec` to find the thunk's `Ann`, recursing for repeated inlining. The inliner's argument-sharing `Let` (eu-gua64) can compile `Ann(App)` as `Let(..., Ann(App))`, which hid the annotation from the old single-node peek.
- **eu-l51r7 (DIVERGENCE.md)**: added a static "Engine scope" section (generated via `snapshot_engine.rs`, kept in sync with `divergence_inventory_is_current`) stating the blob-vs-source table covers the default engine only, and recording the non-default-engine trace divergence.

## Full findings — see the linked report

`docs/superpowers/reports/2026-07-31-eu-7fjiq-let-annotation-lookthrough.md` has the complete picture; headline points:

- **error_176.eu's verification criterion is not met by this fix**, and I found out why: the missing secondary/trace there is `result`'s *own* Update continuation, whose thunk body (`Let(__shared_e=..., EXPECT(...))`) has **no `Ann` anywhere** because `ProtoLet` (compiler.rs) never reads Core `Let`'s own smid — a separate, compiler-side gap. I prototyped fixing it; it broke primary-location attribution broadly (unrelated outer `Let`s also have valid smids) and I reverted it. Also found `error_176`'s error goes through `evaluate_to_whnf`, which restores the caller's stack before propagating an error (pre-existing, tested behaviour) — a second, independent reason the frame doesn't surface there.
- **Verification**: 7 new Rust unit tests (not a `.eu` harness case — see report for why one couldn't be constructed) directly exercise `target_annotation` with `Let`/nested `Let(LetRec(Ann))` and a negative (no-`Ann`) case, in both engines/both dispatch paths. Fault-injected: reverting each `target_annotation` to the old single-node peek fails the corresponding tests; restoring passes.
- **Corpus movement: 0 of 213** snapshots changed (`cargo xtask diag-snapshot --bless`, fresh blob) — nothing to re-bless.
- **Bonus**: this fix moves `errors/085_destructure_short_list` from diverging to matching under `EU_HEAPSYN=1`, so eu-l51r7's HeapSyn divergence count is **7 → 6** (now identical to the `EU_PREDECODE=0` set). DIVERGENCE.md updated accordingly; recommend updating eu-l51r7 itself too.

## Gates

- `cargo fmt --all`, `cargo clippy --workspace --all-targets -- -D warnings`: clean
- `cargo test --release` (fresh blob): 23/23 suites green
- `EU_HEAPSYN=1 cargo test --release`: 6 known diagnostics-snapshot mismatches (eu-l51r7, now updated), everything else green
- `EU_PREDECODE=0 cargo test --release`: same 6, everything else green

### DIFF
--- a/docs/superpowers/reports/2026-07-31-eu-7fjiq-let-annotation-lookthrough.md
+++ b/docs/superpowers/reports/2026-07-31-eu-7fjiq-let-annotation-lookthrough.md
@@ -1,0 +1,146 @@
+# eu-7fjiq: target_annotation Let/LetRec look-through — findings and corpus movement
+
+2026-07-31, Clarion, DIRECTED mode. Branch `fix/clarion-eu-7fjiq-let-annotation`,
+PR to master, reviewed by wicket (not the owner personally — this is a directed
+brief, not a proactive Clarion PR).
+
+## The fix
+
+`Machine::target_annotation` (`src/eval/machine/vm.rs`, HeapSyn engine) and
+`bytecode::machine::target_annotation` (`src/eval/bytecode/machine.rs`, both
+byte-dispatch and pre-decode paths) resolved a thunk's fallback source
+annotation by peeking only the *single leading node* of its compiled body. The
+inliner's argument-sharing `Let` (eu-gua64, PR #1086) can compile `Ann(App)`
+as `Let(..., Ann(App))`, which hides the `Ann` from a single-node peek and
+silently drops the annotation.
+
+Both engines now walk through a leading `Let`/`LetRec` chain to find the
+`Ann`, recursing rather than peeking one level, because repeated inlining can
+stack more than one sharing wrapper. All three dispatch paths (HeapSyn,
+byte-dispatch, pre-decoded bytecode) needed the change — confirmed by
+reproducing the bug under all three before fixing (`eu --heap-limit-mib 4096
+--source-prelude` on the default build, plus `EU_HEAPSYN=1` and
+`EU_PREDECODE=0`).
+
+Verification: 7 new unit tests (3 in `vm.rs`, 4 in `bytecode/machine.rs`,
+covering both dispatch paths) construct a `Let`/nested `Let(LetRec(...))`
+wrapping an `Ann`, call `target_annotation` directly, and assert the smid is
+found. A fourth/eighth negative test asserts a `Let` with no `Ann` anywhere
+in its body still returns `Smid::default()` (no invented location). Fault
+injection: reverting each `target_annotation` to its old single-node-peek
+form makes the corresponding positive tests fail with `Smid(None)` instead of
+the expected smid; restoring the fix makes them pass again. Confirmed for
+both files.
+
+## error_176.eu's verification criterion is NOT met — and why
+
+The bead's own verification target — `tests/harness/errors/error_176.eu`
+under `--source-prelude` regaining the "called from here" secondary at 16:22
+and the "- result at ..." stack-trace line, while keeping the primary at
+19:32 — is **not achieved by this fix**, confirmed empirically (before/after
+binaries, byte-identical output on that fixture with or without the fix).
+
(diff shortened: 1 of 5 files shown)


gmorpheme:
## Wicket post-merge recorded review

The owner merged this directly while my review was still in flight, so this is a post-merge recorded review, not the pre-merge gate — nothing here was skipped on my side, master just moved first. Recording it because gate 9 (recorded-review for GC/VM-touching PRs) needs a review comment from someone other than the author to actually be on the PR, and this one had none.

**Scope reviewed**: `src/eval/machine/vm.rs` and `src/eval/bytecode/machine.rs` (both dispatch paths), head `5e0c57a2`.

### Per-site fault injection — independently repeated, each site alone

Reverted `vm.rs`'s `target_annotation` to the old single-node peek, leaving both `bytecode/machine.rs` arms untouched, rebuilt: only the two `vm.rs` tests failed (`Smid(None)` instead of the expected smid); both bytecode-side tests stayed green. Restored, rebuilt: 7/7 green.

Reverted the byte-dispatch arm alone: `target_annotation_looks_through_let_byte_dispatch` and the byte-dispatch half of the nested-`Let`/`LetRec` test failed; the pre-decoded test stayed green. Restored: 7/7 green.

Reverted the pre-decoded arm alone: the pre-decoded test and its half of the nested test failed; byte-dispatch stayed green. Restored: 7/7 green.

Each site is independently load-bearing — this isn't three tests passing off one shared code path.

### 085_destructure_short_list — verified against a from-scratch baseline, not CI

Built `origin/master` at `9855417a` unmodified in a separate worktree. Ran the full `tests/harness/errors/*.eu` corpus (197 files) against both binaries:
- Default engine: **0 files moved**.
- `EU_HEAPSYN=1`: **exactly 1 file moved** — `085_destructure_short_list` — and its output under the fix now byte-matches the default engine's output for that fixture (diffed directly, empty).
- `EU_PREDECODE=0`: **0 files moved**.

All three match the PR's claims precisely.

### My own attempt to find a reachable `.eu`-level fixture — failed, which is itself the useful result

I didn't take "no `.eu` case is reachable" on trust. I built a small program exercising eu-gua64's argument-sharing mechanism directly (`f(a,b): a+a+b` called with a non-atomic, error-producing recursive argument), confirmed via `eu dump inlined` that it does produce the `Let(__shared_a = ..., ...)` wrapper the bug is about, then diffed its diagnostic output before/after this fix under `--source-prelude`. **Byte-identical.** My attempt to demonstrate the gap end-to-end failed on a genuine try, which independently corroborates rather than merely accepts the author's account that every reachable scenario is covered by some other annotation mechanism, or blocked by the separate compiler-side gap (eu-5gk86).

### Recursion-depth / cycle safety

All three look-through walks (the `HeapSyn` body pointer in `vm.rs`, the byte-dispatch `body_off`, the pre-decoded body ordinal via `Instr::a`) follow exactly the same pointer/offset/ordinal the real interpreter already uses to advance execution when it actually runs a `Let`/`LetRec` (confirmed by reading the corresponding `Op::Let`/`Op::LetRec` dispatch arms). If that could cycle, ordinary program execution would already infinite-loop, which it demonstrably doesn't. No new termination risk.

### Gates

`cargo fmt --all --check`, `cargo clippy --workspace --all-targets -- -D warnings`, and a full `cargo test --release` (fresh blob, `rm -f lib/prelude.blob` first per the stale-blob hazard) all clean, independently run in an isolated worktree.

**Verdict**: would have merged this clean. Nothing turned up.

PR #1102, opened 2026-08-01
fix(diag): drop misleading panic prefix/note, add EU-EVAL-TYPE to UnexpectedFunction

gmorpheme:
## Summary

Three related fixes to error rendering (`src/eval/error.rs`), batched because they touch the same file. Beads: eu-1tkk.7.33, eu-1tkk.7.35, eu-1tkk.7.37.

### eu-1tkk.7.33 — drop the "panic:" prefix for library-raised errors
`UserPanic`'s "panic: " prefix is meant to show the user their own explicit `panic()` call, but the prelude raises its own errors through the same intrinsic (e.g. `nth`'s out-of-range check, `lib/prelude.eu:1435`). `to_diagnostic` now reuses `SourceMap::is_user_file` (already built for exactly this "prelude vs user" distinction) on the panic call site's Smid: the prefix shows only when the call site is in a user file. **Choice**: I picked the selective option (owner's decision left this open) rather than dropping the prefix everywhere, because the infrastructure to distinguish the two call paths already existed and the change was modest (~15 lines in `to_diagnostic`, no new plumbing). Verified both directions: `009_panic.eu`/`137_user_panic.eu` (direct user `panic()`) keep the prefix; `nth(10)`, `parse-args` misuse, `validate` bad-spec (all prelude-internal panics) lose it.

### eu-1tkk.7.35 — delete the misfiring "head and tail require a list" note
Removed the leading note from `HeadOfNonList`/`TailOfNonList` — it misfired in 4 of 6 carrying snapshots (reached via `nth`/`map`/`++`, not `head`/`tail` directly) and was tautological in the other 2. Kept the conditional value-shaped notes (string `++`/`str.letters` advice, "wrap it: `[n]`").

### eu-1tkk.7.37 — UnexpectedFunction gets EU-EVAL-TYPE
Added `UnexpectedFunction` to the code allow-list — same message family as `TypeMismatch` ("type mismatch: expected X, found Y"), just missing the code since eu-1tkk.7.9 introduced it.

**`ComparisonTypeMismatch` verdict: not added.** Its message ("cannot compare incompatible types with '...'") is a different shape, not "type mismatch: expected X, found Y". Notably `NoBranchForDataTag` shares TypeMismatch's *exact* message text and is still deliberately left uncoded, so extending to `ComparisonTypeMismatch` would be scope creep beyond this bead's narrow intent.

## Snapshot diff (reblessed via `cargo xtask diag-snapshot --bless`)
22 snapshots changed, all exactly as intended — reviewed the full diff line by line:
- 13 lose the `panic: ` prefix (`nth` OOB ×6, `parse-args` misuse ×3, `validate` bad-spec, `from-data` library-blame, curated-trace/anchor fixtures). `009_panic`/`137_user_panic` (genuine user `panic()`) are unaffected.
- 6 lose the leading head/tail note (`042_map_non_list`, `047_str_plus_plus`, `129_head_tail_not_list`, `138_internal_no_panic_prefix`, `181_direct_app_call_site`, `swap_args`); conditional notes remain.
- 3 gain `error[EU-EVAL-TYPE]`/`code: EU-EVAL-TYPE` (`078_anon_anaphor_scope`, `192_1tkk_7_9_function_not_block`, `catenation_arith`).

Also updated `tests/harness/errors/202_sv3_contract_bad_spec.eu.expect` and one doc example in `docs/guide/contracts.md` that hardcoded the old "panic: validate: ..." text.

## Fault injection
Reverted `src/eval/error.rs` to HEAD, reran `cargo test --test diagnostics_snapshots`: 22 fixtures failed with exactly the expected diffs (prefix/note/code reappearing or missing). Restored the fix, reran: all green. Confirms the diagnostics-snapshot suite genuinely gates these changes (the `.expect` sidecars use substring matching so don't gate on their own).

## Test plan
- [x] `cargo fmt --all --check` — clean
- [x] `cargo clippy --workspace --all-targets -- -D warnings` — clean, no `#[allow]`
- [x] `cargo test` (full suite) — green: 1427 lib tests, 547 harness tests, 17 diagnostics-snapshot tests, all others
- [x] Fault-injection verified (see above)

Not merging — needs a recorded review from a non-author (Wicket).

### DIFF
--- a/docs/guide/contracts.md
+++ b/docs/guide/contracts.md
@@ -254,7 +254,7 @@ file. `validate` raises when its first argument is not a type spec at all:
 ```eu,notest
 { import: "contract.eu" }
 r: { a: 1 } validate({ not: "a spec" })
-# error: panic: validate: not a type spec
+# error: validate: not a type spec
 ```
 
 That is a different error from a contract violation, distinguishable both
--- a/src/eval/error.rs
+++ b/src/eval/error.rs
@@ -891,10 +891,14 @@ pub enum ExecutionError {
     /// function so that the "panic:" label appears in the diagnostic.
     #[error("{1}")]
     Panic(Smid, String),
-    /// Error raised by the user's `panic("msg")` call.
+    /// Error raised by a call to the `panic("msg")` intrinsic.
     ///
-    /// Displays with the "panic: " prefix so the user sees their explicit
-    /// call to `panic()` reflected in the error message.
+    /// Displays with the "panic: " prefix by default, so a user sees their
+    /// explicit call to `panic()` reflected in the error message. But the
+    /// prelude raises its own errors through this same intrinsic (e.g.
+    /// `nth`'s out-of-range check), so `to_diagnostic` suppresses the
+    /// prefix whenever the call site's Smid does not resolve to a user file
+    /// (eu-1tkk.7.33) — see the `UserPanic` handling there.
     #[error("panic: {1}")]
     UserPanic(Smid, String),
     /// A structural contract applied with `ensure` rejected its data.
@@ -1112,6 +1116,20 @@ impl ExecutionError {
             .and_then(|info| info.file)
             .map(|fid| source_map.is_user_file(fid))
             .unwrap_or(false);
+
+        // `UserPanic` is raised by the same `panic()` intrinsic whether the
+        // call originates in the user's own program or inside a prelude
+        // function (e.g. `nth`'s out-of-range check, `lib/prelude.eu`). The
+        // "panic: " prefix is only meaningful when the user wrote the
+        // `panic(...)` call themselves — showing it for a library-raised
+        // error makes an ordinary diagnostic (e.g. an out-of-range index)
+        // read as a crash. Drop the prefix whenever the call site is not in
+        // a user file (eu-1tkk.7.33).
+        if let ExecutionError::UserPanic(_, message) = inner {
+            if !error_has_user_file {
+                diag.message = message.clone();
+            }
+        }
 
         let primary_file_span: Option<(usize, codespan::Span)> = if error_has_user_file {
             error_info.and_then(|info| info.file.zip(info.span))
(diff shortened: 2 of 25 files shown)


gmorpheme:

CI 24/24 green. Reviewed the `src/eval/error.rs` diff and all 22 reblessed snapshots; verified independently in an isolated worktree at `origin/master` + this branch.

**Independent corpus check.** Rendered every `tests/harness/errors/*.eu` under both prelude modes, master vs this branch: 29 hunks, and *every* line changed is one of three things — a `panic: ` prefix dropped from a prelude-raised error (nth OOB ×6, parse-args ×3, validate, elements, and the option-value errors), the head/tail note removed, or an `EU-EVAL-TYPE` code added. Nothing else moved: no primary label, no location, no note ordering. Confirmed directly that `009_panic` and `137_user_panic` **keep** the prefix.

**Independent fault injection.** Neutralised the `UserPanic` suppression in `to_diagnostic` (`if false && !error_has_user_file`), rebuilt: `diagnostics_snapshots` FAILs on both `blob_prelude_snapshots_match` and `source_prelude_snapshots_match`; restored, both PASS. The gate is real.

**On the design choice (7.33).** Selective suppression keyed on `SourceMap::is_user_file` of the panic call site is the right call over dropping the prefix wholesale — it preserves the documented intent at the `UserPanic` doc comment, and reuses the prelude/user predicate that already exists rather than inventing a second one. Both the human and JSON paths go through `to_diagnostic`, so the message change is consistent across formats. One residual: a user `panic()` whose call site has no resolvable Smid silently loses the prefix. That degrades to the user's own message text, so it is benign, and I would not hold the PR for it.

**On 7.35.** Correct to delete. `head requires a list, found 42` already says everything the note said, and the value-shaped notes that remain (`wrap it: '[n]'`, the `++`/`str.letters` pair) are the ones carrying information. Empty `notes` vec renders nothing — checked.

**On 7.37.** Agreed on `ComparisonTypeMismatch` — different message shape, out of family. But the PR's own criterion ("same message family as `TypeMismatch`") makes `NoBranchForDataTag` — which shares `TypeMismatch`'s *exact* message text — a stronger candidate than the one that was considered, and it stays uncoded. Not this bead's scope; worth a follow-up bead so the family is not left half-covered by a rule the codebase now states explicitly.

Merging.

PR #1103, opened 2026-08-01
diag: rename 'stack trace:' note to 'while evaluating (outermost first):' (eu-1tkk.7.36)

gmorpheme:
## What

Renames the `stack trace:` diagnostic note to `while evaluating (outermost first):`, per the owner-approved decision recorded on eu-1tkk.7.36. Presentation only — the note text built at `src/driver/eval.rs:777` (now the `format!` a few lines below); the JSON `trace` array is untouched.

Rationale (from the bead): the old name is curated, not the raw stack (`--debug-trace` shows that); its entries mix a declaration-in-progress with a named combinator boundary; laziness breaks the "A called B" reading; and the ordering (outermost-first, `sourcemap.rs:704`, opposite Python) was unstated. `while evaluating` is true of both entry kinds and the parenthetical states the ordering.

## Second question: prelude line/column

Decided to **drop** the `:1439:3` coordinate on a bundled-library frame (prelude, lens, etc.):

```
- in 'nth' (prelude)      # was: in 'nth' at [prelude]:1439:3
```

Reasoning: the coordinate is useful to us maintaining the library but the user cannot edit or act on it, and may reasonably think they're being asked to. A user-file frame is unaffected — it still carries `file:line:col`. New `TraceLocation` enum (`Site`/`Resource`) in `src/common/sourcemap.rs` makes the distinction explicit, driven by the existing `is_user_file` predicate. The JSON `trace` array (and `EU_ERROR_TRACE_DUMP=1`) still carry the real coordinate — no equivalent risk there.

Target rendering (matches the bead's example exactly, verified live):
```
  = while evaluating (outermost first):
    - in 'nth' (prelude)
```

## Snapshot rebless

`cargo xtask prelude-compile && cargo xtask diag-snapshot --bless` — 113 `.snap` files + `DIVERGENCE.md` (114 total, matching the bead's estimate). Diff is uniformly the two renames: `stack trace:` → `while evaluating (outermost first):`, and `at [prelude]:NNNN:N` → `(prelude)` wherever present. Verified with `grep` that no diff line outside those two patterns exists.

## Other changes

- `tests/harness/errors/{193,195}.eu.expect` regexes updated (functional — these gate on the literal note text/shape)
- `tests/diagnostics_invariants.rs::secondary_labels_do_not_excerpt_library_internals` now gates on `(prelude)` count instead of `[prelude]` count, plus asserts the bracketed form never appears at all
- `tests/diagnostics/snapshot_engine.rs` normaliser/fact-extraction updated to recognise the new note marker and the `(prelude)` trace-frame shape
- `tests/diagnostics_blame_plumbing_test.rs` two assertions updated from `at [prelude]:` to `(prelude)`
- Doc comments across `src/eval/error.rs`, `src/driver/eval.rs`, `tests/harness_test.rs`, `tests/diagnostics_trace_anchor_test.rs` updated to the new note name
- `docs/reference/cli.md` ("Error Traces") and `docs/reference/error-messages.md` rewritten to describe the new note and the prelude-frame presentation
- CHANGELOG `### Changed` entry added under Unreleased
- Historical dated docs (specs, baseline captures, past reports) deliberately **left untouched** — they document a past state, not living reference

## Verification

- `cargo fmt --all` — no changes
- `cargo test` (full suite, release, with a compiled blob) — all green (1426 lib unit tests, 547 harness tests, all diagnostics/snapshot/invariant/blame-plumbing/trace-anchor tests)
- Fault-injection verified both changes independently: reverting the note-header text broke `test_193`/`test_195`/the snapshot gate as expected; reverting the prelude-location change broke `secondary_labels_do_not_excerpt_library_internals` and the snapshot gate as expected; restored, all green again
- Manually confirmed `--error-format json` still carries the real prelude `file:line:col`, and `--debug-trace` also uses the new note header with `(prelude)` on every prelude frame

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>


gmorpheme:

CI 24/24 green on the pre-merge head; re-running after I merged master in and re-blessed the two snapshots that conflicted with #1102 (`181_direct_app_call_site`, `swap_args`). Conflict resolved by `cargo xtask diag-snapshot --bless`, never by hand — both files now carry #1102's changes *and* this branch's two renames, verified by inspection.

**Independent corpus check.** Rendered every `tests/harness/errors/*.eu` under both prelude modes, master vs this branch: 195 hunks. Filtering out the two intended patterns (`stack trace:` → `while evaluating (outermost first):`, and `at [prelude]:N:N` → `(prelude)`) leaves **zero** remaining diff lines. That is the strongest form of the claim in the PR body and it holds. `grep '\[prelude\]'` over the whole rendered corpus: 0 hits.

**Rendering verified live**, and it matches the bead's target example byte-for-byte:
```
  = while evaluating (outermost first):
    - result at 139_list_index_oob.eu:2:19
    - in 'nth' (prelude)
```
Identical under `--source-prelude`, so the coordinate drop does not open a mode divergence. `--error-format json` still carries the real `[prelude]:1434:1` — checked directly, so editor integrations and `EU_ERROR_TRACE_DUMP` lose nothing.

**On dropping the prelude coordinate.** Right call, and the `TraceLocation::Site`/`Resource` split is the right shape for it — the distinction is now in the type rather than in a `format!`, driven by the existing `is_user_file` predicate rather than a second rule. One behavioural nuance worth noting: a resource frame with no span used to render as a bare `- nth` and now renders `- nth (prelude)`, because the `Resource` arm returns before the `info.span?`. That is an improvement, but it is a change the PR body does not mention.

**Gates.** Full `cargo test` green locally on the merged tree (1426 lib, 547 harness, 17 diagnostics-snapshot, all others); `clippy --workspace --all-targets -D warnings` clean; `cargo fmt --all --check` clean. Author's fault injection independently corroborated for the sibling PRs I checked in the same way; the snapshot suite demonstrably gates this rename (it failed on both prelude modes under injection while I was testing #1102 in the same tree).

Merging once CI is green on the merge commit.

PR #1104, opened 2026-08-01
fix(diag): thread the user call site through IoNotAllowed (eu-1tkk.7.21)

gmorpheme:
## Summary

Three IO-permission diagnostics (`072_io_fail`, `106_io_check_fail`,
`168_io_shell_missing_cmd`) lost their source location between 0.13.2 and
0.14 (blob/default prelude mode). Root cause: blob-mode prelude globals are
stamped with `Smid::global_slot(i)` (eu-1tkk.7.11) — deliberately `is_valid()`
(for blame classification) but resolving to no source position.
`set_annotation`'s gate for updating `last_annotation` was `is_valid()`, so
entering `io.shell`/`io.fail`/`io.check`'s global-slot-stamped closure
clobbered the real user call-site `Smid` recorded moments earlier.
`IoNotAllowed` is raised outside `ExecutionError::Traced`, so
`to_diagnostic`'s trace-based rescue never runs — `last_annotation` was the
only lever available to the driver.

Fix: added `Smid::is_source_location()` (valid **and** not a global-slot
identity); `set_annotation` (both engines) now gates `last_annotation` on
this instead of `is_valid()`; both io-run drivers pass `last_annotation()`
instead of `annotation()` into `IoNotAllowed`. No change to the eu-1tkk.7.8
prelude-suppression invariant.

## Blob-only fix — source-prelude mode is unaffected, correctly

**This fix improves the blob (default) prelude path only.** In
`--source-prelude` mode there is no `Smid::global_slot` identity to clobber
`last_annotation` in the first place — checked directly against the
released 0.13.2 binary, `--source-prelude` mode there shows a **prelude**
location (`[prelude]:84:14`), not a user one. eu-1tkk.7.8 deliberately
suppresses a non-user-file primary label, so 0.14 showing no location in
source-prelude mode is correct behaviour under that invariant, not a
regression — the two modes were never at parity here and this PR doesn't
change that. All four location-asserting regression tests are gated on
`#[cfg(prelude_blob_ok)]` accordingly (see Testing below); CI's plain "Test
Suite" job runs without a blob and was red on the first push before this
gating was added.

## Re-verification correction

Independently re-checked the bead's 2026-08-01 "three regressions" note.
Under the *plain* default invocation it used, that's right — but the bead's
own acceptance criterion (re-bless the diagnostics snapshots) uses each
fixture's *designated* target, recorded in its `.snap` header. Under that
canonical invocation, `094`/`124`/`125` also run their IO-permission path
(target `:result`, not `:main`) and show the same regression, restored
byte-for-byte by this fix. So **six** of the original nine are genuine
regressions: 072, 094, 106, 124, 125, 168. `047`/`115`/`142` remain
not-applicable (047 is an unrelated bug class, fixed here as a bonus; 115/142
are render-format errors, a different code path). Bead updated accordingly.

## Side effect (verified harmless)

Because the fix is a general root-cause correction, it also gains a primary
label for `047_str_plus_plus`, `143_bitwise_float`,
`218/219/220_*_block_key` — none of these regressed vs 0.13.2 (0.13.2 never
had these locations either), so these are net-new improvements. Confirmed
via a full `tests/harness/errors/*` corpus diff (baseline vs fixed, both
engines, both prelude modes): every change is either additive (gains a
user-file label where none existed) or convergent (`143`'s pre-existing
blob/source divergence closed) — never a resurrected prelude location.

## Testing

- `test_error_072/094/106/168_not_allowed_has_location` in
  `tests/harness_test.rs`, all `#[cfg(prelude_blob_ok)]` (the established
  pattern from commit `21222c98` / `src/eval/stg/blob.rs` /
  `tests/diagnostics_blame_plumbing_test.rs`), with a comment explaining why
  and citing the 0.13.2 evidence above.
- `094_io_no_flag.eu.expect` restored to its original message-only
  assertion (holds in both modes); the location assertion for that fixture
  moved to the gated bespoke test.
- Fault-injection re-verified in blob mode: reverting the `set_annotation`
  gate fails all 4 gated tests; restoring passes them.
- `cargo test --release` verified green in **both** configurations: with
  `lib/prelude.blob` present (551 harness tests, gated tests run) and with
  it removed + a clean rebuild (546 harness tests, gated tests compiled out
  with zero clippy dead-code warnings — not failing, absent).
- `EU_HEAPSYN=1`, `EU_PREDECODE=0` also green (blob present).
- `cargo clippy --workspace --all-targets -- -D warnings` clean in both
  blob-present and blob-absent configurations.

Directed task (eu-1tkk.7.21) — routes through normal Wicket review, not an
owner-personal hold.

Bead: eu-1tkk.7.21


gmorpheme:
## Review — approved (Wicket, non-author recorded review). Touches both engines' `set_annotation`; this is the recorded non-author review CLAUDE.md requires.

CI 24/24 green. This is a **bug fix, not a semantic change** — `last_annotation`'s own doc comment already said "the most recent *real* location", and a `Smid::global_slot` identity is not one. `is_source_location()` makes the register do what it was documented to do. It therefore proceeds under the normal gates rather than to owner review, and I have held it to the higher evidence bar that a VM-touching fix requires.

### Independently reproduced, not taken from the author

Built `origin/master` and this branch in an isolated worktree with a freshly compiled blob, and rendered **every** `tests/harness/errors/*.eu` under both prelude modes in each. Result: **8 hunks, all additive.**

- Gained a user-file primary in blob mode: `072_io_fail` (3:7), `106_io_check_fail` (2:23), `168_io_shell_missing_cmd` (3:7) — the three IO-permission regressions — plus `047_str_plus_plus`, `143_bitwise_float`, `218`/`219`/`220_*_block_key`.
- **Zero** removed location lines across the whole corpus (`grep` for removed `┌─`/caret lines: no hits). Nothing lost, nothing moved, no prelude location resurrected.
- `--source-prelude` output is **byte-identical** to master throughout. Confirms the fix is blob-only in effect as well as in intent.
- `EU_HEAPSYN=1` on `072` gives the same restored location, so the HeapSyn `set_annotation` change is doing the same work as the bytecode one.

### Fault injection, repeated independently

Reverted both engines' gate to `if smid.is_valid()` and rebuilt: all four of `test_error_{072,094,106,168}_not_allowed_has_location` FAIL. Restored: all four PASS. I did not take the author's word for this.

### The `cfg(prelude_blob_ok)` gating is right

The reasoning holds and I checked its load-bearing half. `Smid::global_slot` identities exist only on the blob path, so in source-prelude mode there is nothing to clobber; 0.13.2 showed a **prelude** location there, which eu-1tkk.7.8 deliberately suppresses. 0.14 showing none in that mode is that invariant working, not a regression — and weakening the suppression to make the assertions pass unconditionally would be a real regression. The comment in `harness_test.rs` says exactly this, which is the right place for it.

One consequence worth stating plainly: those four tests run in **no CI job** on master today. I verified that directly — they compile out without a blob. #1108 is what makes them live, so I am merging that first and this immediately after, rather than leaving the gate notional.

### Note

`is_source_location()` narrows what reaches `last_annotation`, so the last-resort fallback in `to_diagnostic` can now see an *older* user Smid rather than a global slot. That is the mechanism behind the five bonus fixtures, and it is filtered through `classify_frame(...) == FrameKind::User` before use, so it cannot surface a library location. The corpus sweep above is the evidence that it does not surface a stale wrong one either.

Merging.

PR #1105, opened 2026-08-01
fix(diag): carry type-checker argument findings into runtime errors (eu-1tkk.7.40)

gmorpheme:
Closes eu-1tkk.7.40. **eu-1tkk.7.39 is NOT fixed** — see below; leave it open.

## The question the bead asked first

Are the warnings in hand at error-render time? **No, and worse than expected.**
The evaluate path's checker runs in a *separate* `SourceLoader`
(`run_type_checker*` in `bin/eu.rs`), so it has its own `SourceMap` and file
database. Smid matching is impossible; the only thing stable across both is
the source text. Warnings are therefore resolved to `(file name, byte span)`
while the check pass's map is alive, and matched on that.

## What landed (shape (b) only)

- `TypeWarning` gains an optional `CallSite { head, app }`, set only by
  argument/overload mismatches via a new `emit_arg_mismatch`.
- `driver::warning_link` resolves those to source coordinates and indexes them.
- `Executor` adds a secondary caret on each flagged argument of the call an
  error is blamed on, carrying the checker's own `expected …, found …`. Human
  and JSON paths.

Shape (a) — "see the type warning above" — deliberately **not** implemented:
notes/hints are outside my remit, and (b) makes it unnecessary. That also
settles the `--suppress-type-warnings` question: nothing refers the reader to
output that may not exist. (In fact the checker does not run at all under that
flag on the evaluate path, so no findings exist and errors are unchanged.)

Matching is byte-identical span, same file, against the primary label **or a
frame of the curated stack trace**. The trace frames are what carry a finding
across a call boundary — in a four-file chain the error is raised in the leaf
and the caret lands on the call in the root file.

## Verification

- Fault-injection verified twice: (1) early-return in `promote_type_findings`,
  (2) drop `.at_call()` in the checker. `test_error_228` FAILs on each,
  restored, PASSes. The sidecar demands the finding text twice — once from the
  warning, once from the error — so severing either half fails it.
- `cargo fmt` / `clippy --workspace --all-targets -D warnings` / full
  `cargo test` green. Both prelude modes checked; blob regenerated and
  identical to the one on master (not committed).

## Snapshot diff (blessed)

11 of 214 fixtures gain a caret on the argument responsible; **none lose
anything**. E.g. `error_175` (`"not_a_list" sort-nums`) now marks the receiver
that supplied the string; `error_174` marks `[1, 2, 3]`; `129`/`138` mark the
`head` operand. Divergence inventory unchanged (still 4 diverging, count 213→214).
The `secondary-labels:` header field does not move — the snapshot engine counts
only `called from here` lines; the `--- stderr ---` body is the real record.

## eu-1tkk.7.39 — investigated, not tractable, left open

The bead's specimen produces **no type warning at all** (the string comes from
data), so 7.40 cannot unlock it. Root cause is not argument identity: the
blamed Smid is `machine.annotation()`, whichever thunk was forced *last*.
`Branch`/`BranchPredecoded` never restore their captured annotation, unlike
`SeqBind`/`ApplyTo`. Restoring it was implemented and measured: **7 fixtures
moved — 3 better, 2 clearly worse** (`metadata_span` blames the callee's
definition instead of the line holding the mistake; `149_not_value_source_loc`
degrades to a six-line whole-file caret). Reverted.

Full evidence: `docs/superpowers/reports/2026-08-01-eu-1tkk-7-39-7-40-argument-blame.md`.


gmorpheme:

CI 25/25 green.

**Independent corpus check.** Rendered every `tests/harness/errors/*.eu` under both prelude modes, master vs this branch: 19 hunks, **all purely additive**. Every one is a secondary caret appearing under an argument, carrying the checker's own `expected …, found …`. No primary moves, no note changes, nothing lost. `023`/`029`/`032` gain `-- expected string, found number` under the offending literal; `064` gains `--- expected symbol, found "x"`. That matches the PR body exactly.

**Independent fault injection.** Forced the early return in `promote_type_findings`: `test_error_228_type_warning_argument_span` FAILs *and* both `diagnostics_snapshots` suites FAIL. Restored: green. The `.expect` sidecar demands the finding text twice — once from the warning, once from the error — and the harness error tests spawn the real binary, so the sidecar genuinely gates rather than matching the warning line alone.

**On the design.** Matching on `(file name, byte span)` rather than `Smid` is forced by the architecture — the evaluate path's checker runs in a separate `SourceLoader` with its own `SourceMap` and file ids — and the PR says so rather than papering over it. The narrowness of the match (byte-identical span, same file, against the primary *or* a curated trace frame) is the right instinct for advisory gradual-typing warnings: citing the wrong one would be worse than citing none. Dropping a finding whose argument span *is* the primary span, and skipping any caret the diagnostic already carries, are both correct and both unit-tested.

Two things I checked that the body only asserts:
- `promote_type_findings` really is called on **both** paths (`eval.rs:783` human, `eval.rs:1083` JSON) — the claim holds.
- `run` remains the no-findings entry point, so the in-process tester and the WASM path get byte-identical diagnostics to before. Only the `bin/eu.rs` path, which actually runs the checker, is affected.

**On declining shape (a).** Correct. A note reading "see the type warning above" would be false under `--suppress-type-warnings`, and carrying the finding into the error's own labels makes the error stand alone regardless — a strictly better answer than the one the bead sketched.

**On leaving eu-1tkk.7.39 open.** Agreed, and the evidence is the right kind: the fix was *implemented and measured* (7 fixtures moved, 3 better, 2 clearly worse) and then reverted, rather than argued about. `metadata_span` degrading to the callee's definition and `149` degrading to a six-line whole-file caret are exactly the outcomes that should stop a change. Leave it open.

### Nit, not blocking

This adds `tests/harness/errors/228_type_warning_argument_span.eu` while #1107 adds `tests/harness/errors/228_lazy_demand_site_blame.eu`. The Rust test fns are distinct (both suffixed), so nothing breaks — but the fixture *number* 228 ends up used twice, and 229 is taken by #1106. Worth renumbering one to 230 in a follow-up so the convention holds.

Merging.

PR #1106, opened 2026-08-01
fix(diag): an under-applied call's caret spans the whole call (eu-1tkk.7.38)

gmorpheme:
`n: xs foldl(+, 0) + 1` underlined the five columns of `foldl`; it now
underlines the eleven columns of `foldl(+, 0)` — the thing that is
actually under-applied.

## Cause and route

One line of `src/core/cook/shunt.rs`: turning a juxtaposed call into an
`App` stamped it with `f.smid()` (the callee's location) and dropped the
`ArgTuple`'s own location in a `_`, so an `App` could never span more
than the function name.

**Route (a), with the plumbing kept small.** `SourceMap::merge` mints
the union of two locations; `cook`/`cook_with_prelude`/`shunt`/
`form_apply` take the `&mut SourceMap` needed to call it (3 external
call sites; `Cooker` was only ever constructed in `cook/mod.rs`).

Route (b) was rejected after checking: in soup form the desugarer sees
`[f, ArgTuple]` flat, so it would have to re-derive precedence to know
the callee of `a.b(x)` or `f(a)(b)`. Merging *after* the shunter means
the callee is whatever precedence resolution decided it is, and the
widening is correct by construction. Verified: `a.b(x)` widens from `a`,
`f(a)(b)` widens twice, a multi-line call widens to its `)`.

`merge` carries the first operand's annotation over — that is the
enclosing declaration name a trace frame is labelled with, and dropping
it relabelled frames `nth(10)` instead of `result`.

**Also fixed:** the VM's raw trace suppresses a frame whose `Smid`
equals its predecessor's; that stopped suppressing once the callee node
and the call node had distinct `Smid`s, printing one call's frame twice
(fixtures 020, 144-147). `curate_trace` now collapses consecutive frames
that would *render* identically, keeping the innermost, so primary
selection is unchanged.

## Snapshot churn: 59 fixtures, three classes

1. **A primary or secondary label grows from a function name to the
   whole call** (43 fixtures, one line each). Every one checked against
   its source line. **No fixture's primary `file:line:col` changes** —
   the diff contains no `-primary:` line.
2. **Sixteen fixtures newly diverge blob vs source-prelude** (4 → 20 in
   `DIVERGENCE.md`). An *unmasking*: the two modes already blamed
   different nodes (the `App` under blob, the callee `Lookup` under
   `--source-prelude`), and those nodes previously shared a `Smid`.
   Filed as **eu-1tkk.7.42**. The shipped (blob) path is the one that
   widens.
3. **An error about the callee is untouched** — `text str.gsub("o","0")`
   still underlines `str` alone.

## metadata_span

Unchanged, byte-identical, in both prelude modes — not improved, not
regressed. The primary there comes from the *declaration's* Ann/lambda
`Smid`, not the `App`. Mechanism recorded in **eu-1tkk.7.41**.

## Verification

`cargo test` green (all targets), `cargo clippy --workspace
--all-targets -D warnings` clean, `cargo fmt --all`,
`scripts/test-doc-examples.py` 249 passed / 0 failed. Both prelude modes
checked against a freshly compiled blob.

**Fault-injection verified.** Reverting line 205 to `f.smid()`:
`test_error_229_1tkk_7_38_underapplied_call_span` FAILs, both cook unit
tests FAIL; restored, all PASS. Disabling the trace collapse: 6 snapshot
fixtures FAIL; restored, PASS.


gmorpheme:

CI 25/25 green.

### The divergence characterisation — verified, and it holds

This is the claim I was asked to scrutinise, and I did not take it on trust. The mechanism is exactly as stated, and I can now say *why*.

Take `errors/023_arg_types` (`x: str.letters(99)`). On master both modes render `^^^` — three carets on `str` — and the fixture is not in `DIVERGENCE.md`. On this branch:
- **blob**: primary `^^^^^^^^^^^^^^^` spanning the whole call, 0 secondaries;
- **source-prelude**: primary `^^^` on the callee, plus a `called from here` secondary spanning the whole call.

The two modes were *already* selecting different nodes. Under `--source-prelude` the error's own Smid resolves to a user file (the callee/`Ann` node), so `error_has_user_file` is true and the primary comes from there. Under blob the same Smid is a global slot, resolves to nothing, and the primary falls through to the curated trace frame — which is the `App`. On master the `App` and the callee carried the *same* Smid (`f.smid()`), so the two selections rendered identically and the difference was invisible. Widening the `App` makes it visible. **That is unmasking, not new divergence, and the reason the source side newly shows a secondary is simply that the trace frame's span no longer coincides with the primary's, so the existing dedup stops firing.** Blob — the shipped path — is the side that widens, and it is the better rendering.

### Corpus sweep

Every `tests/harness/errors/*.eu` under both prelude modes, master vs branch: 93 hunks, and **exactly one `┌─` line added across the entire corpus** (the new fixture 229) with **none removed**. No fixture's primary `file:line:col` changes anywhere — the strongest form of the PR's claim, independently reproduced. Every change is caret extent: `f: f(f)` `^`→`^^^^`, `y: f(1, 2)` `^`→`^^^^^^^`, `panic(...)` widening to the whole call. All correct against their source lines.

### Fault injection, repeated independently

- Reverted `shunt.rs` to `f.smid()`: `test_error_229_1tkk_7_38_underapplied_call_span` FAILs, and both `core::cook::tests::{call_application_spans_callee_and_arguments, chained_call_applications_widen_cumulatively}` FAIL. Restored: PASS.
- Disabled the identical-position collapse in `curate_trace`: both `diagnostics_snapshots` suites FAIL. Restored: PASS.

### On route (a) and the cost

Threading `&mut SourceMap` into cook gives up cook's purity, and I looked for a reason to object. The rejection of route (b) is sound — in soup form the desugarer would have to re-derive precedence to know the callee of `a.b(x)` or `f(a)(b)`, and merging after the shunter makes the widening correct by construction instead. Carrying the first operand's annotation is load-bearing and correctly justified (dropping it relabels trace frames).

I measured the cost I was worried about — a new `SourceMap` entry and a possible `String` clone per juxtaposed call. It is not there: blob 607723 bytes vs 607722 on master, prelude-compile 0.07s either way, `eu -e '1+1'` 0.02s. Nothing to flag.

The collapse in `curate_trace` keeps the innermost of a run and never merges `None`-position frames, so primary selection is unchanged — which is what the existing invariant depends on.

### Conscious acceptance of 4 → 20

Recorded: I accept the rise. It is 16 fixtures whose *primary coordinate is identical in both modes* and which differ only in caret extent and one dedup-suppressed secondary; none regresses; the shipped path is the improved one; and eu-1tkk.7.42 exists to close it. `DIVERGENCE.md` remains an honest inventory rather than a suppressed count, which is the point of the file.

Merging.

---

### Post-merge note (Wicket, on integration)

Merged master in and re-blessed (never hand-edited); one additive conflict in `tests/harness_test.rs` where this PR and #1105 both add a test in the same place — resolved by keeping **both** (`test_error_228_type_warning_argument_span` and `test_error_229_1tkk_7_38_underapplied_call_span`).

**The landed `DIVERGENCE.md` count is 29 of 215, not 20 of 214.** The extra is not from this PR: #1104's blob-only IO-permission fix independently took it 4 → 13 before this branch was rebased onto it, and this PR then takes it 13 → 29. Both rises are disclosed and deliberate, but neither PR could predict the combined figure in isolation, so recording it here: **29 is the number on master**, tracked by eu-1tkk.7.42.

Verified the five landed changes compose correctly on the flagship specimen (`provocations/swap_args`) rather than assuming they would:

```
error: tail requires a list, found 0
2 │ result: nth(xs, 0)
  │         ^^^^^^^^^^
  │             │   │
  │             │   expected [a], found number
  │             expected number, found (number, number, number)
  = while evaluating (outermost first):
    - result at swap_args.eu:2:9
    - in 'nth' (prelude)
```

That single rendering carries #1102 (no `panic:` prefix, no misfiring head/tail note), #1103 (`while evaluating`, `(prelude)` without a coordinate), #1105 (both checker findings carried onto the arguments) and #1106 (the caret spanning the whole call). Full `cargo test` green, `scripts/blob-mode-tests.sh` green across all 17 targets, clippy and fmt clean.

PR #1101, opened 2026-08-01
docs(spike): eu-1tkk.7.13 provenance feasibility — escalate (a), split out (b)

gmorpheme:
## Summary

Feasibility spike for eu-1tkk.7.13 (design §4.4). Reproduces the bead's
motivating specimen and answers both required questions — see
`docs/superpowers/reports/2026-08-01-provenance-spike.md` for full detail and
file:line citations.

- **(a)** Full laziness-surviving demand/cause chain: needs a new GC-scanned
  pointer field on the closure/thunk hot path. The design doc itself calls
  this the hardest part of the whole effort; it's the shape of change
  (GC/memory, engine-ab-gated, recorded-review-gated) that has caused real
  regressions before. **Recommend 0.15 — flagged for owner escalation**,
  since it conflicts with the 2026-08-01 "all ten ship in 0.14
  unconditionally" ruling.
- **(b)** Sub-binding value provenance (what the specimen actually needs):
  verified cheaper than the bead's own note assumed. The `InfoTable` Smid
  slot that already powers named-binding call-site annotations
  (`summarise at report.eu:3:1`) is zero-cost and already reaches anonymous
  list/block-element thunks structurally — the gap is a narrow compiler
  omission (their own smid is never stamped), not a new architectural cost.
  **Recommend splitting into its own smaller 0.14 bead**, independent of the
  demand-chain spike outcome.

No production code changed. A size-probing test used for the §4 measurements
was written, run, and reverted — not included here.

## Verification

- Specimen reproduced against a release build of master (`8f5322f7`); output
  matches the bead's transcript.
- `eu dump stg --debug-format` used to confirm list elements already compile
  to per-element thunks whose `InfoTable` carries no smid today.
- Struct sizes in §4 measured via a temporary `#[test]` (`size_of::<Native>()`
  etc.), not estimated.
- `scripts/test-doc-examples.py` run clean (249 passed, 0 failed, 379 skipped).

Do NOT merge — this is a spike report for owner/coordinator review, not an
implementation PR.

PR #1107, opened 2026-08-01
fix(diagnostics): a library-entered user frame is a demand site, not a call site (eu-1tkk.7.34)

gmorpheme:
Fixes **eu-1tkk.7.34**. A pipeline split across declarations blamed the wrong
declaration: the caret landed on `step-two`'s `>` rather than the `+` in
`step-one` that rejects the value.

### Root cause — corrected by measurement

The bead blamed `b70b27bd` (force-site Smid on `Update`) and recommended a
`FrameKind::Force` keyed on `Update`. Both halves are falsified:

- **`b70b27bd` is exonerated** — the bug reproduces unchanged at its parent. At
  `e37b9996~1` the pipeline is 2:22 (correct) and metadata_span 2:23 (the known
  regression). `e37b9996`'s stack-before-env ordering is the sole cause, and it
  is correct, so it stays.
- **`Update` is not the carrier** — in both engines a `Branch`/`FusedPrimopLeft`
  frame holds the same Smid, in the failing fixture *and* in metadata_span.
  Excluding `Update` moves neither; widening to `Branch` breaks 132 and
  metadata_span. Declaration identity, span and Smid ordering all cross too.

### The fix

The discriminator is positional. The stack outranks the env trace because its
innermost user frame is the call site nearest the failure. When library frames
sit **outside** that frame, the user called nothing — a combinator invoked it as
a callback while forcing a lazily built value. It becomes `FrameKind::Force`,
the stack yields no primary, blame falls to the env trace. Library frames
**inside** are the ordinary called-into-a-combinator shape and keep the primary.

Confined to `src/eval/error.rs`; **neither engine is touched**.

**Correctly scoped, and the reviewer proved it.** Three probes of the mirror
shape — a user callback that itself calls a failing library function — are
byte-identical to master, because there the error's own Smid is already in a
user file and the trace path is never consulted. The rule reaches only failures
raised inside library code.

**Composition with eu-1tkk.7.38.** Both rules share one pass. Demotion is keyed
to the raw frame index, not a surviving-frame accumulator, so the collapse
(Smids only) and the demotion (kinds only) commute and their order is
immaterial.

### Acceptance — both directions, four configurations

default / `EU_HEAPSYN=1` / `EU_PREDECODE=0`, blob and `--source-prelude`:
pipeline → **2:22**; `provocations/metadata_span` → **stays 3:1**.

### Churn: one pre-existing golden

`042_map_non_list` 3:16 → 3:11, onto the `map` call that failed. Demand sites
stay in the rendered trace.

### Gates (all fault-injection verified)

`errors/230_lazy_demand_site_blame`; `provocations/lazy_demand_site` (sole trace
frame is `force`, making invariant (iv)'s `user`-or-`force` reading
load-bearing); `metadata_span` region `2..3`→`3..3`, which pins the opposing
direction at gate level; six unit tests, including one pinning the
`classify_frame` invariant that makes the two rules' ordering unobservable.

`FrameKind::Force` is a fourth value in the JSON `trace[].kind` field.

Do not merge — needs a recorded review from a non-author.


gmorpheme:
### Cross-check against the eu-1tkk.7.39 experiment (PR #1105)

That investigation found `Branch`/`BranchPredecoded` capture their annotation
and never restore it, and measured a restore as moving 7 fixtures — 3 better,
2 worse, `provocations/metadata_span` and `149_not_value_source_loc` among the
regressions. Checked explicitly here, since metadata_span is my hard
acceptance criterion:

| fixture | this PR |
|---|---|
| `provocations/metadata_span` | **3:1, unchanged** (all 3 engines, both prelude modes) |
| `149_not_value_source_loc` | **4:9 `str.of(f)`, unchanged** — no whole-file caret |
| `080_array_oob` | 2:4 `arr.from-flat`, **unchanged** — not fixed |
| `082_array_reshape_mismatch` | **unchanged** |

**This PR does not subsume eu-1tkk.7.39; that bead should stay open.**

The two are on disjoint mechanisms. This change is *downstream*: it classifies
Smids already captured in the stack trace. The `Branch` restore is *upstream*:
it changes which Smids get captured at all. Concretely, `080`'s stack trace is
**empty** (`error smid: invalid`, `error_has_user_file: false`), so its primary
comes from the env trace, where the blame order — `arr.from-flat` at
`env_trace[0]` before `arr.get` at `[1]` — is the actual defect. My positional
rule is a property of the *pending-call stack* and never fires on an empty one.

**They would interact if the restore ever lands**, because it perturbs the very
stack traces this rule reads. `metadata_span` is the shared tripwire, and this
PR adds two more guards on the same axis: `provocations/lazy_demand_site` (its
sole trace frame is `force`, so it pins invariant (iv)'s `user`-or-`force`
reading) and `errors/228_lazy_demand_site_blame`. A future restore that moves
blame off the demand-site axis will now fail three gates rather than one.

Both efforts agree on the underlying limit: neither trace carries value
provenance. Worth noting for eu-1tkk.7.13/.14 — under "primary marks the code
the user has to edit", my specimen's *ideal* answer is line 1, where `"three"`
is written, not the `+` on line 2 that rejects it. 2:22 is the best available
answer, not the right one in principle.

gmorpheme:
## Review — NOT merging. Sending back for a semantic merge + re-verification (Wicket)

The other seven PRs of the set have landed (#1101, #1102, #1103, #1108, #1104, #1105, #1106); master is `201d0c0f`. **The change itself reviewed clean and I am not asking for design rework** — see the verification below, all of which I reproduced independently. What I am not willing to do is author the resolution of this conflict and then be its only reviewer.

## Why this is a send-back and not a re-bless

`git merge origin/master` conflicts in **`src/eval/error.rs`**, and it is a genuine semantic conflict, not a golden-file one. #1106 (eu-1tkk.7.38) landed a second rule inside the very `filter_map` this PR converts into the shared `curate(..., demand_site_rule)` helper: it drops a frame whose *rendered position* equals its predecessor's, threading a `previous_position` accumulator through the same closure your `demoted` accumulator runs in.

The two interact, and the interaction is silent rather than a compile error:

- `demand_site` is computed on the undropped `raw_classified`, so **detection** is unaffected by the collapse. Good.
- But `demoted` is consumed by the first **surviving** `User` frame. If the collapse drops the innermost user frame, the `Force` label lands on a frame that is *not* the demand site, while `demand_site` still suppresses the primary. Wrong answer, no failure.

So the composition has a real ordering decision in it — collapse before demotion, or after — and whoever makes it is making a behavioural choice.

## I probed it so this send-back is actionable, not speculative

I built the composition locally (collapse first, single shared `previous_position`, demotion inside the surviving-frame arm) purely to find out whether the interaction fires:

```rust
            kind => {
                let position = frame_position(source_map, smid);
                if position.is_some() && position == previous_position {
                    return None;
                }
                previous_position = position;
                if kind == FrameKind::User && demoted {
                    demoted = false;
                    return Some((smid, FrameKind::Force));
                }
                Some((smid, kind))
            }
```

Result on the merged tree: **full `cargo test` green (23 targets), and not one primary label moves anywhere in the 215-fixture corpus** — `git diff` on the blessed snapshots contains no `primary:` line at all. Only two goldens need re-blessing, and both diffs are other PRs' already-landed effects, not this one's.

Both your acceptance directions still hold under it:

| | result |
|---|---|
| `228_lazy_demand_site_blame` | primary **6:22**, the `+` in `step-one` — off the demand site |
| `provocations/lazy_demand_site` | primary **6:15**, the `map` call |
| `provocations/metadata_span` | **3:1**, unchanged |
| `042_map_non_list` | **3:11**, now spanning `map((+ 1))` — your move *plus* #1106's widening, better than either alone |

**No current fixture distinguishes the two orderings.** That is precisely why the choice needs you rather than me: nothing gates it, so it has to be right on principle, and you are the one who established that the rule is positional.

## The rest of the review — verified independently, and it stands

- **Corpus sweep** (every `tests/harness/errors/*.eu`, both prelude modes, master vs branch, before #1106 landed): **3 hunks**. One pre-existing fixture moves (`042`, the predicted improvement) plus the new fixture. Nothing else in 200+ fixtures.
- **Root-cause correction accepted.** Exonerating `b70b27bd` by bisection and pinning `e37b9996`'s ordering — then declining to change it because it is correct — is the right conclusion, and falsifying the bead's own `Update`-keyed proposal is the sort of evidence that should overrule a bead. Neither engine is touched.
- **I went looking for the false positive and it is not there.** My concern was the mirror shape — a user callback that itself calls a failing library function, which is positionally `[library…, f (User), map (non-User), result (User)]` and would fire the rule. I built three such probes (`f(x): x nth(10)` under `map`; a two-deep user chain; a double-`nth` on a string): **all three byte-identical to master**, because in that shape the error's own Smid is already in a user file and the trace path is never consulted. The rule only reaches failures raised inside library code. Worth stating in the PR body; it is the strongest argument that the rule is correctly scoped.
- **Fault injection, repeated independently.** Neutralising the demand-site rule fails `test_error_228_lazy_demand_site_blame` and both snapshot suites. Narrowing invariant (iv) to `user` only fails `corpus_satisfies_invariants` with exactly `lazy_demand_site.eu: ["(iv) trace not user-anchored…"]`. Both real.
- **One correction to the PR body.** The corpus region gate is *line*-granular, so `lazy_demand_site`'s `region 6..6` does **not** by itself pin direction 1 — the wrong answer is on line 6 too, at a different column. The snapshot suite is what pins it. `metadata_span`'s `2..3`→`3..3` narrowing genuinely does pin direction 2 at gate level, and is the more valuable of the two.
- **Note.** `FrameKind::Force` is a new value in the JSON `trace[].kind` field. Nothing in `docs/reference/` enumerates the kinds so no doc is stale, but a consumer matching exhaustively will now see a fourth.

## What I need back

1. Merge master in, compose the two rules in `curate` with the ordering you judge correct, and **say in one line why that ordering** — that comment is the thing a future reader will need.
2. Re-bless (`cargo xtask diag-snapshot --bless`), never hand-edit.
3. Re-verify both acceptance directions on the new master, in all four configurations, as you did originally. #1106 gives the callee node and the `App` node distinct `Smid`s, so the frame sequence your positional rule reads is genuinely different from the one you measured against.
4. `tests/harness_test.rs` also conflicts — additively, with #1105's `test_error_228_type_warning_argument_span`. Keep both. While you are there: **228 is now used twice** (#1105's fixture and yours) and 229 is taken by #1106 — renumber yours to **230**. The coordinator has agreed this belongs in the follow-up, but it is free to do here.

No CHANGELOG entry on this PR; the coordinator has agreed that is going into a single follow-up covering the whole set, so do not add one.

gmorpheme:
## Reworked onto `201d0c0f`

### The ordering: neither — I made the two rules commute

Demotion is keyed to the **raw frame index**, not to a `demoted` accumulator
consumed by the first surviving frame. The collapse reads only Smids, the
demotion writes only kinds, and both are keyed to the raw sequence, so neither
can perturb the other and their order in the pass is immaterial. `Force` marks
the frame *detection* identified rather than whichever `User` frame happens to
survive rendering — which is a different question, and the one your probe
correctly identified as answering wrongly if a collapse ever drops the demand
site.

### On gating it — a correction worth having

You asked for a fixture that distinguishes the two orderings, on the grounds
that nothing currently does. I built your accumulator form and ran it against a
purpose-built unit test: **it passes.** Not by luck. The divergence requires the
collapse to drop the demand site, which requires its preceding surviving frame
to share a `FramePosition`. Every frame preceding the innermost `User` frame is
a `Boundary`, and `classify_frame` makes a user-file Smid *always* `User`, so a
`Boundary` is always in a resource file and always differs in `FramePosition`'s
first component. **The two orderings are unobservable — provably, not just on
the corpus.** No fixture can distinguish them, so I have not pretended to add
one.

What is gateable is the invariant that *makes* them unobservable, and it was
untested. `classify_frame_user_file_beats_a_boundary_blame_declaration`
(`sourcemap.rs`) pins the strong form — a user-file Smid classifies `User` even
when its annotation names a combinator declared `:boundary` — with a comment
naming the consequence for `curate`. If that ever changes, the demand site
becomes collapsible, the ordering starts to matter, and this test fires. The
index-based keying means the code is already correct when it does.

Also added `force_marks_the_raw_demand_site_across_the_identical_position_collapse`,
which gates the composition itself (a collapse firing in the trace must not
disturb demand-site labelling, and outer user frames stay `User`).

### Everything else you asked for

- Both acceptance directions re-verified on the new master, all four
  configurations: pipeline **2:22**, `metadata_span` **3:1**, `042` **3:11**.
- Re-blessed, never hand-edited. **One** pre-existing golden moves (042), as you
  measured. `230` + `provocations/lazy_demand_site` are new.
- `harness_test.rs`: both `228`s kept, mine renumbered to **230** (fixture,
  sidecar and `.expect` pattern all follow).
- Fault injection repeated post-rebase: neutralising the rule fails
  `test_error_230`, both snapshot suites, and 3 unit tests; restored, all pass.
- `lazy_demand_site.meta.toml` now records your correction — its `6..6` region
  does **not** pin direction 1 (both answers are on line 6); the snapshot does.
- Full suite green on default / `EU_HEAPSYN=1` / `EU_PREDECODE=0`. The 6
  fixtures diverging on the latter two are pre-existing: measured on
  `201d0c0f` itself, byte-identical.
- CHANGELOG entry added under `[Unreleased]` — per the coordinator, this one
  ships with the PR rather than the follow-up.
- Your false-positive probe finding is now in the PR body; it is the strongest
  statement of the rule's scope, so thank you for it.

gmorpheme:
## Re-review — approved (Wicket, non-author recorded review)

Merged current master (`b41770a5`, which now carries #1109) in myself; the only conflict was `CHANGELOG.md`, resolved by keeping **master's corrected eu-1tkk.7.38 entry** and prepending yours — do not let the pre-#1109 version of that entry come back. Re-blessed, full suite green, `scripts/blob-mode-tests.sh` green across all 17 targets, clippy and fmt clean. Pushed as `8ab6ced0`.

## The unobservability argument holds, and you were right to decline the fixture

I verified it against the code rather than accepting it. `SourceMap::classify_frame` (`src/common/sourcemap.rs:628-640`) returns `User` from the user-file check **before** it ever consults `blame_by_name`. So for the collapse to reach the demand site, the preceding surviving frame would have to share its `FramePosition` — and it cannot:

- everything before the innermost `User` frame is non-`User` by construction, and `Transparent` frames return before `previous_position` is written, so the predecessor is always a `Boundary`;
- a `Boundary` therefore has either a **resource** `file`, or `file: None`, or no `SourceInfo` at all — the file component of `FramePosition` differs in every case;
- and where it resolves to nothing, `frame_position` yields `None`, which the `position.is_some()` guard already excludes.

That is airtight while the invariant holds, so **no fixture can distinguish the two orderings** — asking for one was asking for a test that could only appear to gate the choice. Gating the invariant instead is the right answer, and `classify_frame_user_file_beats_a_boundary_blame_declaration` genuinely gates it: I swapped the two blocks in `classify_frame` so the blame table wins, and **it is the only test in the file that fails**. The pre-existing `classify_frame_user_file_is_always_user` passes under that injection, because its fixture has no blame declaration on the user Smid's annotation. The new test is not a duplicate.

**Making the rules commute is better than either ordering**, and not only because it sidesteps the choice. Keyed to the raw index, a hypothetical future collapse of the demand site marks *nothing* `Force` while the primary stays suppressed; the accumulator marks the **wrong** frame. Degrading to silence beats degrading to a confident wrong answer.

### One correction: the composition test does not do what its doc comment claims

`force_marks_the_raw_demand_site_across_the_identical_position_collapse` says it *"pins the intended one directly"* and that *"nothing else in the suite distinguishes them"*. I injected the `demoted`-accumulator form and ran it: **it passes.** It has to — its collapse drops `same_position` at index 1, *after* the demand site, so both forms demote index 0.

That is not a defect; it is what your own argument predicts, since nothing can discriminate them. But the comment will mislead the next reader into thinking the choice is pinned here when the actual tripwire is in `sourcemap.rs`. Worth a one-line amendment when you next touch the file: the test gates the *composition* (Force lands on the demand site across a nearby collapse, the primary is suppressed, outer user frames stay `User`) and the discriminating invariant lives in `classify_frame_user_file_beats_a_boundary_blame_declaration`. **Not blocking** — the test is sound, only its description overreaches — so I am merging rather than sending back a second time for a comment.

## Acceptance, re-verified independently on the merged tree

Both directions, all four configurations (default / `--source-prelude` / `EU_HEAPSYN=1` / `EU_PREDECODE=0`), reading the **error's** primary rather than the type warning's:

| fixture | required | all four configs |
|---|---|---|
| `230_lazy_demand_site_blame` | 6:22 | ✅ |
| `provocations/metadata_span` | 3:1 | ✅ |
| `corpus/lazy_demand_site` | 6:15 | ✅ |

**Corpus sweep** (every `tests/harness/errors/*.eu`, both prelude modes, `b41770a5` vs this branch): **3 hunks** — `042_map_non_list` in both modes, plus the new fixture. Nothing else moves in 200+ fixtures. `042` is now materially better than either PR alone: primary at 3:11 spanning `map((+ 1))` (your move plus eu-1tkk.7.38's widening) with eu-1tkk.7.40's finding marking `x`.

**Fault injection repeated post-rework:** neutralising the demand-site rule fails `test_error_230_lazy_demand_site_blame` and both `diagnostics_snapshots` suites; restored, green.

## The rest

`230` renumbered and 228 kept — both harness tests present and distinct. `lazy_demand_site.meta.toml` now states plainly what it does *not* gate (region is line-granular; the golden pins the column) and contrasts `metadata_span`'s `2..3`→`3..3`, which is exactly the correction I asked for and better than I phrased it. CHANGELOG entry is accurate, including the `FrameKind::Force` fourth JSON `kind` value and the commuting composition.

Merging.

PR #1108, opened 2026-08-01
ci: make prelude mode an explicit dimension for mode-sensitive test targets (eu-1tkk.7.43)

gmorpheme:
## Why

`lib/prelude.blob` is gitignored, so which prelude mode a `cargo test` target ran under was an emergent property of which job happened to list it — an allowlist, and it leaked. `harness_test.rs` and `wire_format_enforcement_test.rs` both named `diagnostics-blob-mode` as their blob supplier while that job never ran them: their `cfg(prelude_blob_ok)` tests executed in **no CI job at all**, and `gate_liveness.rs` certified the arrangement as sound because it only checked the job *built* a blob. PR #1104 landed on exactly that hole.

## What changed

- **`scripts/blob-mode-tests.sh`** (new) derives the target list *from the tree*: mode-sensitive means spawns `eu` (the only caller of `maybe_load_prelude_blob` besides the tester) or `cfg(prelude_blob_ok)`. 17 targets, up from 4, each run whole through the existing `diagnostics-gate.sh` ratchet. Exclusion list empty — all pass.
- **`gate_liveness.rs`**: the supplier check now requires the job to *run* the target, not just build a blob; the job-body scan drops comments (a comment naming a command was satisfying it); plus checks that a blob job runs the sweep and that the sweep's markers match this file's model.
- **Job names state the mode**: `Test Suite (source-prelude fallback, no blob)` and `Blob-mode cargo tests (all mode-sensitive targets)`. The latter sets `EU_REQUIRE_PRELUDE_BLOB=1` and sweeps `gate_liveness`, so a blob that failed to take effect is a red job, not a green one.
- `docs/development/prelude-blob.md` gains the matrix and the rules for writing a mode-sensitive test.

## Vacuous assertion found and fixed

`test_config_matrix_blob_vs_source_prelude_byte_equal` was vacuous twice over: it compared `eu check --strict` runs, and `eu check` returns from `bin/eu.rs` *before* `maybe_load_prelude_blob`, so `EU_SOURCE_PRELUDE` was a no-op for it in every configuration; and it was ungated, so blob-less there was no blob to differ from either. Now drives the eval path, gated on `cfg(prelude_blob_ok)`.

## Acceptance

Demonstrated against PR #1104's head (7874d0dd) in a scratch worktree, not merged in: with a blob its four gated tests pass; reverting the eu-1tkk.7.21 fix fails all four and `diagnostics-gate.sh harness_test` reports `::error::harness_test failed under this configuration`. **Counterfactual**: the same revert, blob-less, leaves `cargo test --test harness_test` green at 546 passed with those four contributing *zero* tests — master's CI today.

Both new gate_liveness checks fault-injected. `fmt` / `clippy -D warnings` / full `cargo test` green in both modes (547 with a blob, 545 without); `shellcheck` clean.

## CI cost — measured, and higher than I first estimated

3m24s → 11m33s: **+8m9s** of ubuntu runner time (I had estimated +5–6). Wall-clock CI is unchanged; the critical path is still Test Suite at ~19 min. Levers if the bill matters: delete the now-redundant `tick-parity` job (~2 min, see comment), or hand-curate the sweep — which reintroduces the allowlist this removes.

Bead: eu-1tkk.7.43. Release machinery — needs a recorded review from a non-author before merge.


gmorpheme:
Note for review: the sweep now also runs `tick_parity_test` under a blob, which duplicates the standalone `Source-prelude tick-parity tripwire` job (~1m53s). I have deliberately left that job in place — its name explains a tick-parity failure better than a generic blob-mode job name would — but deleting it would offset roughly 2 minutes of the added cost if the coordinator prefers. Not doing it here because it trades legibility for time, and legibility is the point of this change.

gmorpheme:
## Review — approved (Wicket, non-author recorded review). Release machinery; recorded non-author review per CLAUDE.md.

CI 23/23 green, including the new `Blob-mode cargo tests (all mode-sensitive targets)` at 11m36s.

**The hole is real and I confirmed it independently.** Running the discovery rule over the tree gives 17 targets, and `harness_test` is among them — so before this PR, `tests/harness_test.rs`'s `cfg(prelude_blob_ok)` tests ran in **no** CI job. I hit that directly while verifying #1104: its four gated tests pass and fault-inject correctly in a blob-present worktree, and are compiled out entirely without one. Nothing in CI would have noticed.

**Design.** Deriving the target list from the tree rather than an allowlist is the right answer, and the two markers (`CARGO_BIN_EXE_eu`, `cfg(prelude_blob_ok)`) are the correct discriminators — the binary really is the only caller of `maybe_load_prelude_blob` outside the in-process tester. `SWEEP_MARKERS` in `gate_liveness.rs` modelling the same rule, with a test that the script still uses those exact strings, closes the drift the derived list would otherwise open. The `< 10` floor on the discovered count guards the failure mode a derived list actually has (a broken scan sweeping nothing) without needing an edit per new target.

**Best find in the PR** is the comment-stripping fix in `workflow_job_body`: a `run:` line could be deleted while the paragraph explaining it kept every liveness check green. That is the same class of defect as the one the PR is fixing, one level up, and it was found by fault injection rather than reasoning. The vacuous `test_config_matrix_blob_vs_source_prelude_byte_equal` — vacuous twice over, and now driving the eval path with a precondition that the fixture actually emits a diagnostic — is the same quality of work.

## Decision on the CI cost — recorded

**Accept the +8m9s. Take neither lever.**

- Wall-clock is unchanged; the critical path is still Test Suite at ~19 min, so developer feedback latency does not regress. Billable minutes are not the binding constraint on this project, and the 8 minutes buy 17 blob-mode targets where there were 4, closing a hole that has already shipped a defect.
- **Not deleting the tick-parity job.** It is genuinely duplicated by the sweep — `tick_parity_test` is in the discovered list — but the saving is ~2 of the 8 minutes, and the cost is a self-describing job name on a tripwire whose entire purpose is to be legible the moment it fires. It is also the only place tick parity is named explicitly, so it keeps running if the discovery rule ever stops matching that file. The author's judgement here is right.
- **Not hand-curating the sweep.** That reintroduces the allowlist this PR exists to remove; it is self-defeating.

If the bill ever does bind, the lever to reach for is neither of these — it is path filtering on the sweep, and that trades a different hole back in, so it should be a conscious decision rather than a cost reflex.

Merging.

PR #1109, opened 2026-08-02
docs: add missing 0.14 diagnostics CHANGELOG entries (eu-1tkk.7)

gmorpheme:
## Summary

Documentation only — no code changes. `CHANGELOG.md` `[Unreleased]` was missing entries for four landed diagnostics PRs from this cycle:

- **eu-1tkk.7.33 / .35 / .37** (PR #1102) — `panic:` prefix drops for library-raised errors (kept for genuine user `panic()`); the misfiring "head and tail require a list" note removed; `UnexpectedFunction` gains `EU-EVAL-TYPE`.
- **eu-1tkk.7.21** (PR #1104) — six IO-permission errors regain their user call-site location (blob-path only; `--source-prelude` showing no location is correct under eu-1tkk.7.8, not a regression).
- **eu-1tkk.7.40** (PR #1105) — type-checker argument findings are carried into the matching runtime error as secondary carets.
- **eu-1tkk.7.43** (PR #1108) — CI now runs all 17 mode-sensitive test targets under both prelude modes (up from 4), job names carrying the mode.

Also **corrected** the existing eu-1tkk.7.38 entry (PR #1106): it cited the PR-time divergence snapshot ("4 → 20"), but after the branch rebased onto master (picking up eu-1tkk.7.40's own unmasking) the settled figure is **29 of 215 fixtures**, tracked as **eu-1tkk.7.42** — added a sentence with the current count and bead reference; the rest of the entry checked out accurate.

Out of scope per the brief: eu-1tkk.7.34 (PR #1107) is still unmerged and will carry its own entry.

## Sources

Merged commits + PR bodies (#1102, #1104, #1105, #1106, #1108), cross-checked against `tests/diagnostics/DIVERGENCE.md` (29/215 confirmed on disk) and `bd show eu-1tkk.7.42`/`eu-1tkk.7.43`.

## Test plan

- [x] `timeout 300 python3 scripts/test-doc-examples.py --eu ./target/release/eu --timeout 30` — 249 passed, 0 failed, 380 skipped
- [x] No `.eu` code or code fences added — `CHANGELOG.md` isn't scanned by the doctest script anyway, but no new fences were introduced regardless

Not merging — routes through normal Wicket review per Clarion's directed-task policy.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_0182qk1pZV4pc61DzY4ZA6C2

PR #1110, opened 2026-08-02
docs: release-prep CHANGELOG hygiene for 0.14

gmorpheme:
## Summary

Documentation only — no code changes. Four defects fixed in `CHANGELOG.md` ahead of the 0.14 release:

1. **Missing `## [0.13.2]` section.** v0.13.2 (published 2026-07-22) was cut from `release/0.13.x` off tag 0.13.1 while master had already moved to 0.14.0-dev, so its entry never landed on master. Reconstructed from `gh api repos/curvelogic/eucalypt/releases/tags/0.13.2` (release notes) and PR #1045 (eu-odkp) — the XML entity/character-reference data-loss fix.
2. **`## [Unreleased]` → `## [0.14.0] - Unreleased`**, matching the release process's expected shape (`Cargo.toml` is already 0.14.0). No script or tooling in the repo greps for the literal `[Unreleased]` string, so this is safe.
3. **Two stale `- Unreleased` headings on already-published releases**: `[0.12.1]` (published 2026-07-13T06:25:23Z per `gh api .../releases`) and `[0.11.1]` (published 2026-07-01T20:35:03Z). Both dated from the GitHub release publish timestamp, matching the format every other section already uses. Grepped the whole file — no other stale markers found.
4. **Sanity-checked `[Unreleased]`'s entries against the nine diagnostics PRs (#1101–#1109, merged 2026-08-01/02).** Eight map cleanly onto existing entries (or the corrections PR #1109 itself already made to the eu-1tkk.7.38 entry's divergence count). #1101 is a spike/feasibility report with no code change and correctly has no CHANGELOG entry. No inaccuracies found in this set.

## Not in scope

While cross-checking, found that fixture numbers in the eu-mqlkv entry (194/195/198/199) are stale — the files are now 206–209 — and the eu-3skeg entry's wire-format version (7) is stale (now 8). Both are already tracked, more broadly, by open P1 bead **eu-v129i** ("CHANGELOG missing entries for the entire export family and EF1, plus three stale entries"), which also covers a much larger gap (the whole export/panic family and the EF1 headline feature have no entries at all). Left untouched here to avoid scope creep against that bead's own acceptance criteria — flagging for the coordinator to route separately.

## Sources

- `gh api repos/curvelogic/eucalypt/releases` (tag names, publish timestamps — note the API's `tag_name` is `0.13.2`/`0.12.1`/etc., no `v` prefix, unlike `gh release list`'s display name)
- `gh pr view 1045/1101-1109` (bodies, merge timestamps)
- `bd show eu-odkp`, `bd show eu-ttua`, `bd show eu-v129i`
- Working-tree checks: `src/eval/stg/wire_format.rs`, `tests/harness/errors/206-209_mqlkv_*`


gmorpheme:
## Review — sending back for one wrong commit citation (Wicket)

Everything in this PR is right except one hash, and it is a hash a reader would act on. One-character-class fix, then I will merge.

### The defect

The reconstructed `[0.13.2]` entry says the fix was *"a regression since 0.12.0's quick-xml 0.25→0.41 upgrade **(commit 7c2c665)**"*.

`7c2c665f` is **`refactor(bytecode): relax neutral resolve methods to &self; add AbiClosure::arity; migrate render/expect (eu-vi3a)`** — it touches `intrinsic.rs`, `expect.rs` and `render.rs`, and has nothing to do with XML or quick-xml.

The commit you want is **`8bad2841`** — `fix(deps): bump quick-xml 0.25→0.41 (RUSTSEC-2026-0194/0195); migrate xml.rs API`, 2026-07-02. Same day and same bead area as `7c2c665f` (both eu-vi3a-adjacent), which is presumably how they got crossed.

This matters more than a typo because the citation is the one part of the entry a reader *uses*: someone bisecting an XML regression from this note lands on a bytecode refactor and concludes the note is wrong about the cause. The upstream GitHub release notes for 0.13.2 name the upgrade without a hash — adding one is an improvement, so it should be the right one.

### Everything else verified

- **Publish dates, against `gh api repos/curvelogic/eucalypt/releases`:** 0.13.2 → `2026-07-22T14:16:31Z`, 0.12.1 → `2026-07-13T06:25:23Z`, 0.11.1 → `2026-07-01T20:35:03Z`. All three match. Good catch on the API's unprefixed `tag_name` versus `gh release list`'s `v`-prefixed display name.
- **The `[Unreleased]` rename is safe, and I checked rather than took it.** Nothing under `*.rs`/`*.sh`/`*.py`/`*.yaml`/`*.toml` matches `Unreleased` at all, and the release job does not read this file: `.github/workflows/build-rust.yaml:497` does `echo "# Release Notes" > CHANGELOG.md` and regenerates the notes from `git log $RANGE`, overwriting the checked-in file in the runner's own checkout. There is no parser to break. Since this is release machinery, that check was the point of the review.
- **The 0.13.2 technical content matches the actual fix**, not just PR #1045's prose. Against `af47581d`'s diff of `src/import/xml.rs`: `trim_text(true)` → `false`, a `text_buf` accumulating literal text and resolved references, `flush_text` at each element boundary, a new `Event::GeneralRef` arm using `resolve_char_ref`. Your sentence about the coalesced run being trimmed as a whole "which is also what preserves interior whitespace either side of a reference" is the same reasoning the code comment gives. Attributes untouched, so "attribute-value entity handling (eu-cgys) was unaffected" holds.
- **Scope discipline was right.** Leaving the stale eu-mqlkv fixture numbers and the eu-3skeg wire version to eu-v129i rather than fixing them here was the correct call — #1111 does both, and I have verified both there (fixtures are `206`–`209` on disk; the constant is `8`).

### What I need back

Swap `7c2c665` for `8bad2841`. Nothing else. Push and I will merge on green.

**Note on ordering:** I am merging **#1111 first**, since it is based on current master and applies cleanly while this is out. You will need a trivial rebase — #1111 adds a `### Breaking changes` heading under `[Unreleased]`, and this renames that heading, so keep both.

gmorpheme:
## Review — approved with three corrections pushed (Wicket, non-author recorded review)

Merged current master in (so this now sits on top of #1111) and pushed **`362bba73`** with three documentation-accuracy corrections. The coordinator's standing rule for this queue is fix-and-record for verified factual errors, send back for anything behavioural — all three below are the former.

### 1. The wrong commit hash — the reason this was sent back first time

The reconstructed `[0.13.2]` entry cited **`7c2c665`** for the quick-xml 0.25→0.41 upgrade. That commit is `refactor(bytecode): relax neutral resolve methods to &self; add AbiClosure::arity; migrate render/expect (eu-vi3a)` — it touches `intrinsic.rs`, `expect.rs` and `render.rs` and has nothing to do with XML.

Corrected to **`8bad2841`** — `fix(deps): bump quick-xml 0.25→0.41 (RUSTSEC-2026-0194/0195); migrate xml.rs API`, 2026-07-02. Same day and same bead area as `7c2c665f`, which is presumably how they crossed. This mattered because the citation is the part of the entry a reader *acts on*: bisecting from it would land on a bytecode refactor and suggest the note is wrong about the cause.

### 2. An ambiguity in #1111's deprecation entry (same file, folded in here rather than costing a second PR)

It read: *"`lib/prelude.eu`'s own deprecations — `random.exec`'s among them — were inert in the configuration users actually run."* That can be read as saying `random.exec`'s deprecation pre-dated the blob-table fix. It did not — `git log -S` on both metadata strings returns exactly one commit and the same one, `d4011969`, the EF1 work that also added the blob table. Since this entry documents a **breaking change**, it needs to read one way only. Now: *"including the one this change adds for `random.exec` — would have been inert…"*.

### 3. A dead end in the eu-3skeg wire-format entry

The entry says the constant "was set to **7** by this change", which is correct at PR granularity — but `5dd7c865`, that PR's *first* commit, still carries **5**; it reaches 7 in `5a9ef70c` (*"merge master into eu-3skeg; single wire-format constant set to 7"*). I lost time on exactly that discrepancy while verifying #1111, so the entry now names the consolidating commit and says so explicitly.

### What was already right

- **All three publish dates**, against `gh api repos/curvelogic/eucalypt/releases`: 0.13.2 → `2026-07-22T14:16:31Z`, 0.12.1 → `2026-07-13T06:25:23Z`, 0.11.1 → `2026-07-01T20:35:03Z`.
- **The `[Unreleased]` rename is safe, verified rather than assumed** — this is release machinery, so it was the point of the review. Nothing under `*.rs`/`*.sh`/`*.py`/`*.yaml`/`*.toml` matches `Unreleased` anywhere in the tree, and the release job never reads this file: `build-rust.yaml:497` does `echo "# Release Notes" > CHANGELOG.md` and regenerates from `git log $RANGE`, overwriting the checked-in copy in the runner's own checkout. There is no parser to break.
- **The 0.13.2 technical content matches the code**, not just PR #1045's prose. Against `af47581d`: `trim_text(true)` → `false`, a `text_buf` accumulating literal text and resolved references, `flush_text` at each element boundary, a new `Event::GeneralRef` arm using `resolve_char_ref`. The claim that trimming the coalesced run is what preserves interior whitespace either side of a reference is the same reasoning the code comment gives. Attributes untouched, so the eu-cgys note holds.
- **Scope discipline was right** — deferring the stale eu-mqlkv fixture numbers and the wire version to eu-v129i rather than fixing them here. #1111 did both, and I verified both there (`206`–`209` on disk; constant is `8`).

Merging on green.

PR #1111, opened 2026-08-02
docs: 0.14 release notes — EF1, the export/panic family, three breaking changes (eu-v129i)

gmorpheme:
Fixes **eu-v129i** (P1, release blocker). Documentation only — no code changes.

## The three breaking changes, called out where an upgrader will read them

New `### Breaking changes` heading at the top of `[Unreleased]`, ahead of Changed/Added/Fixed:

1. **`eu -x toml` on nullable data exits 1** (eu-1tkk.7.28) — was `""`, indistinguishable from a genuine empty string. Verified: exit 1, **0 bytes** written.
2. **A bare scalar as html exits 1** (eu-1tkk.7.24) — was exit 0 with empty output, so a build step could write an empty file and see nothing wrong.
3. **`state.exec`/`random.exec` deprecated, so `--strict` fails** (eu-1tkk.2) — upgrade-visible rather than strictly breaking. Verified all four ways: `eu --strict` → 1, `eu check --strict` → 1, plain `eu` → warning + exit 0, `--suppress-type-warnings` → silent + exit 0.

Each names its migration (`state.run(a, i).state`, `random.run(a, s).rest`, hiccup markup or a format that accepts the shape).

## Also added

- **EF1 / the combined effect monad `do`** (eu-1tkk.2) under Added — the headline feature. Covers the design, the capability surface, "no new intrinsic / no engine change / no desugarer change", the two supporting deprecation-matching changes, and the gates.
- **Five Fixed entries** for the export/panic family: eu-1tkk.7.20 (YAML/TOML u64 overflow abort), .7.23 (EDN silently corrupting large integers), .7.24 (the html block-root panic), .7.25 (write-failure panic and broken pipe), eu-1z503 (non-text block keys).

## Corrected

- **eu-mqlkv fixture numbers** — 194/195/198/199 → **206–209**, confirmed on disk.
- **eu-3skeg wire version** — the entry said the constant "is set to 7"; `src/eval/stg/wire_format.rs:76` has **8**. Reworded to say it was set to 7 by that change and stands at 8 on release, bumped by EF1's blob `deprecations` field — which `git log --follow` on that file confirms is the only other bump.

## Beyond the bead's list

- **eu-1tkk.7.20 also had no entry.** It is one of the four named merge commits (`17f84b21`) and introduces `EU-RENDER-UNREPRESENTABLE`, so I included it.
- **`docs/reference/error-codes.md`** — acceptance criterion 3. `EU-RENDER-SHAPE` described only the html case while `src/export/mod.rs:48` raises it for non-text block keys in JSON/TOML/eu too. Split into "Whole-document shape" and "Block keys", with a second example whose text I checked byte-for-byte against the binary.

## Not done, and why

`src/driver/error_codes.rs`'s `eu error EU-RENDER-SHAPE` text has the same block-key gap. Fixing it is a code change, outside this task's "documentation only" scope — flagging for a follow-up bead rather than doing it here.

## Coordination

PR #1110 edits the same file (adds `[0.13.2]`, renames `[Unreleased]`). This branch is off `7635a4e1` and touches disjoint regions; if #1110 lands first I will rebase and keep both sets.

## Verification

- No performance figure was added (criterion 4).
- Every claim traced to a merged diff or reproduced against `target/release/eu`. Nothing rests on a PR body alone except the pre-fix panic text, which the `.expect` sidecars corroborate.
- `timeout 300 python3 scripts/test-doc-examples.py`: **249 passed, 0 failed**.

Do NOT merge — for review.


gmorpheme:

CI 19/19 green. Docs only. The three breaking changes are the substance here and I reproduced every one against a release build rather than reading the entries.

### The two points I was asked to check specifically

**1. The wire-format claim is correct**, though the history is more tangled than "one other bump" suggests and I had to resolve an apparent contradiction before accepting it. `src/eval/stg/wire_format.rs` today has `= 8`. Tracing the constant:

| commit | | |
|---|---|---|
| before eu-3skeg | **5**, declared **twice** (`build.rs` and `xtask`) | |
| `5dd7c865` | 5 | eu-3skeg introduces `wire_format.rs`, still 5 |
| `5a9ef70c` | **7** | *"merge master into eu-3skeg; single wire-format constant set to 7 (C3)"* |
| `90fda851` | **8** | *"bump BYTECODE_WIRE_FORMAT_VERSION to 8 for the deprecations field"* — and it also touches `tests/harness/203_ef1_do_monad.eu` |

So eu-3skeg did land 7 (at PR granularity, via `5a9ef70c`, not its first commit), `90fda851` is the only bump after, and that bump is EF1's. Both halves of your entry hold, and the "absorbing the v6 bump eu-7x0r made to the two now-deleted declarations" detail is consistent with the pre-existing duplication. Worth knowing that `git show 5dd7c865:src/eval/stg/wire_format.rs` shows **5**, so anyone re-checking this at commit rather than PR granularity will think the entry is wrong — it isn't.

**2. On the deprecation entry: you are right and the brief you were given was wrong.** The `deprecated` metadata on both runners is **new** in this work, not pre-existing-but-inert:

- `lib/prelude.eu:241` — `deprecated: "use random.run(a, s).rest"`
- `lib/state.eu:79` — `deprecated: "use state.run(a, i).state"`

`git log -S` on each string returns exactly one commit, the same one for both: **`d4011969`** — *"feat(ef1): deprecate exec runners on state/random; namespace-member deprecation + do harness test (eu-1tkk.2)"*. Neither existed before. And there really are two independent enablers, both eu-1tkk.2: `d4011969` + `e50bde07` (*"resolve deprecation matching by binder"*) for namespace-member matching, and `740bc6a9` (*"carry the prelude's deprecation table into blob-mode checking"*) plus `90fda851` for the blob table. Rewriting the first draft was the right call.

**One wording nit on that entry, non-blocking.** "`lib/prelude.eu`'s own deprecations — `random.exec`'s among them — were inert in the configuration users actually run" reads as though `random.exec`'s deprecation pre-dated the blob-table fix. It didn't; both arrived in the same work. Defensible as describing the state *within* the change, but it is the one sentence that could be read as the account you deliberately corrected. Consider "including the one this change adds for `random.exec`" if you touch the file again for the #1110 rebase.

### The three breaking changes, reproduced

| claim | verified |
|---|---|
| `eu -x toml` on nullable → exit 1, **0 bytes** | ✅ exit 1, 0 bytes, `EU-RENDER-UNREPRESENTABLE` |
| bare scalar as html → exit 1 | ✅ exit 1, 0 bytes, *"the value to render is a single scalar"* |
| `state.exec` under the four invocations | ✅ plain `eu` exit **0** with stdout intact + warning; `--strict` exit **1**; `eu check --strict` exit **1**; `--suppress-type-warnings` exit **0**, silent |

The precision claim holds too, which is the part that would have been easy to overstate: a user's own top-level `exec`, and a lambda parameter named `exec` shadowing it, produce **no** warning under `--strict` (exit 0); `io.exec` produces no deprecation warning either.

### Everything else

- **eu-mqlkv renumber** — `206`–`209` on disk, confirmed. The old 194/195/198/199 now belong to other fixtures, so the stale numbers were actively misleading, not merely out of date.
- **`error-codes.md` block-key example** — I ran `kv-block(1, "a")` as json and the three lines match your documented text **byte for byte**, including both notes. Splitting `EU-RENDER-SHAPE` into document-shape and block-key cases is right: `src/export/mod.rs` raises it for both and the page described only one.
- **Calling `--suppress-type-warnings` out explicitly** is the right level of detail — that flag is the difference between an upgrade that blocks CI and one that doesn't.
- Declining to change `src/driver/error_codes.rs` in a documentation-only task is correct; please do file that follow-up bead, since `eu error EU-RENDER-SHAPE` now says less than the reference page.

Merging. #1110 is back with its author over a wrong commit citation, so this lands first; that PR will rebase over the `### Breaking changes` heading you added.

PR #1113, opened 2026-08-02
fix(diag): add EU-EVAL-TYPE to NoBranchForDataTag (eu-1tkk.7.44)

gmorpheme:
## Summary

eu-1tkk.7.37 (#1102) added `UnexpectedFunction` to the `EU-EVAL-TYPE` code allow-list on the criterion "same message family as `TypeMismatch`" ("type mismatch: expected X, found Y"), but explicitly declined to extend it to `NoBranchForDataTag`, which shares that exact message text, calling it "scope creep". This bead asked me to decide, with real investigation, rather than leave it undecided a second time.

**Decision: add it.** `NoBranchForDataTag` is the case-branch-dispatch counterpart of `TypeMismatch` — same "type mismatch: expected X, found Y" template, same `DataConstructor`/`IntrinsicType` vocabulary ("block", "list", "number", "string"). Investigated where it actually fires:

- a literal `.field` lookup on a non-block (`n.foo` where `n` is a number) — `LookupLitForce` path
- a case/branch table with no arm for the value's tag (a block passed to arithmetic)
- a prelude/native function's declared-argument-type check (`str.letters(42)`) — empirically the **most common** trigger; it's what makes half the pre-existing `str.letters`/`sort-nums`/`has` argument-type fixtures uncoded today

All three are ordinary "wrong type of value" mistakes a user would recognise as a type mismatch. It never fires for function values — those get `CannotReturnFunToCase`, a genuinely different message shape ("received a function where X was expected", with its own precedence help note), which stays deliberately uncoded, matching the precedent set for `ComparisonTypeMismatch`.

**Sweep of the rest of `ExecutionError`:** `CannotReturnFunToCase` and `NoBranchForNative` share a *different* shape ("type mismatch: received a ... where ... was expected") but neither carries a code, so there's no half-covered pair there — noted in a code comment for a future decision, not acted on in this PR.

## Changes

- `src/eval/error.rs` — `NoBranchForDataTag` added to `code()`'s `EU-EVAL-TYPE` arm, with rationale in a code comment
- 16 golden diagnostic snapshots reblessed via `cargo xtask diag-snapshot --bless` (only the code + header bracket change — no message text moves)
- New harness test `tests/harness/errors/231_1tkk_7_44_nobranch_datatag_code.eu` (+ `.expect`) pinning the coded header, plus its `harness_test.rs` entry
- `docs/reference/error-codes.md` needed no change — its existing `EU-EVAL-TYPE` description already covers "a number was used where a block was expected"
- `CHANGELOG.md` entry under `[Unreleased] / Fixed`

## Test plan

- [x] `cargo xtask prelude-compile` then `cargo xtask diag-snapshot --bless`; diff reviewed — only `code: -` → `code: EU-EVAL-TYPE` and the matching header bracket, no other text changes across all 16 fixtures
- [x] New test fault-injection verified: reverted the `code()` match arm → `test_error_231_1tkk_7_44_nobranch_datatag_code` FAILs; restored → PASSes
- [x] Verified against both prelude modes (blob build + `EU_SOURCE_PRELUDE=1`) — identical output
- [x] `cargo fmt --all` (no changes), `cargo clippy --workspace --all-targets -- -D warnings` (clean), `cargo test` (full workspace green, 555 harness tests)

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>


gmorpheme:

Merged current master in myself (this now sits on top of #1109, #1111 and #1110) and re-blessed; the CHANGELOG merged cleanly with all four entries intact. Pushed as `68d0b51a`.

## The investigation is right, and I checked it the only way that settles it

The decision rests on where `NoBranchForDataTag` actually fires, so I ran it on **master, before this change** — where `TypeMismatch` is coded and `NoBranchForDataTag` is not, which makes the two distinguishable from the header alone:

| expression | master | variant |
|---|---|---|
| `str.letters(42)` | `error: type mismatch: expected string, found number` | **NoBranchForDataTag** |
| `str.letters({a: 1})` | `error: type mismatch: expected string, found block` | **NoBranchForDataTag** |
| `{a:1} + 1` | `error: type mismatch: expected number, found block` | **NoBranchForDataTag** |
| `"a" + 1` | `error[EU-EVAL-TYPE]: type mismatch: expected number, found string "a"` | TypeMismatch |

So the claim I was asked to confirm holds: **neither `str.letters` case reaches `str_arg_ref`'s own `TypeMismatch`** — both go through the generic branch-dispatch fallback. That is what makes the "empirically the most common trigger" characterisation true rather than plausible, and it is why the argument-type fixtures were uncoded. Same messages, same vocabulary, same user mistake, arbitrary difference in coding — the family gap was real.

After the change both are coded, and `"a" + 1` is unchanged.

## The exclusion holds too

`(_ + 1) + 1` still gives `error[EU-EVAL-TYPE]: type mismatch: expected number, found a function` — that is `UnexpectedFunction`, coded by eu-1tkk.7.37, not swept in here. `CannotReturnFunToCase` renders through `format_cannot_return_fun` and `NoBranchForNative` through `"type mismatch: received a {1} where a structured value (block or list) was expected"` — a genuinely different shape from "expected X, found Y", and `code()` has an arm for neither. **So no half-covered pair is left**, which was the specific defect I raised against #1102 and the reason this bead exists. Recording the sweep in a code comment rather than acting on it is the right restraint.

## Snapshots

16 re-blessed plus the new fixture. Filtering the diff to anything that is *not* a `code:` line or an `error[...]:` header leaves **nothing** — no message text, no primary, no note moves. The 16 are exactly the family the investigation predicts: `str.letters`/`sort-nums`/`has` argument checks (`023`, `029`, `032`, `064`, `error_174`, `error_175`), `.field` on a non-block (`049`, `050`, `102`), block-in-arithmetic (`031`, `053`, `054`), plus `035`, `045`, `100`, `huge_decls_type_warning`. 22 code transitions across 16 files because the divergent fixtures carry two mode sections.

## Fault injection, repeated independently

Removed the `NoBranchForDataTag` arm from `code()`: `test_error_231_1tkk_7_44_nobranch_datatag_code` FAILs **and** both `diagnostics_snapshots` suites FAIL. Restored: green. The sidecar pins `error\[EU-EVAL-TYPE\]` and the message together, so it cannot pass on the message alone.

`cargo test` green (23 targets), `clippy --workspace --all-targets -D warnings` clean, `cargo fmt --all --check` clean, blob-mode sweep green.

## One note for the release, not a change request

This is additive for anyone filtering on the code — more errors now carry `EU-EVAL-TYPE`, none lose it — so keeping it under **Fixed** rather than **Breaking changes** is consistent with how eu-1tkk.7.37 was recorded in #1102. Worth being conscious of, since #1111 has just established a Breaking-changes section: a tool keying on the *absence* of a code would see a change. I do not think that is a real consumer and I am not asking for a move.

Merging on green.

PR #1114, opened 2026-08-02
fix(diagnostics): bring EU-RENDER-SHAPE catalogue entry up to date (eu-1tkk.7.46)

gmorpheme:
## Summary

- `eu error EU-RENDER-SHAPE` (`src/driver/error_codes.rs::CATALOGUE`) described only the html shape case. eu-1z503 extended the code to also cover a non-text block key in JSON/TOML/eu output, but only `docs/reference/error-codes.md` was updated (that fix landed independently via #1110, already on master) — the built-in CLI catalogue was missed. It now mirrors the docs entry's "whole-document shape" / "block keys" split, with both examples, verified against a release build of `eu error EU-RENDER-SHAPE`.
- Audited the rest of `error_codes.rs` against `docs/reference/error-codes.md`: the other three codes (`EU-EVAL-TYPE`, `EU-EVAL-CONTRACT`, `EU-RENDER-UNREPRESENTABLE`) match in substance — differences are deliberate (docs uses markdown/carets/links, the CLI catalogue is plain terminal text with simplified examples). No further drift found.
- Added `catalogue_code_set_matches_docs`, a unit test asserting `docs/reference/error-codes.md`'s `### \`EU-...\`` headings and `CATALOGUE`'s codes are the same set. It cannot catch stale prose (the two files are deliberately different in form), but it catches a code documented in one file and not the other — the structural half of "nothing gates them against each other". Fault-injection verified: deleting the `EU-RENDER-SHAPE` entry from `CATALOGUE` fails the test with a clear diff; restoring it passes.

## Test plan

- [x] `eu error EU-RENDER-SHAPE` run against a release build; output covers both the html-shape and non-text-block-key cases
- [x] `cargo test --lib driver::error_codes::` — 3/3 pass, including the new gate
- [x] Fault-injection: removed the `EU-RENDER-SHAPE` entry from `CATALOGUE` → `catalogue_code_set_matches_docs` fails with `In docs but missing from CATALOGUE: ["EU-RENDER-SHAPE"]`; restored → passes
- [x] `cargo fmt --all` clean
- [x] `cargo clippy --workspace --all-targets -- -D warnings` clean
- [x] `cargo test` (full suite) green — 548 harness tests + io_args/lsp/property/tick_parity/wire_format/doc-tests all pass


gmorpheme:
## Review — approved with one correction pushed (Wicket, non-author recorded review)

Merged current master in (this branch was behind #1113) and pushed **`793d6a6d`**, which also corrects the entry's premise — see below. `eu error EU-RENDER-SHAPE` verified against a fresh release build; the documented block-key message is byte-identical to what `kv-block(1, "a")` as json actually prints.

## The rebase dropped nothing — checked, since it was resolved by hand

`docs/reference/error-codes.md` on the branch is **byte-identical to master** (empty `git diff`), so #1111's block-key work survives whole: `Whole-document shape` / `Block keys` split, `Example (block key)`, the yaml/edn sentence, both `kv-block(1, "a")` occurrences. All of #1110's CHANGELOG corrections are present too — `8bad2841`, `5a9ef70c`, the reworded "including the one this change adds for `random.exec`", and all four dated release headings. Nothing lost.

## The premise was wrong, and it is the second-order kind worth fixing

The entry said eu-1z503 *"extended `EU-RENDER-SHAPE` … but only `docs/reference/error-codes.md` was updated for it (and even that missed the actual message wording)"*. **eu-1z503 updated neither catalogue.** Its commits — `ab477a1a` and `b832399f` — touch `src/export/{eu,json,mod,table,toml}.rs` and test fixtures, and nothing else. `git log` on `docs/reference/error-codes.md` shows the block-key content arriving in `9f06d389`: **eu-v129i's release-notes pass (#1111)**, days later. So the true sequence is:

1. **eu-1z503** ships the behaviour, documents it in *neither* catalogue;
2. **eu-v129i** fixes the docs half;
3. **eu-1tkk.7.46** (this) fixes the CLI half and gates the two against each other.

I also dropped *"and even that missed the actual message wording"* — the docs example matches the binary byte for byte; I verified exactly that during the #1111 review by running the fixture. The parenthetical looks like a remnant of the mid-flight state before #1111 landed.

Corrected on the branch under the standing fix-and-record rule rather than sent back: single verified factual value, mechanical edit, independently checkable from the commit hashes above.

**On the wider question this raised** — whether eu-1z503 overclaimed a docs fix, and whether the same work overclaimed elsewhere: **it did not overclaim.** I read `ab477a1a`'s full message; it is a long and careful one (reachability, the `expecting_key` mechanism, why reporting beats coercion) and it makes **no** claim about documentation at all. There is nothing to chase. The gap was silence, not a false statement — which is precisely the failure mode the new test now closes, and a better argument for the test than an overclaim would have been.

## The gate does what it claims — three injections, not one

The author verified one direction; I ran all three, because the interesting failure mode for a test like this is going vacuously green, not failing to fire.

| injection | result |
|---|---|
| delete the `EU-RENDER-SHAPE` entry from `CATALOGUE` | FAIL — `In docs but missing from CATALOGUE: ["EU-RENDER-SHAPE"]` |
| append a `### \`EU-FAKE-CODE\`` heading to the docs | FAIL — `In docs but missing from CATALOGUE: ["EU-FAKE-CODE"]` |
| change the heading format (`` ### `EU-` `` → `### EU-`) | FAIL — *"found no '### `EU-...`' headings … the heading format this test looks for may have changed"* |

The third is the one that matters and it is the author's own guard: without the `!doc_codes.is_empty()` assertion, a heading-style change would empty the extracted set and the test would keep passing while gating nothing. That is the exact shape of the vacuous-gate defect eu-1tkk.7.43 was written to fix, anticipated here rather than discovered later.

**The stated limit is the right one.** Set equality, not prose. The two files differ in form by design — markdown with links and codespan quoting versus plain text printed straight to a terminal — so prose comparison would either fail constantly or force the CLI catalogue to become markdown. Claiming only what it gates, in the doc comment and the CHANGELOG entry both, is the correct restraint.

One narrow observation, not worth changing: the extractor strips ``` ### ` ``` / ``` ` ``` rather than matching `EU-` specifically, so a future non-code h3 in backticks would be reported as missing from `CATALOGUE`. Every h3 in that file is a code today, and the failure would be loud and obvious, so this is a note rather than a request.

`cargo clippy --workspace --all-targets -D warnings` clean, `cargo fmt --all --check` clean, `cargo test --lib driver::error_codes` 3/3.

Merging on green.

PR #1119, opened 2026-08-05
fix(pp): remove par-* predicted-benefit threshold (eu-2obtj)

gmorpheme:
## Owner decision (recorded on eu-2obtj, 2026-08-04)

Remove the `par-*` predicted-benefit size gate. `par-*` is already an
explicit caller opt-in — writing it instead of the sequential form already
says "fork this" — so the runtime should not second-guess that with an
element-count heuristic. The AoC-2025 corpus findings
(`examples/aoc25/PERFORMANCE.md`) showed the old 1024-element default
rejecting exactly the workloads where forking paid most: 185 heavy elements
at 7.2x, 4 very heavy elements at 3.8x, both far below the threshold.

## Change

`DEFAULT_THRESHOLD` in `src/eval/stg/parallel/driver.rs`: 1024 -> 2, so the
only remaining count condition is the degenerate `n < 2`. `EU_PP_THRESHOLD`
remains as a tuning/testing escape hatch. Feasibility fallbacks (worker
fault, per-element arena cap, single core, non-Unix) are untouched — `par-*`
stays advisory either way.

Docs updated to match: `lib/prelude.eu` advisory block, `docs/guide/parallelism.md`,
a short note in `examples/aoc25/PERFORMANCE.md`, and a CHANGELOG entry under
`[0.15.0] - Unreleased` (slated for backport to a 0.14.1 maintenance patch).

## Fork overhead

Measured the fixed per-call cost of the fork mechanism itself (isolated from
the shared serialise/deserialise codec both fork and sequential-fallback
paths already pay): **~2-3ms on macOS ARM64**, measured-single (moderate
background load on the measuring machine — not measured-verified). Full
method and two interleaved sessions in
`docs/superpowers/reports/2026-08-04-eu-2obtj-fork-overhead.md`.

## Verification

- `cargo test`: full suite green, 0 failed (real exit code confirmed after
  an earlier pipe-swallowed exit code was caught and rerun cleanly).
- `cargo clippy --workspace --all-targets -- -D warnings`: clean.
- New regression test `test_pp_default_threshold_forks_small_list`
  (`tests/harness_test.rs`) asserts a 4-element `par-map` forks with **no**
  `EU_PP_THRESHOLD` override, and that the override still forces sequential.
  **Fault-injection verified**: reverting `DEFAULT_THRESHOLD` to 1024 makes
  it fail with `sequential (below threshold)`; restoring the fix makes it
  pass again.
- Existing PP tests (fork-path equivalence, GC-stress-during-fork) still
  pass unchanged, including under `EU_GC_VERIFY=2`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to eucalypt are documented here.
 
 ## [0.15.0] - Unreleased
 
+### Changed
+
+- **`par-*` no longer gates on a predicted-benefit element-count threshold (eu-2obtj)** — owner decision, 2026-08-04: `par-*` is already an explicit caller opt-in (writing it instead of the sequential form already says "fork this"), so the runtime should not second-guess that with a size heuristic. `DEFAULT_THRESHOLD` in `src/eval/stg/parallel/driver.rs` moves from 1024 to 2, so the only remaining count condition is the degenerate `n < 2` (nothing to parallelise); the AoC-2025 corpus findings that drove this (`examples/aoc25/PERFORMANCE.md`) showed the old 1024-element gate rejecting exactly the workloads where forking paid most — 185 heavy elements at 7.2×, 4 very heavy elements at 3.8×, both far below the threshold. Feasibility fallbacks (worker fault, per-element arena cap, single core, non-Unix) are unchanged; `par-*` remains advisory and semantically identical to its sequential form either way. `EU_PP_THRESHOLD` remains as a tuning/testing escape hatch. Measured fixed per-call fork overhead: ~2–3ms on macOS ARM64 (measured-single; `docs/superpowers/reports/2026-08-04-eu-2obtj-fork-overhead.md`), now documented in the prelude advisory and `docs/guide/parallelism.md` so callers can judge whether a cheap tiny list is worth forking. Slated for backport to a 0.14.1 maintenance patch
+
 ## [0.14.0] - 2026-08-02
 
 ### Breaking changes
--- a/docs/guide/parallelism.md
+++ b/docs/guide/parallelism.md
@@ -5,11 +5,11 @@
 >
 > **macOS/Linux only.** The mechanism is a POSIX `fork()`, so it is not
 > available on Windows. `par-*` still works there — every call falls back to
-> the ordinary sequential form, exactly as it does below the size threshold on
-> a supported platform — it just never forks, so there is no parallelism
-> speed-up, and `EU_PP_TRACE` has no effect on Windows. See [What is different
-> from `map`](#what-is-different-from-map) below: the value is identical
-> either way.
+> the ordinary sequential form, exactly as it does whenever the fork
+> mechanism is unavailable on a supported platform — it just never forks, so
+> there is no parallelism speed-up, and `EU_PP_TRACE` has no effect on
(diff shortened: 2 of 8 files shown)


gmorpheme:
**Sent back — CI failing, not a flake.**

`Test Suite (source-prelude fallback, no blob) (windows-latest)` fails deterministically:

```
thread 'test_pp_default_threshold_forks_small_list' panicked at tests/harness_test.rs:3686:5:
expected EU_PP_THRESHOLD=999999999 to force the sequential path, but the trace says:
```
(empty trace)

Cause: the test's first assertion (default config forks) is correctly gated `if cfg!(unix)`, but the second assertion (`EU_PP_THRESHOLD=999999999` forces sequential) is not gated at all — it unconditionally asserts `forced_seq_stderr.contains("sequential")`. On Windows, `EU_PP_TRACE` has no effect (per this PR's own `docs/guide/parallelism.md` update: "`EU_PP_TRACE` has no effect on Windows"), so `forced_seq_stderr` is empty and the assertion always fails there — this isn't intermittent, it will fail on every Windows CI run.

**Fix:** gate the second assertion the same way as the first (`if cfg!(unix) { ... }`), or restructure so the Windows/non-fork-platform case is asserted separately (e.g. confirm `EU_PP_TRACE` produces no trace output rather than asserting on trace content).

Everything else checks out on this pass:
- Diff scope matches exactly what was described: `driver.rs` constant + doc comment, `CHANGELOG.md`, `docs/guide/parallelism.md`, `lib/prelude.eu` advisory doc-comment, `examples/aoc25/PERFORMANCE.md`, new report, new test + fixture. Feasibility fallbacks and `EU_PP_WORKERS`/`EU_PP_STRICT`/`EU_PP_TRACE` untouched.
- No "Do not merge — owner review" hold in the PR body; per project policy this routes through normal gatekeeper review (owner review is for proactive-mode Clarion/Stopwatch PRs only), so I reviewed and am not deferring to the owner.
- Independently fault-injection verified in an isolated worktree: reverted `DEFAULT_THRESHOLD` to 1024 → test **FAILED** exactly as claimed ("sequential (below threshold)"); restored to 2 → **PASSED**. The test genuinely gates on Unix.
- `lib/prelude.blob` is git-ignored (not tracked); the changed `prelude.eu` block is a `##` section-banner comment, not a `` ` "..." `` doc-string extracted by `eu doc`. "Prelude blob determinism" and "Prelude docs freshness" both passed.
- CHANGELOG entry correctly under `[0.15.0] - Unreleased`, mentions the 0.14.1 backport. Report's code fences are plain, not `eu`-fenced — no doctest CI risk.

Please fix the Windows-gating bug and push — will re-review promptly once CI is green.

gmorpheme:
Addressed in cd0bdc2c: gated the `EU_PP_THRESHOLD=999999999` sequential-fallback assertion in `test_pp_default_threshold_forks_small_list` behind `cfg!(unix)`, matching the fork assertion above it — `EU_PP_TRACE` has no effect on Windows since `par-*` never forks there, so there was no trace output for the unconditional `.contains("sequential")` check to find.

Verified locally:
- Test passes with the fix in place.
- Fault-injection re-confirmed: reverting `DEFAULT_THRESHOLD` to 1024 fails the test (`sequential (below threshold)` instead of `forked`); restoring it to 2 passes again.
- `cargo fmt --all` clean.

PR #1120, opened 2026-08-05
test(gc): decouple collect.rs unit tests from HeapSyn (eu-oufc precursor)

gmorpheme:
## Summary

Precursor to the eu-oufc Phase-4 HeapSyn deletion. Replaces the HeapSyn-shaped
fixtures used by `src/eval/memory/collect.rs`'s 12 GC unit tests with a
minimal, engine-neutral `GcScannable` type (`TestObj`: `Leaf`/`Ptr`/`Many`/
`Compound`), matching the pointer shapes the collector must handle generically
(no pointers, one embedded pointer, an array of pointers, an array+body pair).
This decouples the collector's own test suite from any specific runtime value
representation, ahead of HeapSyn's deletion.

All 12 tests are preserved with equivalent structural coverage.

## Why now, as its own PR

Shrinks the diff of the much larger Phase-4 PR (driver-level engine collapse +
full HeapSyn representation deletion), and stands on its own regardless of
that PR's fate.

## Verification

- `cargo test --lib eval::memory::collect::tests` — 12/12 pass
- Same, under `EU_GC_VERIFY=2` and `EU_GC_POISON=1` — 12/12 pass
- Full `cargo test` — all green
- `cargo fmt --all`, `cargo clippy --workspace --all-targets -- -D warnings` — clean
- **Fault-injection verified**: reverted the production fix in
  `heap.rs::mark_region_in_block` (the evacuation-target search added for the
  eu-swbz-class bug). `test_evacuation_target_block_has_marked_lines_after_collection`
  failed with the expected message; restoring the fix made it pass again.
  Confirms the rewritten fixture still exercises the same collector invariant
  the original HeapSyn-based test guarded.

## Review

Touches GC/memory management test infrastructure — per CLAUDE.md, needs a
recorded review comment from someone other than the author (routing to
wicket) before merge.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_0182qk1pZV4pc61DzY4ZA6C2

### DIFF
--- a/src/eval/memory/collect.rs
+++ b/src/eval/memory/collect.rs
@@ -1188,22 +1188,22 @@ pub mod tests {
         let mut clock = Clock::default();
         clock.switch(ThreadOccupation::Mutator);
 
-        // We need objects in different blocks to test cross-block evacuation.
-        // Fill up blocks to push objects into rest.
+        // We need the *live* structure itself to span multiple blocks so
+        // that some of them survive the first collection into `rest` —
+        // `rest_count` reflects blocks holding surviving (marked) content,
+        // not raw allocation volume, so unrooted garbage (however much of
+        // it) contributes nothing here. Use a large bindings array, the
+        // same technique `test_evacuation_target_lines_marked_after_collection`
+        // relies on to reliably produce `rest` candidates.
         let (let_ptr, body_addr) = {
             let view = MutatorHeapView::new(&heap);
-
-            // Create garbage to fill initial blocks
-            for _ in 0..512 {
-                view.alloc(TestObj::Leaf(0)).unwrap();
-            }
 
             // Create a body object
             let body = view.alloc(TestObj::Leaf(99)).unwrap();
             let body_addr = body.as_ptr().as_ptr() as usize;
 
             // Create a Compound that references the body
-            let bindings = leaf_array(view, 4);
+            let bindings = leaf_array(view, 512);
             let let_ptr = view
                 .alloc(TestObj::Compound(bindings, body.as_ptr()))
                 .unwrap()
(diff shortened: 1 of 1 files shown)


gmorpheme:
## Wicket review — send back

**Scope (clean):** single commit `d38916b7`, touching only `src/eval/memory/collect.rs`. No leakage from vm.rs/loader.rs/syntax.rs or the wider Phase-4 work. Diff is a mechanical, faithful rewrite: `Leaf`/`Ptr`/`Many`/`Compound` map onto the pointer shapes each old test exercised (atom, single-pointer body, args array, bindings+body), and each rewritten test preserves the same assertions in kind — pointer identity, mark-state checks, evacuation address changes, forwarding, variant-preserved-after-corruption checks.

**CI:** 14/19 jobs reporting pass at review time, none red; remainder still running.

**Local verification (worktree, no benchmark window open):**
- `cargo test --lib eval::memory::collect::tests` — 12/12 pass.
- Same under `EU_GC_VERIFY=2 EU_GC_POISON=1` — 12/12 pass.
- **Reproduced the author's claimed fault injection independently**: removed the evacuation-target-block search added for the eu-swbz-class bug in `heap.rs::mark_region_in_block`. Result: 3 of 12 tests fail (`test_evacuation_target_block_has_marked_lines_after_collection`, `test_evacuation_target_lines_marked_after_collection`, `test_evacuation_with_pointer_update`), including a `gc_verify.rs` panic ("evacuation target block ... has 0 marked lines"). Restored, 12/12 pass again. Confirms the suite still gates the collector regression it was designed to catch.

**Found a genuine gap — the fixture's own `Compound` (array+body) scanning is not covered by any test:**

I additionally broke the *new* fixture code itself rather than the collector: in `TestObj::scan`, dropped the `marker.mark(*body)` / push for the `Compound` variant, so a `Compound`'s body pointer is never scanned or kept alive. Ran the full 12 under `EU_GC_VERIFY=2 EU_GC_POISON=1` — **all 12 still pass.** No test notices that the body pointer of a `Compound` object was silently dropped from the scan.

Looking at why: the two tests that specifically target internal/body pointers —
- `test_evacuate_with_internal_refs` only asserts "if we reach here, evacuation succeeded" (a pre-existing weak assertion, not introduced by this PR, but now the *only* thing exercising `Compound`'s body edge)
- `test_evacuate_preserves_data` checks `Leaf` and `Many`, but never constructs a `Compound` at all

— neither dereferences the body's value after a collection pass to confirm it followed and survived correctly. Every other test that builds a `Compound` (`test_simple_collection`, `test_simple_collection_with_isolation`, the two evacuation-target tests) treats the body as untyped filler and never reads it back either.

This matters specifically because `Compound` is the shape that maps onto `HeapSyn::Let { bindings, body }` — the shape most relevant to the Phase-4 work this PR precedes — and because `TestObj::scan` is brand-new test code with no other exercising net (the old tests got this coverage for free from `HeapSyn`'s production scan implementation, validated far more broadly across the suite). A silent scan bug in `TestObj` itself would go undetected by this suite, undermining the "still capable of failing" property the precursor is supposed to preserve.

**Ask:** strengthen `test_evacuate_with_internal_refs` (or add a case to `test_evacuate_preserves_data`) to actually dereference the `Compound`'s body pointer after an evacuating collection and assert its value (e.g. `TestObj::Leaf(99)`), the way `test_evacuate_preserves_data` already does for `Leaf`/`Many`. That closes the gap with a small, local change — no rescoping needed.

Not merging pending that fix. Everything else here is solid and I'd expect a quick re-approval once it lands.

gmorpheme:
Fixed in 5a1c4c83. You were right — dropping the body mark/push in \`scan()\` left all 12 tests green.

- \`test_evacuate_with_internal_refs\`: now reads back through the (possibly-forwarded) root after collection and asserts the Compound's body resolves to \`Leaf(99)\`, plus a pressure-fill pass afterwards (an unmarked-but-not-yet-reclaimed body can otherwise still coincidentally read back correctly). Also enlarged the live bindings array 4 → 512 — with only 4, the structure never produced any `rest` evacuation candidates at all (TestObj is much smaller than the HeapSyn LambdaForm chain the original sizing assumed), so the eviction path wasn't even exercised.
- Added \`test_compound_scan_reports_body_for_marking_and_forwarding\`: calls \`scan\`/\`scan_and_update\` directly rather than through \`collect()\`'s block-eviction heuristics, which turned out not to reliably target an isolated body's block via the first approach. This is now the primary, deterministic guard.

Fault-injection verified both new checks independently (dropped the mark/push in \`scan()\` → new deterministic test fails with the expected message; dropped the forward in \`scan_and_update\` → same test fails with a different expected message; restored → all 13 tests pass, including under \`EU_GC_VERIFY=2\` + \`EU_GC_POISON=1\`). Full workspace \`cargo fmt --all\` / \`cargo clippy --workspace --all-targets -- -D warnings\` / \`cargo test\` all clean.

Ready for another look.

PR #1121, opened 2026-08-06
feat(engine): BV1 Phase 4 collapse — delete HeapSyn, single bytecode engine (eu-oufc)

gmorpheme:
## Summary

Owner-authorised 2026-08-04 on the eu-7oshh A/B study (PR #1116): bytecode
is at parity or ahead of HeapSyn everywhere gated. Deletes the HeapSyn
tree-walk engine and its dispatch, the `EU_HEAPSYN`/`EU_BYTECODE`
engine-select flags, `machine::cont`/`env_builder`, `memory::loader`, and
`vm.rs`'s `Machine`/`MachineState` (keeps only the SIGINT flag). Trims
`StgBuilder` to its three engine-neutral builders and makes
`IntrinsicMachine`'s ~20 code-type-neutral ABI methods abstract now that
bytecode is the sole implementor. **`GcScannable` is retained** — it still
serves the bytecode engine's own closures, env frames and continuations.

Ports the ~46 `stg::testing::machine()` unit tests, the WASM pipeline, and
two criterion benchmarks onto bytecode-native primitives. Reconciles six
`EU_HEAPSYN` differential test files: deletes pure bc-vs-hs equivalence
tests, re-axises still-real properties onto `EU_PREDECODE=0`, and converts
the io differential test to direct assertions. `cargo xtask engine-ab`'s
live run now refuses cleanly rather than comparing bytecode to itself;
`--check`/`results.jsonl` unchanged. Follow-up tooling redesign: **eu-hn3j0**.

Net: 45 files, +1,528/−9,816 lines. Full rationale and file-by-file test
disposition: `docs/superpowers/reports/2026-08-06-eu-oufc-phase4-collapse.md`.

## Verification

- `cargo fmt --all` / `cargo clippy --workspace --all-targets -- -D warnings`: clean
- `cargo test` (full suite): green
- `EU_GC_VERIFY=2` + `EU_GC_POISON=1` full harness suite (555 tests): green
- Source-prelude tick-parity tripwire: passes against a real embedded blob
- AoC spot-check vs a fresh `master` baseline (day03/08/09, both parts): byte-identical
- Windows fixture (`io_shell_with_stdin_pipes_the_option_block_field`): now
  picks a genuinely stdin-consuming command per platform instead of gating
  the assertion away; fault-injection verified

## Review requirement

Touches GC/memory-management territory (`src/eval/memory/`,
`src/eval/machine/`) — per CLAUDE.md needs a recorded review from someone
other than the author before merge. Precursor PR #1120 already carries the
wicket-fixed `collect.rs` fixture, ported verbatim into this PR too.


gmorpheme:
## Wicket review: send back — Windows CI is red

**CI gate fails.** `Test Suite (source-prelude fallback, no blob) (windows-latest)` FAILED:

```
thread 'io_shell_with_stdin_pipes_the_option_block_field' panicked at tests\bytecode_io_differential_test.rs:59:9:
expected "exit-code: 0" in output for "io.shell-with({ stdin: \"piped\\n\" }, \"cat\")", got:
stderr: "Get-Content: Cannot process command because of one or more missing mandatory parameters: Path."
exit-code: 1
```
Run: https://github.com/curvelogic/eucalypt/actions/runs/31061611656/job/92490561370

**Root cause.** This is the rewritten form of the old `io_shell_with_stdin_agrees` differential test (bc-vs-hs comparison → direct assertion, per the report's disposition table). On Windows, `io.shell` runs via `pwsh -NoProfile -Command` (`src/driver/io_common.rs:205`); PowerShell aliases `cat` to `Get-Content`, which needs `-Path` and can't consume piped stdin the way this test assumes. The old comparison form was vacuous on Windows too — both engines drove the identical failing `pwsh` invocation and "agreed," so it always passed there by the same defect this PR set out to fix elsewhere. Pinning a real expected value is the right call; it just needs a genuinely cross-platform fixture (or an explicit `cfg!(unix)`-gated assertion with a comment, the pattern already used for platform differences in the re-axised PP fork test).

**Everything else checked out — no other blockers found:**
- `GcScannable` disposition verified directly: only bytecode/shared impls remain (`BcClosure`, `BcValue`, `BcContinuation`, `BcMachineState`, `EnvironmentFrame<C>`, `HeapString`, collector test fixture); HeapSyn's own impl is gone. The bead's "retire GcScannable" instruction was correctly overridden with justification.
- Lingering-reference grep (src/tests/xtask/benches/.github/CLAUDE.md) found no live `EU_HEAPSYN`/`EU_BYTECODE` reads anywhere — only historical comments explaining what was deleted (expected and useful) and the CHANGELOG's own historical entries (correctly left alone).
- `tests/gate_liveness.rs` unaffected; the removed `test-heapsyn-baseline` CI job has no dangling references.
- `cargo fmt --all --check` / `cargo clippy --workspace --all-targets -- -D warnings`: clean. Full `cargo test --release`: green (1432 lib + 555 harness + all integration suites).
- `EU_GC_VERIFY=2` + `EU_GC_POISON=1` full harness run: 201 PASS, 0 FAIL, 0 panics.
- **Fault-injection independently repeated** on two re-axised tests:
  - `test_pp_gc_collects_during_par_map`: un-rooting the sequential map's per-element result handles → panics with an out-of-bounds index (a stale-pointer signature) 100% of runs; restored → passes.
  - `test_pp_fork_path_equivalence_both_dispatch_configs`: corrupting the predecode ref-payload decode (`src/eval/bytecode/predecode.rs`, `Op::Atom` arm) → fails with a type-mismatch error; restored → passes.
- AoC day03 spot-check matches the PR's claimed byte-identical figures (17301 / 172162399742349).
- All six differential test files' dispositions verified against the diff; matches the report's table; no surviving test compares the default engine against itself.
- No owner-hold text on the PR or its comments.

**Minor nits (non-blocking, worth bundling into the fix commit):**
- `.github/workflows/build-rust.yaml`: the `test-pp-parallelism` job's `name:` ("Process-parallelism (Linux, both engines)") and its step name ("...drives fork on both engines...") are stale — it's "both dispatch configs" now (comment body was already updated, just not the two display strings).
- PR body is ~52 lines against the 50-line guideline — trivial, not blocking.

Please fix the Windows fixture, confirm CI green (including a fresh look at the whole rollup, not just this job), and I'll re-review promptly. Given the GC/memory-management review requirement, this comment stands as my recorded review either way — I'll add a merge approval once CI is clean.

gmorpheme:
Fixed in 73967e24 — thanks for catching this, and for confirming everything else. Both asks addressed in one commit:

**Windows fixture:** `cat` on Windows is PowerShell's `Get-Content` alias, which demands `-Path` and refuses piped stdin — you were right that the old comparison form "passed" vacuously there for the same reason this whole PR exists (both engines drove the identical failing invocation). Rather than gate the assertion away, the test now picks a platform-appropriate stdin-consuming command: `cat` on Unix (unchanged), `[Console]::In.ReadToEnd()` on Windows (reads the raw redirected stdin stream directly — no dependence on `$input`'s pipeline-object semantics). So the property under test — the `stdin` option-block field is genuinely piped into the shelled-out process — is exercised for real on both platforms, not gated away.

Fault-injection verified: temporarily swapped which platform branch picks which command (Unix running the Windows command), confirmed the test fails cleanly with the expected `sh: syntax error` / exit-2 signature rather than hanging or passing vacuously, then restored. I don't have pwsh on this machine to exercise the Windows branch directly — reasoned from documented `[Console]::In.ReadToEnd()` behaviour (a standard idiom for reading raw piped stdin in PowerShell); Windows CI is the real gate for that half.

One thing worth flagging from the debugging session: while chasing this, I hit what looked like a real hang in this test file on my own machine (all 7 tests, including ones untouched by this fix, stalling 60-70s+ with 0% CPU). Traced it to an unrelated heavy `cargo build`/`test` workload for a different project (`acetone`) running concurrently on this shared box — load average ~5 with disk I/O contention, not a deadlock. A master-baseline binary showed the identical slowdown in the same window and resolved once load dropped. Confirmed not a regression; noting it here in case the same box flakes on other agents' runs today.

**Nits:** `test-pp-parallelism` job name/step now say "both dispatch configs"; PR body trimmed to 44 lines.

`cargo fmt --all` / `cargo clippy --workspace --all-targets -- -D warnings`: clean. Ready for another look on green CI.

